PR #10, opened 2021-11-28
Create a UI from which a user can update their config

### LINKED ISSUE #7: Create a config UI
- Add a UI layer for manipulating the config

sabaimran:
# Incoming Changes
- Create a basic UI page which the user can utilize to update their config and regenerate file embeddings. If there are type errors in their config input, they will not be able to update the config as intended.
- Integrate pydantic.BaseModel into the config data models in order to streamline input validation and mapping

## Screenshot
image

# Test
- Check out this branch and start the service
- Navigate to https://localhost:8000/ui to view the config page
  - Click on "update" to update the source config file
  - Click on "regenerate" to regenerate the embeddings

# Refer:
- https://pydantic-docs.helpmanual.io/usage/models/

Closes #7


.gitignore
```diff
@@ -3,4 +3,6 @@ __pycache__
 .emacs.desktop*
 tests/data/.*
 src/.data
-.vscode
\ No newline at end of file
+.vscode
+*.gz
```
sabaimran, line 7: Should these be in `.gitignore`?

src/utils/config.py
```diff
@@ -43,49 +45,22 @@ class SearchModels():
 
 
 class TextSearchConfig():
```
debanjum, line 47: Would making the `TextSearchConfig` directly inherit from `ConfigBaseModel` simplify the code even further? That is have `TextSearchConfig(ConfigBaseModel)`

src/utils/rawconfig.py
```diff
@@ -0,0 +1,62 @@
...
+
+class ImageSearchTypeConfig(ConfigBaseModel):
+    encoder: Optional[str]
+
+class SearchTypeConfigTest(ConfigBaseModel):
+    asymmetric: Optional[AsymmetricConfig]
+    image: Optional[ImageSearchTypeConfig]
+
+class ProcessorConversationConfig(ConfigBaseModel):
+    open_api_key: Optional[str]
```
debanjum, line 52: Shouldn't `open_api_key`, `conversation_logfile` and `conversation_history` all be required field in the `ProcessorConversationConfig`?

src/utils/rawconfig.py
```diff
@@ -0,0 +1,62 @@
...
+    cross_encoder: Optional[str]
+
+class ImageSearchTypeConfig(ConfigBaseModel):
+    encoder: Optional[str]
+
+class SearchTypeConfigTest(ConfigBaseModel):
+    asymmetric: Optional[AsymmetricConfig]
+    image: Optional[ImageSearchTypeConfig]
+
+class ProcessorConversationConfig(ConfigBaseModel):
```
debanjum, line 51: Can we call this `ConversationProcessorConfigModel` to follow the current naming convention being followed?
```suggestion
class ConversationProcessorConfigModel(ConfigBaseModel):
```

src/main.py
```diff
@@ -1,26 +1,49 @@
...
 from src.utils.helpers import get_absolute_path
 from src.utils.cli import cli
 from src.utils.config import SearchType, SearchModels, TextSearchConfig, ImageSearchConfig, SearchConfig, ProcessorConfig, ConversationProcessorConfig
+from src.utils.rawconfig import FullConfig
 from src.processor.conversation.gpt import converse, message_to_log, message_to_prompt, understand
 
-
 # Application Global State
 model = SearchModels()
 search_config = SearchConfig()
```
debanjum, line 22: Maybe we can collate all these `config` related global variables into a single global variable/class/object? Don't have to do it as part of this PR necessarily, just a note for later reference

.gitignore
debanjum: I think some *.gz files under tests/data maybe used for tests

sabaimran: I believe those are already ignored by the line `tests/data/.*`

src/utils/rawconfig.py
sabaimran: Should all of the types here have the `Model` suffix?

debanjum: That's what I would suggest. It'll keep the naming more consistent

src/utils/rawconfig.py
sabaimran: Yes, but changing this will require refactoring the code a bit. I would suggest doing this in a separate CR, so that this one can be limited to creating the config UI and integrating the most minimal addition of BaseModel types. Opening #12 to track that work.

src/utils/config.py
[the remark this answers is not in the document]
sabaimran: Removed `TextSearchConfig` altogether. I replaced these types with the `rawconfig.py` types. The only meaningful difference between them was the `verbose` flag, which I moved out of the config models. I feel that flag doesn't logically fall under the purview of the config models, as it's more pertinent to the global application state. Open to discussion!

src/main.py
```diff
@@ -69,22 +92,7 @@ def search(q: str, n: Optional[int] = 5, t: Optional[SearchType] = None):
...
-
-    if (t == SearchType.Ledger or t == None) and search_config.ledger:
-        # Extract Entries, Generate Embeddings
-        model.ledger_search = symmetric_ledger.setup(search_config.ledger, regenerate=True)
-
-    if (t == SearchType.Image or t == None) and search_config.image:
-        # Extract Images, Generate Embeddings
-        model.image_search = image_search.setup(search_config.image, regenerate=True)
-
+    initialize_search(regenerate=True)
```
debanjum, line 95: We should pass type (i.e `t: Optional[SearchType]`) to the `initialize_search` method

src/search_type/symmetric_ledger.py
```diff
@@ -143,7 +141,7 @@ def collate_results(hits, entries, count=5):
         in hits[0:count]]
 
 
-def setup(config: TextSearchConfig, regenerate: bool) -> TextSearchModel:
+def setup(config: TextSearchConfigModel, regenerate: bool, verbose: bool) -> TextSearchModel:
```
debanjum, line 144: Let's drop `Model` from classes with names ending with `ConfigModel` now that we've removed `TextSearchConfig` and `ImageSearchConfig`

src/main.py
sabaimran: Opened #13.

PR #21, opened 2022-01-24
Add Docker support to semantic-search

### LINKED ISSUE #20: Dockerize the application
This should make setup and deployment more straightforward.

### LINKED ISSUE #22: Config expects str for use_xmp_metadata, rather than bool

sabaimran:
## Add Docker support
- Add a Dockerfile which builds from the `continuumio/miniconda3` image
   - Installs system-level dependencies (such as exiftool)
   - Performs Conda environment setup required to run the semantic-search application
- Add a docker-compose.yml
   - Exposes application on port 8000 and passes the program arguments (e.g., `port`, `host`, `config`)
   - Mounts the files to be indexed into volumes in the Docker image 
- Add relevant comments for documentation and rationale for why the files are setup in the way they are

Tested locally. To run:
- Checkout the branch and modify the file paths to match your local directories in `docker-compose.yml`
- Run `docker-compose up` from the root directory
- The application should be hosted on http://localhost:8000/ and can be run as usual

Closes #20

**NOTE**: I haven't figured out how to cache the Conda environment between builds, so building the image takes like 20-30 minutes (every time). You can make changes to `docker-compose.yml` and `docker_sample_config.yml` without having to rebuild.

## Bug fix
- Fixes a bug in the `rawconfig` specification which was setting the `use_xmp_metadata` field in the image search config to be a `str`, rather than `bool`. Update relevant configs.

Closes #22

## Update README.org
- Add instructions for running using Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+# syntax=docker/dockerfile:1
+FROM ubuntu:18.04
+
+# Install system dependencies and Python packages
+RUN apt-get update -y && \
+    apt-get -y install libimage-exiftool-perl
+
+FROM continuumio/miniconda3
+
+COPY . /src
+WORKDIR /src
+
+COPY environment.yml .
+COPY config.yml .
+
+RUN conda env create -f environment.yml
+
+EXPOSE 5000
+COPY . .
+# CMD python3 -m main -c=config.yml -vv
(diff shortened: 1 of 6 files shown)


debanjum:
Can we update the Readme with instructions on setting up semantic search via Docker?

PR #99, opened 2022-09-19
fix: explicitly set url-request-method

telotortium:
In my installation, it appears that `url-request-method` is sometimes set globally to POST.  Need to explicitly set it to ensure that GET is always used as intended.

### DIFF
--- a/src/interface/emacs/khoj.el
+++ b/src/interface/emacs/khoj.el
@@ -226,7 +226,8 @@ Use `which-key` if available, else display simple message in echo area"
 
 (defun khoj--get-enabled-content-types ()
   "Get content types enabled for search from API."
-  (let ((config-url (format "%s/config/data" khoj-server-url)))
+  (let ((config-url (format "%s/config/data" khoj-server-url))
+        (url-request-method "GET"))
     (with-temp-buffer
       (erase-buffer)
       (url-insert-file-contents config-url)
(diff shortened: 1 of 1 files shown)


debanjum:
Apologize for the delayed review. @telotortium just to get a better handle on the specific issue:
- Could you specify which version of Emacs you're on (i.e `M-x emacs-version`)?
- Do you know where `url-request-method` is being set globally? In your `init.el` or some library being loaded earlier? 
  - I'd expect GET requests to fail in other libraries as well if `url-request-method` is set to POST globally as `url-request-method` isn't explicitly set before GET requests usually (as it's the default)

telotortium:
@debanjum 

- `emacs-version`: GNU Emacs 28.1 (build 2, x86_64-apple-darwin21.5.0, NS appkit-2113.50 Version 12.4 (Build 21F79)) of 2022-07-25
- Where is `url-request-method` being set globally?
   - Using `debug-on-variable-change`, I found that a library in my installation sets this variable globally when making requests. It uses `unwind-protect`, so in theory it should reset the variable correctly, but sometimes that appears to fail - not sure why, maybe a macro is interfering?
- Why haven't I noticed this before?
   - Most of the libraries I have use the `request` library, which has a cURL backend that is used preferentially over `url-retrieve`. Thus, most of my libraries don't look at `url-request-method`. Even if you don't have cURL, `request` binds `url-request-method` in all cases using `let`.

debanjum:
@telotortium, thanks for sharing details and finding the root-cause of the issue! It makes sense

PR #102, opened 2022-10-28
Fix khoj.el reaching for a URL that doesn't exist

thatboyjake:
since /api doesn't exist I just changed the urls the emacs search grabs and I have tested it

### DIFF
--- a/src/interface/emacs/khoj.el
+++ b/src/interface/emacs/khoj.el
@@ -226,7 +226,7 @@ Use `which-key` if available, else display simple message in echo area"
 
 (defun khoj--get-enabled-content-types ()
   "Get content types enabled for search from API."
-  (let ((config-url (format "%s/api/config/data" khoj-server-url))
+  (let ((config-url (format "%s/config/data" khoj-server-url))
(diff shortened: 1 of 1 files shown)


debanjum:
Hi @thatboyjake, apologize for the confusion. 
The latest code on master has `/api` prepended to the API routes since [this PR](https://github.com/debanjum/khoj/pull/101). 
Upgrading your khoj app by following either of the below steps should fix the issue:
- Use latest khoj code on master by running `pip install --pre khoj-assistant`
  This gets you the pre-release version of the package which tracks the khoj master branch. 
- Once the next version of khoj is released (soon), you can just run `pip install --upgrade khoj-assistant`

thatboyjake:
okay seems like you should keep a melpa package or other way to install a khoj.el that works on non master branch khoj installs

debanjum:
Yeah, need to submit `khoj.el` to MELPA. Then you'd be able to use MELPA stable to install `khoj.el` and only get the (git) tagged versions of `khoj.el`. It'll help avoid this kind of issue.

Issue #134, opened 2023-02-05
[Feature Request] [Obsidian] Show name of the note in the search modal


sohanglal:
Would it be possible to show the name of the note in the seach modal?

PR #135, opened 2023-02-05
Restyled Obsidian Search Results

jimprince:
Thought search results were a bit hard to interpret, so I restyled them a bit. Sorry, I'm not a developer so apologizes in advance if there's etiquette around contributions I'm missing.

Here's what the search results look like:

image

It should maintain reasonable contrast with most themes as the CSS uses theme variables rather than hard coded colors. 

Should work with Mac, Windows and Linux, though I've only tested it on Mac. The interoperability challenge comes from the fact that the API returns a path as a string, which presumably would have different slashes on OSX, Linux compared to windows?

Also addresses this issue: #134 

Cheers,

### DIFF
--- a/src/interface/obsidian/src/modal.ts
+++ b/src/interface/obsidian/src/modal.ts
@@ -1,4 +1,4 @@
-import { App, SuggestModal, request, MarkdownRenderer, Instruction, Platform, TFile } from 'obsidian';
+import { App, SuggestModal, request, MarkdownRenderer, Instruction, Platform} from 'obsidian';
 import { KhojSetting } from 'src/settings';
 
 export interface SearchResult {
--- a/src/interface/obsidian/src/modal.ts
+++ b/src/interface/obsidian/src/modal.ts
@@ -105,7 +105,7 @@ export class KhojModal extends SuggestModal<SearchResult> {
             path = result.file.split('/').slice(0, -1).join('/');
             basename = result.file.split('/').slice(-1)[0];
         }
-
+        basename = basename.replace('.md', '');
         let entry_lines_split = result.entry.split('\n')
         // remove frontmatter if present
         if (entry_lines_split[0] == '# ---'){
--- a/src/interface/obsidian/src/modal.ts
+++ b/src/interface/obsidian/src/modal.ts
@@ -1,4 +1,4 @@
-import { App, SuggestModal, request, MarkdownRenderer, Instruction, Platform } from 'obsidian';
+import { App, SuggestModal, request, MarkdownRenderer, Instruction, Platform, TFile } from 'obsidian';
 import { KhojSetting } from 'src/settings';
 
 export interface SearchResult {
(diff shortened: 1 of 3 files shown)


src/interface/obsidian/src/modal.ts
```diff
@@ -90,11 +90,42 @@ export class KhojModal extends SuggestModal<SearchResult> {
...
+            //loop through lines until we find the end of the frontmatter
+            for (let i = 1; i < entry_lines_split.length; i++){
+                if (entry_lines_split[i] == '---'){
+                    //remove the frontmatter
+                    entry_lines_split = entry_lines_split.slice(i+1);
+                    break;
+                }
+            }
+        }
+        entry_lines_split = entry_lines_split.join('\n').trim().split('\n');
```
debanjum, line 121: This line feels unnecessary. The `trim()` used to collate `entry_words` should suffice to remove random whitespaces at the ends of each result, if any.

src/interface/obsidian/src/modal.ts
```diff
@@ -90,11 +90,42 @@ export class KhojModal extends SuggestModal<SearchResult> {
...
+        let entry_lines = entry_lines_split.slice(0, lines_to_render).join('\n');
+        let entry_words_split = entry_lines.split(' ');
+        let entry_snipped_indicator = entry_words_split.length > words_to_render ? ' **...**' : '';
+        let entry_words = entry_words_split.slice(0, words_to_render).join(' ').trim();
+        el.createEl("div",{ cls: 'khoj-result-file' }).setText(basename);
+        var result_div = el.createEl("div", { cls: 'khoj-result-entry' })
+        MarkdownRenderer.renderMarkdown(entry_words + entry_snipped_indicator, result_div, path, null);
```
debanjum, line 128: 1. The `renderMarkdown` expects the markdown file path. This seems to pass the path to the markdown file's directory instead. So passing `results.file` maybe more relevant than passing `path`?
2. I'd left this as `null` as (relative/internal) links within a result shown in the search modal can't be interacted with anyway.

Let me know if you know of any benefits to passing the file path in to the renderer? Else IMO, we should just pass `null` for the `path` argument

src/interface/obsidian/src/modal.ts
```diff
@@ -90,11 +90,42 @@ export class KhojModal extends SuggestModal<SearchResult> {
...
+        // set path and basename of file depending on OS
+        if (result.file.indexOf('\\') > -1) {
+            path = result.file.split('\\').slice(0, -1).join('\\');
+            basename = result.file.split('\\').slice(-1)[0];
+        }
+        if (result.file.indexOf('/') > -1) {
+            path = result.file.split('/').slice(0, -1).join('/');
+            basename = result.file.split('/').slice(-1)[0];
+        }
+        basename = basename.replace('.md', '');
```
debanjum, line 108: 1. We should leave the files `.md` extension in, to act as an indicator that the string is a filename. It'll also be a useful indicator later if/when we support searching for other file types (e.g for images in the users obsidian vault)
2. We don't need to extract `path` based on my other comment
3. Based on above two points, this code can be simplified like below:
```suggestion
        // Extract filename of result
        let os_path_separator = result.file.includes('\\') ? '\\' : '/';
        let filename = result.file.split(os_path_separator).pop();
```

debanjum (changes requested):
Hey @jimprince, thanks for raising a PR!
1. Can you share the latest screenshots of the changes. Don't see the filenames rendered in the screenshot you shared
2. Can you elaborate what about the default search modal styling isn't good enough? 
     I'd rather avoid the maintenance burden of adding custom styling unless there's a solid use-case for it. 
     For example: We'd need to ensure the custom styling looks acceptable on mobile, light/dark mode etc. 

Did a first pass review of the suggested changes and left some code comments

debanjum:
Hey @jimprince, not super urgent but just wanted to check if you'll get time to resolve the PR feedback anytime soon?
Else I can look into resolving some of the related Obsidian plugin issues like #134 separately.

jimprince:
Hey, sorry for the delay, I'm just getting back on my feet after a head injury. Your comments seem pretty reasonable, I'll make the adjustments.

debanjum:
No issues, thanks for informing. Take care of yourself, this can wait!

debanjum:
Hey @jimprince, hope you're recovery fine. I've gone ahead and implemented your changes from this PR in #198. Thanks for taking the initiative 🙏🏾, it helped reduce my time to incorporate them

PR #152, opened 2023-02-10
Fix obsidian plugins doc link

axelson:
Also make it more obvious where the link is going, initially I thought the link was to another official khoj documentation site.


debanjum:
Thanks for creating PR to fix the link and description to the obsidian plugin docs!

PR #224, opened 2023-06-05
Pass truncated message as string in ChatMessage when exceeding max prompt size

### LINKED ISSUE #223: Prompt exceeding max prompt size returns validation error
When I have a result from my notes passed to Open AI that exceeds the size of the maximum prompt length, I get the following error:
```
1 validation error for ChatMessage
content
  str type expected (type=type_error.str)
```

sabaimran:
# Incoming
Fix the way the `truncated_message` is passed to the `ChatMessage` object in the `/chat` flow. It expects a `string`, but it's currently passing the data in a `list`.

Closes #223

### DIFF
--- a/src/khoj/processor/conversation/utils.py
+++ b/src/khoj/processor/conversation/utils.py
@@ -97,23 +97,33 @@ def generate_chatml_messages_with_context(
     messages = user_chatml_message + rest_backnforths + system_chatml_message
 
     # Truncate oldest messages from conversation history until under max supported prompt size by model
+    messages = truncate_message(messages, max_prompt_size[model_name], model_name)
+
+    # Return message in chronological order
+    return messages[::-1]
+
+def truncate_message(messages, max_prompt_size, model_name):
+    """Truncate messages to fit within max prompt size supported by model"""
     encoder = tiktoken.encoding_for_model(model_name)
     tokens = sum([len(encoder.encode(message.content)) for message in messages])
-    while tokens > max_prompt_size[model_name] and len(messages) > 1:
+    logger.info(f"num tokens: {tokens}")
+    while tokens > max_prompt_size and len(messages) > 1:
         messages.pop()
         tokens = sum([len(encoder.encode(message.content)) for message in messages])
 
     # Truncate last message if still over max supported prompt size by model
-    if tokens > max_prompt_size[model_name]:
-        last_message = messages[-1]
-        truncated_message = encoder.decode(encoder.encode(last_message.content))
+    if tokens > max_prompt_size:
+        last_message = '\n'.join(messages[-1].content.split("\n")[:-1])
+        original_question = '\n'.join(messages[-1].content.split("\n")[-1:])
+        original_question_tokens = len(encoder.encode(original_question))
+        remaining_tokens = max_prompt_size - original_question_tokens
+        truncated_message = encoder.decode(encoder.encode(last_message)[:remaining_tokens]).strip()
         logger.debug(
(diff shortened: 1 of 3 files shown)


src/khoj/processor/conversation/utils.py
```diff
@@ -110,7 +110,7 @@ def generate_chatml_messages_with_context(
         logger.debug(
```
debanjum, line 110: I forgot to actually truncate the last message if the last message exceeds max prompt size! 🙈
It'd be great if you can add that here as well. 

The logic will be a little more complicated than just `encoder.decode(encoder.encode(last_message.content)[:max_prompt_size[model_name]]` as the last bit of each message contains the actual question. Sharing sample diff for reference here:

```diff
-        last_message = messages[-1]
-        truncated_message = encoder.decode(encoder.encode(last_message.content))
+        last_message = '\n'.join(messages[-1].content.split("\n")[:-1])
+        original_question = '\n'.join(messages[-1].content.split("\n")[-1:])
+        truncated_message = encoder.decode(encoder.encode(last_message)[:max_prompt_size[model_name]]).strip()
         logger.debug(
             f"Truncate last message to fit within max prompt size of {max_prompt_size[model_name]} supported by {model_name} model:\n {truncated_message}"
         )
-        messages = [ChatMessage(content=[truncated_message], role=last_message.role)]
+        messages = [ChatMessage(content=truncated_message + original_question, role=messages[-1].role)]
```

src/khoj/processor/conversation/utils.py
```diff
@@ -97,23 +97,33 @@ def generate_chatml_messages_with_context(
     messages = user_chatml_message + rest_backnforths + system_chatml_message
 
     # Truncate oldest messages from conversation history until under max supported prompt size by model
+    messages = truncate_message(messages, max_prompt_size[model_name], model_name)
+
+    # Return message in chronological order
+    return messages[::-1]
+
+
+def truncate_message(messages, max_prompt_size, model_name):
```
debanjum, line 106: Nit: naming of `truncate_messages` function
```suggestion
def truncate_messages(messages, max_prompt_size, model_name):
```

tests/test_conversation_utils.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class TestTruncateMessage:
+    max_prompt_size = 4096
+    model_name = "gpt-3.5-turbo"
+    encoder = tiktoken.encoding_for_model(model_name)
+
+    def test_truncate_message_all_small(self):
+        chat_messages = ChatMessageFactory.build_batch(500)
+        assert len(chat_messages) == 500
```
debanjum, line 22: nit: this assert seems unnecessary as it's testing the factory package rather than unit testing khoj code

tests/test_conversation_utils.py
```diff
@@ -0,0 +1,75 @@
...
+        tokens = sum([len(self.encoder.encode(message.content)) for message in chat_messages])
+        assert tokens > self.max_prompt_size
+
+        prompt = utils.truncate_message(chat_messages, self.max_prompt_size, self.model_name)
```
debanjum, line 26: nit: variable name isn't descriptive
```suggestion
        truncated_messages = utils.truncate_message(chat_messages, self.max_prompt_size, self.model_name)
```

tests/test_conversation_utils.py
sabaimran: True, fair point. Let me remove it.

debanjum (approved):
Thanks for addressing the comments. The PR looks good!

PR #225, opened 2023-06-05
Log clients calling API endpoints on Khoj server

debanjum:
- Make API endpoints on Khoj server accept `client` as request parameter
  - Khoj API endpoints: /chat, /search, /update
- Make Khoj clients set `client` request param when calling the API endpoints on the Khoj server
  - Khoj clients: Emacs, Obsidian and Web

### DIFF
--- a/src/interface/emacs/khoj.el
+++ b/src/interface/emacs/khoj.el
@@ -634,7 +634,7 @@ CONFIG is json obtained from Khoj config API."
       (buffer-string)))
   ;; Update index on khoj server after configuration update
   (let ((khoj--server-ready? nil))
-    (url-retrieve (format "%s/api/update?t=org" khoj-server-url) #'identity)))
+    (url-retrieve (format "%s/api/update?t=org&client=emacs" khoj-server-url) #'identity)))
 
 (defun khoj--get-enabled-content-types ()
   "Get content types enabled for search from API."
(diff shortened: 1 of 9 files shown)


src/interface/emacs/khoj.el
```diff
@@ -634,7 +634,7 @@ CONFIG is json obtained from Khoj config API."
       (buffer-string)))
   ;; Update index on khoj server after configuration update
   (let ((khoj--server-ready? nil))
-    (url-retrieve (format "%s/api/update?t=org" khoj-server-url) #'identity)))
+    (url-retrieve (format "%s/api/update?t=org&client=emacs" khoj-server-url) #'identity)))
```
sabaimran, line 637: Don't know enough e-lisp to tell you how you can do this, but would it make sense to have some string where you can format the values that are changing for each function? For example, in the Python way, I'd have a `url` variable as below and swap out parameters as needed.
```
url = f"/api/{function}?t=org&client=emacs"
```

Same suggestion for other endpoints for simplicity.

src/khoj/utils/helpers.py
```diff
@@ -184,6 +184,7 @@ def log_telemetry(telemetry_type: str, api: str = None, client: str = None, app_
     request_body = {
         "telemetry_type": telemetry_type,
         "server_id": get_server_id(),
+        "server_version": version("khoj-assistant"),
         "os": platform.system(),
```
sabaimran, line 188: Does the telemetry server require any schema changes to accept the new value?

debanjum: Ah, nice catch! Yes, the table schema in sqlite db would need to be updated

src/interface/emacs/khoj.el
debanjum: I'm not sure how much reuse we get from doing this. as the api and request parameters are different across the different api endpoints. Let me know if it's fine to defer looking into this deeper a little later?

src/khoj/utils/helpers.py
debanjum: 1. Manually added a `server_version` column to the new telemetry server
2. Update the telemetry server code to get and push server_version column from API to sqlite3 db when server_version available in request payload
3. Updated the deployed version of the telemetry server to use this new server code

sabaimran (approved):

PR #226, opened 2023-06-13
Add a Github plugin which can be used to read from a Github repository

sabaimran:
# Incoming
- [x] Add a Github plugin which can be used to read from a single Github repository using [llamahub](https://github.com/emptycrown/llama-hub)
  - Requires a [classic PAT token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
- [x] Update the desktop GUI application to take in the required parameters to read from Github

In the future, we should setup a really easy pipeline that drives plugin configuration to the respective UIs and autogenerates a screen to set the schema.

### DIFF
--- a/src/khoj/configure.py
+++ b/src/khoj/configure.py
@@ -16,6 +16,7 @@ from khoj.processor.jsonl.jsonl_to_jsonl import JsonlToJsonl
 from khoj.processor.markdown.markdown_to_jsonl import MarkdownToJsonl
 from khoj.processor.org_mode.org_to_jsonl import OrgToJsonl
 from khoj.processor.pdf.pdf_to_jsonl import PdfToJsonl
+from khoj.processor.github.github_to_jsonl import GithubToJsonl
 from khoj.search_type import image_search, text_search
 from khoj.utils import constants, state
 from khoj.utils.config import SearchType, SearchModels, ProcessorConfigModel, ConversationProcessorConfigModel
@@ -152,6 +153,20 @@ def configure_search(model: SearchModels, config: FullConfig, regenerate: bool, 
         model.image_search = image_search.setup(
             config.content_type.image, search_config=config.search_type.image, regenerate=regenerate
         )
+
+    if (t == state.SearchType.Github or t == None) and config.content_type.github:
+        logger.info("🐙 Setting up search for github")
+        # Extract Entries, Generate Github Embeddings
+        try:
+            model.github_search = text_search.setup(
+                GithubToJsonl,
+                config.content_type.github,
+                search_config=config.search_type.asymmetric,
+                regenerate=regenerate,
+                filters=[DateFilter(), WordFilter(), FileFilter()],
+            )
+        except Exception as e:
+            logger.error(f"Failed to setup github search: {e}")
 
     # Initialize External Plugin Search
     if (t == None or t in state.SearchType) and config.content_type.plugins:
--- a/src/khoj/configure.py
+++ b/src/khoj/configure.py
@@ -91,7 +91,7 @@ def configure_search(model: SearchModels, config: FullConfig, regenerate: bool, 
     if (t == state.SearchType.Org or t == None) and config.content_type.org:
         logger.info("🦄 Setting up search for orgmode notes")
         # Extract Entries, Generate Notes Embeddings
-        model.orgmode_search = text_search.setup(
+        model.org_search = text_search.setup(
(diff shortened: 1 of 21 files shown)


src/khoj/configure.py
```diff
@@ -48,6 +49,7 @@ def configure_server(args, required=False):
     # Initialize the search type and model from Config
     state.search_index_lock.acquire()
     state.SearchType = configure_search_types(state.config)
+    state.model = SearchModels()
```
sabaimran, line 52: I added this so that the state of models can be reset between configuration steps. This helps in case one of the configuration steps fails below, we won't send stale data.

src/khoj/interface/desktop/labelled_text_field.py
```diff
@@ -3,14 +3,18 @@
...
 class LabelledTextField(QtWidgets.QWidget):
-    def __init__(self, title, processor_type: ProcessorType = None, default_value: str = None):
+    def __init__(
+        self, title, search_type: SearchType = None, processor_type: ProcessorType = None, default_value: str = None
+    ):
         QtWidgets.QWidget.__init__(self)
         layout = QtWidgets.QHBoxLayout()
         self.setLayout(layout)
         self.processor_type = processor_type
+        self.search_type = search_type
```
debanjum, line 17: Where is this used?

src/khoj/configure.py
```diff
@@ -153,6 +155,20 @@ def configure_search(model: SearchModels, config: FullConfig, regenerate: bool,
...
+            model.github_search = text_search.setup(
+                GithubToJsonl,
+                config.content_type.github,
+                search_config=config.search_type.asymmetric,
+                regenerate=regenerate,
+                filters=[DateFilter(), WordFilter(), FileFilter()],
+            )
+        except Exception as e:
+            logger.error(f"Failed to setup github search: {e}")
+
```
debanjum, line 171: Should we just throw exception if the Github plugin setup fails? This is what we're doing for the other content types as well. It prevents unexpected application behavior from going unnoticed

src/khoj/configure.py
debanjum: Is this required because of the Github plugin setup being wrapped in a try/except below? A failure to configure the other content types should fail the application from my understanding. Should we let the Github plugin setup fail the application in case it errors out?

sabaimran: My initial thought was that it's better to fail gracefully and let the other indexing continue while making the user aware in some other medium that failure occurred. Especially for less-technical folks, might be irksome if half your functionality breaks. I'm fine either way ultimately.

src/khoj/interface/desktop/labelled_text_field.py
sabaimran: It's used in `add_github_settings_panel`. It's a new addition because the Github input is the first search_type to use the `LabelledTextFIeld`,

src/khoj/configure.py
sabaimran: But yes, correct. That is why the reset was required. Otherwise, because we fail only silently, it has cached data from the previous successful run.

src/khoj/processor/jsonl/jsonl_to_jsonl.py
```diff
@@ -41,7 +41,7 @@ def process(self, previous_entries=None):
             if not previous_entries:
                 entries_with_ids = list(enumerate(current_entries))
             else:
-                entries_with_ids = self.mark_entries_for_update(
+                entries_with_ids = TextToJsonl.mark_entries_for_update(
```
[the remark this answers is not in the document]
sabaimran: Hmmm. I'd changed it to be a static method because it made it easier to construct an integration with this method without having to inherit from this class.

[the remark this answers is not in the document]
sabaimran: In what circumstances do you foresee need for overriding `mark_entries_for_update`?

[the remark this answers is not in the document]
debanjum: Never mind, making it static makes sense. This will allow both non-child and child classes to use the `TextToJsonl.mark_entries_for_update` method. And they can of course still use their own method to `mark_entries_for_update` instead of `self.mark_entries_for_update` or `TextToJsonl.mark_entries_for_update`

PR #232, opened 2023-06-21
Search across all Asymmetric Text Content Types in Parallel

debanjum:
- Allow searching across asymmetric text content types using threads
   - Query time on my Mac averages 95ms latency (140ms at 90 percentile) across (Org, Markdown, Github, PDF and Music content types)
   - This is not too much more than search for a single content type (maybe max ~50% latency increase?). Encoding query is what takes most of the time anyway and that's just done once like before, threading adds some overhead
   - An **average** of `95 ms` latency or `140ms` at **90th percentile** is inline with keeping an incremental search (search-as-you-type) experience
- Put logic to remove filter terms from query in a `defilter` method for each filter
- Encode query once during search to encode query once across all (asymmetric) content types
- Search across all content types via the web and emacs interfaces in [d5fb419](https://github.com/khoj-ai/khoj/pull/232/commits/d5fb4196de3afefd04c72c68d2d7b121d5e88005) and [5c4eb95](https://github.com/khoj-ai/khoj/pull/232/commits/5c4eb950d53a723f124cac9ebf1a6f352a360645) respectively
- Allow Khoj Chat to pull relevant data from across content types (without the perf hit). Khoj chat is only pulling data from a single content type currently


src/khoj/interface/web/index.html
```diff
@@ -94,9 +94,12 @@
                 setQueryFieldInUrl(query);
 
             // Generate Backend API URL to execute Search
-            url = type === "image"
-                ? `/api/search?q=${encodeURIComponent(query)}&t=${type}&n=${results_count}&client=web`
-                : `/api/search?q=${encodeURIComponent(query)}&t=${type}&n=${results_count}&r=${rerank}&client=web`;
+            if (type == 'all')
```
sabaimran, line 97: nit: Can use a helper function + a string constant for the URL to create these urls. For example:

```
function createRequest(query, n_count, content_type, rerank) {
    url = `/api/search?q=${encodeURIComponent(query)}&n=${results_count}`;
    if (content_type):
        url += `&t={type}`;
    // ...add on other query parameters as required.
    return url;
}
```

src/khoj/search_filter/date_filter.py
```diff
@@ -49,6 +49,12 @@ def can_filter(self, raw_query):
         "Check if query contains date filters"
         return self.extract_date_range(raw_query) is not None
 
+    def defilter(self, query):
```
sabaimran, line 52: Neat 👀

sabaimran (approved):
Sweet! This is a game changer 🎊 .

src/khoj/configure.py
```diff
@@ -78,16 +79,20 @@ def configure_search_types(config: FullConfig):
...
         plugin_search_types = {plugin_type: plugin_type for plugin_type in config.content_type.plugins.keys()}
 
     # Dynamically generate search type enum by merging core search types with configured plugin search types
     return Enum("SearchType", merge_dicts(core_search_types, plugin_search_types))
 
 
-def configure_search(model: SearchModels, config: FullConfig, regenerate: bool, t: state.SearchType = None):
+def configure_search(model: SearchModels, config: FullConfig, regenerate: bool, t: Optional[state.SearchType] = None):
+    if config.content_type is None or config.search_type is None:
+        logger.error("🚨 Content Type or Search Type not configured.")
```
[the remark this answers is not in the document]
debanjum: During first-run the content_type isn't expected to be populated. So putting a `pydantic` validator inside FullConfig would throw earlier than is desired. Here we're only saying we'll configure search if content_type is defined but we do not throw. So calls on startup during first_run will only log the error but not throw. 

Realized 2 things though:
1. Code should actually just log warning, instead of log error as no search or content type being configured is expected during first run
2. Code should check if `config is None` as welll

sabaimran (approved):

PR #235, opened 2023-06-21
Add ReliableGPT to Khoj - Handle OpenAI Errors

ishaan-jaff:
Add https://github.com/BerriAI/reliableGPT to Khoj to handle the following errors:

- OpenAI APIError, OpenAI Timeout, OpenAI Rate Limit Errors, OpenAI Service UnavailableError / Overloaded
- Context Window Errors
- Invalid API Key errors

Khoj users can specify fallback strategies for error handling - eg. if GPT-4 is overloaded use GPT 3.5 Turbo, Anthropic etc

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ dependencies = [
     "langchain >= 0.0.187",
     "pypdf >= 3.9.0",
     "requests >= 2.26.0",
+    "reliableGPT == 0.2.6",
 ]
 dynamic = ["version"]
 
--- a/src/khoj/processor/conversation/utils.py
+++ b/src/khoj/processor/conversation/utils.py
@@ -17,6 +17,8 @@ from tenacity import (
     wait_exponential,
     wait_random_exponential,
 )
+from reliablegpt import reliableGPT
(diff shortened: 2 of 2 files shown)


ishaan-jaff:
cc @debanjum

debanjum:
Hi @ishaan-jaff, thanks for raising a PR to increase the reliability of Khoj OpenAI responses with reliableGPT (reliableGPT looks neat!). 

Unfortunately incorporating `reliableGPT` such that it actually helps out with Khoj chat reliability will require changes in terms of use-case. Because Khoj currently uses the users OpenAI API key:
1. We don't need to deal with multiple keys and key rotations etc yet
2. All users may not have access to GPT4 or Claude etc
3. We'd need different prompts for different LLMs (i.e Claude vs GPT4 vs Llama). We don't have the infra for that yet.

Having fallback strategies and working with multiple models in case of OpenAI server overload etc. sounds super useful. But Khoj will only be able to benefit from reliableGPT at a later stage once [Khoj cloud](https://khoj.dev) and using different classes of LLM within Khoj matures. 

So I'm closing this PR for now but let me know if you have any questions or concerns!

PR #236, opened 2023-06-22
Update Web Interface with Lantern Theme

debanjum:
- Style all pages with consistent lantern theme styling
  - Add navigation pane to all web interface pages
  - a200af68b38d0625c42e2098d171c6ddab121bd2 Keep pico.css locally for offline usage
  - cd8d069e6673b4db4c14f736c3d8af80bf94614d Highlight currently active tab in web interface
- Update config pages to use Lantern theme
   | Before | After |
   | -- | -- |
   | Screenshot 2023-06-17 at 3 50 08 PM | Screenshot 2023-06-22 at 00 00 14 |


src/khoj/interface/web/config.html
```diff
@@ -1,70 +1,114 @@
...
+                    <h3 class="card-title">Github</h3>
+                </div>
+                <div class="card-description-row">
+                <p class="card-description">Set repositories for Khoj to index</p>
+                </div>
+                <div class="card-action-row">
+                    <button class="card-button" onclick="window.location.href='/config/content_type/github';">
+                        Setup
+                        <svg xmlns="http://www.w3.org/2000/svg" width="1em" height="1em" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M5 12h14M12 5l7 7-7 7"></path></svg>
+                    </button>
```
sabaimran, line 20: Could we make this an `anchor` tag? It makes the navigation nicer and allow the user to `ctrl`+`click` and open in a new window. Might make the styling slightly annoying.

src/khoj/interface/web/config.html
```diff
@@ -1,70 +1,114 @@
...
             method: 'GET',
         })
         .then(response => response.json())
         .then(data => {
             console.log('Success:', data);
-            alert("Regenerated!");
+            alert("Saved!");
             regenerate.disabled = false;
-            regenerate.innerHTML = "Regenerate";
+            regenerate.innerHTML = "Save";
```
sabaimran, line 105: What if this was `Refresh` instead? I think `Save` doesn't do justice to indicating to the user that the freshly retrieved data is now available via Khoj, but I also understand that `Regenerate` is overly technical.

src/khoj/interface/web/content_type_input.html
```diff
@@ -1,75 +1,82 @@
...
-            <td>
-                <button type="button" id="input-filter-button">Add</button>
-            </td>
-        </tr>
-    </table>
+<div class="page">
+    <div class="section">
+        <h2 class="section-title">
+            <img class="card-icon" src="/static/assets/icons/{{ content_type }}.svg" alt="{{ content_type|capitalize }}">
+            <span class="card-title-text">{{ content_type|capitalize }}</span>
```
sabaimran, line 7: What is the `|capitalize` for? I'm assuming that capitalizes the text in jinja-langauge?

src/khoj/interface/web/index.html
```diff
@@ -188,9 +189,15 @@
     </script>
 
     <body>
-        <h1>
-            <img class="khoj-logo" src="static/assets/icons/khoj-logo-sideways.svg" alt="Khoj"></img>
-        </h1>
+        <!--Add Header Logo and Nav Pane-->
+        <div class="khoj-header">
+            <img class="khoj-logo" src="/static/assets/icons/khoj-logo-sideways.svg" alt="Khoj"></img>
```
sabaimran, line 194: Does this asset need to be checked in?

debanjum: That's been already checked in a previous commit

src/khoj/interface/web/config.html
debanjum: Changed it to "Configure"? I'm actually using "Configure" for the button default text. So these should anyway have read as Configure, Configuring, Configured. Let me know if that doesn't look fine

PR #253, opened 2023-06-29
Support indexing Github issues as well as corresponding comments

### LINKED ISSUE #248: Index Github Issues in Github Plugin for Khoj
Update github plugin for Khoj to index Issues from Github repository with Khoj

sabaimran:
## Incoming
- Add support for indexing all of the issues across a repository, including corresponding comments
- Use streaming and request sessions to improve the latency and memory footprint
- Improve some of the styling for links and avatar images
- Rename all the .md test files to .markdown to simplify indexing tests

Example screenshot of a result:
Screenshot 2023-06-29 at 11 29 36 AM

Closes #248


src/khoj/interface/web/assets/markdown-it.min.js
```diff
@@ -3227,6 +3227,16 @@
         result += (token.nesting === -1 ? "</" : "<") + token.tag;
     // Encode attributes, e.g. `<img src="foo"`
         result += this.renderAttrs(token);
+
```
debanjum, line 3230: 1. What is this token.tag code doing? Seems like it has to do with styling the github profile picture from the issues? Maybe we can add a comment to say what this code is about?

2. It'd be great if we can keep this code outside of `markdown-it.min.js` as that's an external js library and upgrading it later will be problematic if we modify it. If that's not easy/possible we should add a bold comment or something warning that this is custom code we've added to `markdown-it` so hopefully someone notices when we upgrade `markdown-it` next

src/khoj/processor/github/github_to_jsonl.py
```diff
@@ -201,3 +287,32 @@ def extract_org_entries(org_files):
...
+            content = issue["content"]
+            if "comments" in issue:
+                for comment in issue["comments"]:
+                    content += "\n\n" + comment["content"]
+            entries.append(content)
+            entry_to_file_map[content] = {"path": issue["path"]}
+        return entries, entry_to_file_map
+
+    @staticmethod
+    def convert_issue_entries_to_maps(parsed_entries: List[str], entry_to_file_map: Dict[str, Dict]) -> List[Entry]:
```
debanjum, line 305: Shouldn't `entry_to_file_map` be called `entry_to_location_map` or more generically `entry_to_metadata_map`?

src/khoj/processor/github/github_to_jsonl.py
```diff
@@ -130,42 +140,118 @@ def get_files(self, repo_url: str, repo: GithubRepoConfig):
...
+            response = self.session.get(comments_url, params=params, stream=True)
+            raw_comments = response.json()
+
+            # Wait for rate limit reset if needed
+            result = self.wait_for_rate_limit_reset(response, self.get_comments, comments_url)
+            if result is not None:
+                return result
+
+            for comment in raw_comments:
+                created_at = comment["created_at"].split("T")[0]
```
debanjum, line 238: I'm guessing the comment["created_at"] returns an ISO-8601 datetime string? Would it be better if we parse it as a `datetime` object and string format it out in `YYYY-MM-DD` or even `YYYY-MM-DD HH:SS` format? This would avoid us manually parsing date time string and hitting timezone bugs
```suggestion
                created_at = datetime.datetime.fromisoformat(comment["created_at"]).strftime("%Y-%m-%d %H:%M")
```

src/khoj/interface/web/index.html
```diff
@@ -370,6 +375,17 @@
             max-width: 100;
         }
 
+        a {
+            color: #3b82f6;
+            text-decoration: none;
```
debanjum, line 380: Can we use links colors (like dark yellow) to stay aligned with the new lantern theme? The blue while nice, gives it a different feel

debanjum (approved):
Neat! Using profile pictures in the rendered issue comments look great. And streaming the Github issue comments etc should improve perf/memory

src/khoj/interface/web/index.html
sabaimran: Let me experiment with this. Our default yellow was resulting in an accessibility issue with contrast, but might work for a darker color (I think).

src/khoj/interface/web/assets/markdown-it.min.js
sabaimran: Fair point. Yes, this is for adding a specific class if this is an img tag element with a src that indicates an avatar in the URL. This allows us to have that tiny profile picture in the summary comment thread. Otherwise, it blows up in full resolution and looks unpleasant.

src/khoj/interface/web/index.html
sabaimran: I'm going to say let's leave this as-is for now, for two reasons.
1. The yellowish color from our branding poses certain accessibility constraints. We can think of what our "accent" color should be (or ask someone who can actually do UX 🤣).
2. Blue for links is an extremely standard UX pattern that everyone expects, even if they don't totally realize it. Hence, it's a good starting place.

PR #260, opened 2023-07-01
Add demo view for Khoj

sabaimran:
# Incoming
- Add a separate view if the instance is meant for demo purposes. This adds a banner prompting users to sign up for the waitlist and guides them to the landing page. This view also hides the settings page.
- In theory, this will be suitable for any Khoj instance that's meant for external-facing purposes (as in, outside of the user's network)
- Prevent re-indexing for Github data if this is a demo instance
- Fix up some issues with the CSS which made settings page small in mobile
- In the frontend views for Khoj, add a button to get on the waitlist and links to the landing page

### DIFF
--- a/src/khoj/interface/web/base_config.html
+++ b/src/khoj/interface/web/base_config.html
@@ -2,6 +2,7 @@
 <html data-theme="light">
     <head>
         <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0 maximum-scale=1.0">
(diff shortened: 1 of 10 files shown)


src/khoj/interface/web/base_config.html
```diff
@@ -2,6 +2,7 @@
 <html data-theme="light">
     <head>
         <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0 maximum-scale=1.0">
```
sabaimran, line 5: This fixes the mobile rendering issue.

PR #263, opened 2023-07-02
[Fix] Remove the default behavior of using GUI for Khoj

### LINKED ISSUE #262: Change default behavior of Khoj to be no gui
Update the default behavior of Khoj to be bypassing the PyQT GUI. There was an [error reported](https://discord.com/channels/1112065956647284756/1112066447947083826/1124867250294370304) in the Discord related to the default behavior. Given we have a nicer web-based UI now, we can just direct users there.

sabaimran:
# Incoming
- Remove the default behavior of opening the desktop GUI when starting `khoj`. Instead, direct users to go to `http://localhost:8000` to manage their settings and configure their instance.
- Update corresponding documentation, unit tests, CLI configuration

Closes #262 

## Additional
- Do not run the task scheduler if this is a demo instance
- Remove bespoke logic for indexing `GithubToJsonl` in demo


tests/test_cli.py
```diff
@@ -36,11 +36,11 @@ def test_cli_invalid_config_file_path():
 # ----------------------------------------------------------------------------------------------------
 def test_cli_config_from_file():
     # Act
-    actual_args = cli(["-c=tests/data/config.yml", "--regenerate", "--no-gui", "-vvv"])
+    actual_args = cli(["-c=tests/data/config.yml", "--regenerate", "-vvv"])
```
debanjum, line 39: We should pass the `--gui` field to test if passing it results in `actual_args.gui = True`
```suggestion
    actual_args = cli(["-c=tests/data/config.yml", "--regenerate", "--gui", "-vvv"])
```

src/khoj/processor/github/github_to_jsonl.py
```diff
@@ -39,17 +39,13 @@ def wait_for_rate_limit_reset(response, func, *args, **kwargs):
             return
 
     def process(self, previous_entries=None):
-        # If demo mode is enabled, don't re-process any of the repositories. This is resource intensive.
-        if state.demo and previous_entries is not None:
-            return self.update_entries_with_ids(previous_entries, previous_entries)
-
```
debanjum, line 45: Don't we want to disable github (re-)indexing in demo mode?

src/khoj/main.py
```diff
@@ -118,6 +112,15 @@ def run():
 
         gui.exec()
 
+    if not state.demo:
```
debanjum, line 116: We only need to have a wake-up thread for scheduled jobs when khoj in "daemon" (i.e no gui) mode.
```suggestion
    if not args.gui and not state.demo:
```

src/khoj/main.py
```diff
@@ -118,6 +112,15 @@ def run():
...
+        # Setup task scheduler
+        poll_task_scheduler()
+
+    # Start Server
+    configure_server(args, required=False)
+    configure_routes(app)
+    start_server(app, host=args.host, port=args.port, socket=args.socket)
+
```
debanjum, line 124: This block should be put under `if not args.gui` to prevent it being called twice in `gui` mode. I suggest we keep the ordering of code execution same as before. The app start sequence is sensitive to change because of the multiple threads (GUI, polling thread)

```suggestion
    # Start Server
    configure_server(args, required=False)
    configure_routes(app)
    start_server(app, host=args.host, port=args.port, socket=args.socket)

```

debanjum (changes requested):

src/khoj/processor/github/github_to_jsonl.py
sabaimran: No, I don't think so. I've set it up so that the periodic scheduler should not be enabled in `demo` mode, and I think that should be sufficient. We should have a backup mechanism to trigger a re-indexing if required on the live app.

src/khoj/main.py
```diff
@@ -137,6 +141,12 @@ def start_server(app, host=None, port=None, socket=None):
     if socket:
         uvicorn.run(app, proxy_headers=True, uds=socket, log_level="debug", use_colors=True, log_config=None)
     else:
+        url = f"http://{host}:{port}"
+        logger.info(f"🌗 Khoj is running at {url}")
+        try:
+            webbrowser.open(url)
+        except:
+            logger.warning("🚧 Unable to open browser. Please open it manually to configure Khoj.")
```
debanjum, line 149: Should this only be shown in GUI mode? Obsidian and Emacs users do not need the server to open the browser. They are already on the interfaces they'll be interacting with Khoj from

src/khoj/processor/github/github_to_jsonl.py
debanjum: Ah I see. Makes perfect sense

debanjum (approved):
Looks great! Happy we've reduced the first-party UI components we've to maintain 🙂

PR #268, opened 2023-07-02
Add try-except-finally blocks around configure calls in /update

sabaimran:
# Incoming
- Add try-except-finally blocks around configure calls in `/update`. If the user's configuration isn't correct on initialization, the application still won't start up, but this will at least solve our deadlocking issue when a bad configuration is sent to `/update/`.


debanjum (approved):
Looks great! ✨

PR #270, opened 2023-07-03
Drop Support for Org Music, Ledger Content Types

debanjum:
**Please raise an issue if you actively use any of these content types with Khoj**

### Why
Removing unused content types will reduce khoj code to manage

- 0f993b3 Drop support for Ledger as a separate content type
  Khoj will soon get a generic text indexing content type in #237. 
  This along with a file filter should suffice for searching through Ledger transactions

- c9db532 Remove unused org-music as an indexable content type from Khoj
  Org-music was just a custom content type that worked with org-music.
  It was mostly only useful for me.


sabaimran (approved):
❤️ the cleanup.

PR #274, opened 2023-07-03
Skip indexing a PDF that has a loading error

sabaimran:
# Incoming
- If there's a single PDF file that has an indexing issue, the entire job fails and Khoj setup is killed. Instead, loudly log the error but continue processing.

There are additional details for the reason for this fix [in this Discord thread](https://discord.com/channels/1112065956647284756/1125330945550073876/1125550990997864550). A user was unable to index their vault due to a single failure, but also unable to pinpoint which file was causing the issue.


debanjum (approved):
Looks good. Maybe we should add this for the other content types later as well

Issue #257, opened 2023-06-30
Stream Chat results from OpenAI to Khoj UIs


debanjum:
Streaming Khoj chat response should improve responsiveness of conversing with Khoj. This should improve user experience (and hopefully reduce OpenAI API call failure rates as well)

sabaimran:
[This seems like a helpful issue](https://github.com/hwchase17/langchain/issues/2428) on langchain's end that explains some possible solutions to the streaming question.

PR #278, opened 2023-07-05
⚡️ Add chat streaming to OpenAI responses

sabaimran:
# Incoming
- Add chat streaming to the OpenAI API request. Implement using a callback function which manages a queue where new tokens can be placed as they come on. As the thread is read from, tokens are removed.
  - When the final token has been processed, add the `compiled_references` to the queue to be rendered by the `chat` client
  - When the thread has been closed, save the accumulated conversation log in the user's history using a `partial func`
  - Incrementally decode tokens on the front end and add them as they appear from the streamed response
- This significantly reduces perceived latency and will hopefully eliminate incident rate for timeouts

closes https://github.com/khoj-ai/khoj/issues/257

https://github.com/khoj-ai/khoj/assets/65192171/8a67ad26-0b0b-44a9-97a4-4483d7a9f642

### DIFF
--- a/src/interface/obsidian/src/chat_modal.ts
+++ b/src/interface/obsidian/src/chat_modal.ts
@@ -156,6 +156,10 @@ export class KhojChatModal extends Modal {
         // Get chat response from Khoj backend
         let encodedQuery = encodeURIComponent(query);
         let chatUrl = `${this.setting.khojUrl}/api/chat?q=${encodedQuery}&client=obsidian`;
+        let responseElement = this.createKhojResponseDiv();
+
+        // Temporary status message to indicate that Khoj is thinking
+        this.renderIncrementalMessage(responseElement, "🤔");
 
         let response = await fetch(chatUrl, {
             method: "GET",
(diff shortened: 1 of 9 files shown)


src/interface/obsidian/src/chat_modal.ts
```diff
@@ -1,6 +1,6 @@
 import { App, Modal, request, Setting } from 'obsidian';
 import { KhojSetting } from 'src/settings';
-
+import fetch from "node-fetch";
```
sabaimran, line 3: This should work for locally-hosted Khoj + Obsidian clients. @debanjum , I'll need to enlist your help for Emacs support 😁

src/interface/obsidian/src/chat_modal.ts
```diff
@@ -121,10 +156,30 @@ export class KhojChatModal extends Modal {
         // Get chat response from Khoj backend
         let encodedQuery = encodeURIComponent(query);
         let chatUrl = `${this.setting.khojUrl}/api/chat?q=${encodedQuery}&client=obsidian`;
-        let response = await request(chatUrl);
-        let data = JSON.parse(response);
 
-        // Render Khoj response as chat message
-        this.renderMessage(data.response, "khoj");
+        let response = await fetch(chatUrl, {
```
debanjum, line 160: We should add a thinking status message in obsidian as well

debanjum (commented):
Tested this out locally. Chat with streaming feels so much better 🌈! Makes the chat feel a lot more snappy and reduces all those OpenAI API call failures

src/interface/obsidian/src/chat_modal.ts
debanjum: Let me try add streaming to `khoj.el` for Emacs separately?

debanjum (approved):

PR #284, opened 2023-07-08
Add a plugin which allows users to index their Notion pages

sabaimran:
# Incoming
- Add a plugin which allows users to index their Notion pages, given an integration API token
- Add corresponding front-end views to update the configuration and render Notion results
- Fix `/delete` calls to update `state.model` when a particular content type is removed
- Update misc. errors around typing
- When processing chats, rather than assuming the compiled references are at the start of the note, assume they're in the middle


src/khoj/interface/web/config.html
```diff
@@ -224,40 +254,20 @@ <h3 class="card-title">
...
-        });
+        updateIndex(false, "Configured successfully!", "Unable to configure. Raise issue on Khoj Discord or Github.", configure, "Configuring...");
     });
 
     var reinitialize = document.getElementById("reinitialize");
     reinitialize.addEventListener("click", function(event) {
         event.preventDefault();
-        reinitialize.disabled = true;
-        reinitialize.innerHTML = "Reinitializing...";
+        updateIndex(true, "Reinitialized successfully!", "Unable to reinitialize. Raise issue on Khoj Discord or Github.", reinitialize, "Reinitializing...");
```
debanjum, line 263: We should prefix the loading, success and error messages with the emojis. Maybe we can prefix these messages with emojis in the `updateIndex` function itself for consistency?

src/khoj/processor/notion/notion_to_jsonl.py
```diff
@@ -0,0 +1,126 @@
...
+                            raw_content = ""
+                            curr_heading = ""
+                            for block in content["results"]:
+                                block_type = block.get("type", None)
+                                if block_type == None:
+                                    continue
+                                block_data = block[block_type]
+                                raw_content = ""
+                                if block_data.get("rich_text", None) and len(block_data["rich_text"]) > 0:
+                                    if block_type in ["heading_1", "heading_2", "heading_3"]:
```
debanjum, line 65: Are there only 3 level of headings or do we drop `heading_4` etc from being indexed?

src/khoj/processor/notion/notion_to_jsonl.py
```diff
@@ -0,0 +1,126 @@
...
+                            title, content = self.get_page_content(page_id)
+                            raw_content = ""
+                            curr_heading = ""
+                            for block in content["results"]:
+                                block_type = block.get("type", None)
+                                if block_type == None:
+                                    continue
+                                block_data = block[block_type]
+                                raw_content = ""
+                                if block_data.get("rich_text", None) and len(block_data["rich_text"]) > 0:
```
debanjum, line 64: Is the default to None required here or is something like below fine?
```suggestion
                                if block_data.get("rich_text") and len(block_data["rich_text"]) > 0:
```

src/khoj/processor/notion/notion_to_jsonl.py
```diff
@@ -0,0 +1,126 @@
...
+                        elif p_or_d["object"] == "page":
+                            page_id = p_or_d["id"]
+                            title, content = self.get_page_content(page_id)
+                            raw_content = ""
+                            curr_heading = ""
+                            for block in content["results"]:
+                                block_type = block.get("type", None)
+                                if block_type == None:
+                                    continue
+                                block_data = block[block_type]
```
debanjum, line 62: Given we seem to only ever use block_data["rich_text"], should we just try extract that directly here?
```suggestion
                                block_rich_text = block[block_type].get("rich_text")
```

src/khoj/processor/notion/notion_to_jsonl.py
```diff
@@ -0,0 +1,126 @@
...
+                        if p_or_d["object"] == "database":
+                            # TODO: Handle databases
+                            continue
+                        elif p_or_d["object"] == "page":
+                            page_id = p_or_d["id"]
+                            title, content = self.get_page_content(page_id)
+                            raw_content = ""
+                            curr_heading = ""
+                            for block in content["results"]:
+                                block_type = block.get("type", None)
```
debanjum, line 59: Don't think we need to default to None, given `block.get("type")` will return `None` if block doesn't have "type" key or the value associated with `block["type"]` is `None`
```suggestion
                                block_type = block.get("type")
```

src/khoj/processor/notion/notion_to_jsonl.py
```diff
@@ -0,0 +1,126 @@
...
+                                if block_type == None:
+                                    continue
+                                block_data = block[block_type]
+                                raw_content = ""
+                                if block_data.get("rich_text", None) and len(block_data["rich_text"]) > 0:
+                                    if block_type in ["heading_1", "heading_2", "heading_3"]:
+                                        if raw_content != "":
```
debanjum, line 66: Can we add comment to explain the logic. 

Seems like `raw_content != null` right after `block_type in heading_{1,2,3}` indicates we've moved to next heading and can dump the previous headings contents into an Entry?

src/khoj/processor/notion/notion_to_jsonl.py
```diff
@@ -0,0 +1,126 @@
...
+                                                    file=p_or_d["url"],
+                                                )
+                                            )
+                                        curr_heading = block_data["rich_text"][0]["plain_text"]
+                                    if curr_heading != "":
+                                        raw_content = curr_heading + "\n"
+                                    for text in block_data["rich_text"]:
+                                        raw_content += text["plain_text"]
+                                        raw_content += "\n"
+                                if raw_content != "":
```
debanjum, line 81: Also it'd be great if we can add a comment here, specifying that this should add any last heading into the current_entries list
```suggestion
                                // Append the last entry, if any, from Notion for indexing
                                if raw_content != "":
```

src/khoj/utils/rawconfig.py
```diff
@@ -148,4 +153,9 @@ def __repr__(self) -> str:
 
     @classmethod
     def from_dict(cls, dictionary: dict):
-        return cls(raw=dictionary["raw"], compiled=dictionary["compiled"], file=dictionary.get("file", None))
+        return cls(
+            raw=dictionary["raw"],
+            compiled=dictionary["compiled"],
+            file=dictionary.get("file", None),
+            heading=dictionary.get("heading", None),
```
debanjum, line 160: Great catch! The heading wasn't getting returned in the API responses

debanjum (approved):
Notion integration is going to be super useful for a lot of folks! 🙏🏾🌈

src/khoj/processor/notion/notion_to_jsonl.py
sabaimran: There are only three levels of headings thankfully 😅. As a writer, I always like having additional layers of nesting. But as a developer, I'm thankful that it gets cut off at some point.

PR #287, opened 2023-07-09
Make streaming optional for the /chat endpoint

sabaimran:
# Incoming
I'd previously split the `chat` endpoint into a separate call just for history and a separate call which streamed the response from OpenAI.
Some clients will not be able to received a streamed response. For them, keep the non-streamed response which returns a JSON object. Update the khoj.el to use the new paradigm, as well as web and Obsidian interfaces.
By default, `stream` is set to `False` and the API will return a JSON object with the raw response text and the compiled references.

Closes https://github.com/khoj-ai/khoj/issues/288

### DIFF
--- a/src/khoj/routers/api.py
+++ b/src/khoj/routers/api.py
@@ -482,7 +482,7 @@ async def chat(
 
     actual_response = aggregated_gpt_response.split("### compiled references:")[0]
 
-    response_obj = {"response": actual_response, "references": compiled_references}
+    response_obj = {"response": actual_response, "context": compiled_references}
 
     return Response(content=json.dumps(response_obj), media_type="application/json", status_code=200)
 
--- a/src/khoj/routers/api.py
+++ b/src/khoj/routers/api.py
@@ -5,9 +5,7 @@ import time
 import yaml
 import logging
 import json
-from datetime import datetime
 from typing import List, Optional, Union
-from functools import partial
 
 # External Packages
 from fastapi import APIRouter, HTTPException, Header, Request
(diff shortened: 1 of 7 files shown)


src/khoj/routers/api.py
```diff
@@ -444,33 +438,56 @@ def chat_init(
...
+    )
+
+    if gpt_response is None:
+        return Response(content="No response from GPT", media_type="text/plain", status_code=500)
+
+    if stream:
+        return StreamingResponse(gpt_response, media_type="text/event-stream", status_code=200)
+
+    if isinstance(gpt_response, str):
+        return Response(content=gpt_response, media_type="text/plain", status_code=200)
```
debanjum, line 474: Shouldn't this result in status_code=500 as `gpt_response` is string only on error? If so, can this be merged with the above `gpt_response is None` conditional?

src/khoj/routers/api.py
```diff
@@ -480,17 +497,17 @@ def _save_to_conversation_log(
             status_code=500, detail="Set your OpenAI API key via Khoj settings and restart it to use Khoj Chat."
         )
 
+
+async def _get_top_k_results(
```
debanjum, line 501: Maybe we can just extract search queries in this function and call it `extract_search_queries` or some such for better modularity (and naming). The code snippets that calls the `search` function can then just be in the main `chat` function?

debanjum (approved):

src/khoj/routers/api.py
sabaimran: Very good point.

debanjum (approved):

PR #289, opened 2023-07-09
Fix Packaging the Khoj Desktop Apps

debanjum:
- The desktop apps have been broken because of missing imports and restructuring of the codebase. - This PR fixes these issues to allow running Khoj using a desktop app


sabaimran (approved):
Have tested this out locally and verified that the Mac app at least will work.

PR #292, opened 2023-07-10
Allow using configured GPT chat model

tjsousa:
My account doesn't have gpt-4 enabled and it wouldn't work as the default value was always used from extract_questions, where the caller could use the configured model.

### DIFF
--- a/src/khoj/routers/api.py
+++ b/src/khoj/routers/api.py
@@ -535,6 +535,7 @@ async def extract_references_and_questions(
 
     # Initialize Variables
     api_key = state.processor_config.conversation.openai_api_key
+    chat_model = state.processor_config.conversation.chat_model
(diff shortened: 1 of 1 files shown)


sabaimran:
Hey @tjsousa , have you intentionally disabled GPT4? Open AI made GPT4 [generally available](https://openai.com/blog/gpt-4-api-general-availability) recently, so we assumed everyone with an API key should have access.

sabaimran:
Yeah, upon further thought I think that makes sense. Thanks for raising a PR for the issue.

If you don't have it, it might be affecting many people other people as well. I'll accept this change after a bit of testing 🙂.

sabaimran:
Looks good! I'll merge this in for the time being while availability is being rolled out. Main thing to note is that the portion of the code to extract the questions is going to be less good in quality, but it's preferable that more people be able to use it.

Issue #190, opened 2023-03-20
Periodic index corruption


sabaimran:
I'm not sure about the exact characterization of when this happens, but it seems to be when I've been running the `khoj` server for a while. It seems like the index frequently becomes corrupted and requires a full regeneration. I would guess that this pertains to the temporal incremental indexing functionality.

debanjum:
Thanks for opening an issue to track this! This is [a known issue](https://github.com/debanjum/khoj#search-starts-giving-wonky-results). But I haven't been able to get to debugging this yet. 

We have a reference test set from #186. Just need to stress test khoj using that (by running search/chat in parallel) and (hopefully) reproduce and fix the index corruption bug

jh0274:
hi! so i ran into a corruption of the .gz markdown embeddings file, here are the logs:

` 📩 Saved current chat session to conversation logs                              
configure.py:229[11:44:53] INFO     📬 Updating content index via Scheduler                                         configure.py:115[11:44:54] INFO     💎 Setting up search for markdown notes                                           
indexer.py:255           INFO     127.0.0.1:59973 - "OPTIONS /api/v1/index/update?force=false&client=obsidian      h11_impl.py:431                    HTTP/1.1" 200
          ERROR    Exception in callback _ProactorBasePipeTransport._call_connection_lost(None) base_events.py:1771                    handle: <Handle _ProactorBasePipeTransport._call_connection_lost(None)>
                    ╭─────────────────── Traceback (most recent call last) ────────────────────╮
                    │ C:\Python311\Lib\asyncio\events.py:80 in _run                            │
                    │                                                                          │
                    │    77 │                                                                  │
                    │    78 │   def _run(self):                                                │
                    │    79 │   │   try:                                                       │
                    │ ❱  80 │   │   │   self._context.run(self._callback, *self._args)         │
                    │    81 │   │   except (SystemExit, KeyboardInterrupt):                    │
                    │    82 │   │   │   raise                                                  │
                    │    83 │   │   except BaseException as exc:                               │
                    │                                                                          │
                    │ C:\Python311\Lib\asyncio\proactor_events.py:165 in _call_connection_lost │
                    │                                                                          │
                    │   162 │   │   │   # just close our end.  First calling shutdown() seems  │
                    │   163 │   │   │   # cure it, but maybe using DisconnectEx() would be bet │
                    │   164 │   │   │   if hasattr(self._sock, 'shutdown') and self._sock.file │
[...]

PR #300, opened 2023-07-11
Modify hash_func in text_to_jsonl.py

mzamini92:
The `mark_entries_for_update` method uses the `hash_func` function to hash the `compiled` field of each entry. However, the `hash_func` function is not deterministic, which means that it can return different hash values for the same entry. This means that the `mark_entries_for_update` method might not correctly identify new and existing entries. This code uses the `hashlib.sha256` function to hash the compiled field of each entry. This hash function is deterministic, which means that it will always return the same hash value for the same entry. This will ensure that the `mark_entries_for_update` method correctly identifies new and existing entries.

### DIFF
--- a/src/khoj/processor/text_to_jsonl.py
+++ b/src/khoj/processor/text_to_jsonl.py
@@ -62,34 +62,35 @@ class TextToJsonl(ABC):
 
     @staticmethod
     def mark_entries_for_update(
-        current_entries: List[Entry], previous_entries: List[Entry], key="compiled", logger: logging.Logger = None
-    ) -> List[Tuple[int, Entry]]:
-        # Hash all current and previous entries to identify new entries
-        with timer("Hash previous, current entries", logger):
-            current_entry_hashes = list(map(TextToJsonl.hash_func(key), current_entries))
-            previous_entry_hashes = list(map(TextToJsonl.hash_func(key), previous_entries))
+            current_entries: List[Entry], previous_entries: List[Entry], key="compiled", logger: logging.Logger = None
+        ) -> List[Tuple[int, Entry]]:
+            # Hash all current and previous entries to identify new entries
+            with timer("Hash previous, current entries", logger):
+                current_entry_hashes = list(map(hashlib.sha256, map(getattr, current_entries, [key])))
+                previous_entry_hashes = list(map(hashlib.sha256, map(getattr, previous_entries, [key])))
 
-        with timer("Identify, Mark, Combine new, existing entries", logger):
-            hash_to_current_entries = dict(zip(current_entry_hashes, current_entries))
-            hash_to_previous_entries = dict(zip(previous_entry_hashes, previous_entries))
+            with timer("Identify, Mark, Combine new, existing entries", logger):
+                hash_to_current_entries = dict(zip(current_entry_hashes, current_entries))
+                hash_to_previous_entries = dict(zip(previous_entry_hashes, previous_entries))
 
-            # All entries that did not exist in the previous set are to be added
-            new_entry_hashes = set(current_entry_hashes) - set(previous_entry_hashes)
-            # All entries that exist in both current and previous sets are kept
-            existing_entry_hashes = set(current_entry_hashes) & set(previous_entry_hashes)
+                # All entries that did not exist in the previous set are to be added
+                new_entry_hashes = set(current_entry_hashes) - set(previous_entry_hashes)
+                # All entries that exist in both current and previous sets are kept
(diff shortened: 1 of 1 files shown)


sabaimran:
Whoa, amazing find, @mzamini92! If I'm right, I think this may solve one of our biggest issues, #190. Let me do a bit of testing, but this looks really good. Thanks a ton!

src/khoj/processor/text_to_jsonl.py
```diff
@@ -62,34 +62,35 @@ def split_entries_by_max_tokens(
...
-
-            existing_entries_sorted = sorted(existing_entries, key=lambda e: e[0])
-            entries_with_ids = existing_entries_sorted + new_entries
-
-        return entries_with_ids
+            current_entries: List[Entry], previous_entries: List[Entry], key="compiled", logger: logging.Logger = None
+        ) -> List[Tuple[int, Entry]]:
+            # Hash all current and previous entries to identify new entries
+            with timer("Hash previous, current entries", logger):
+                current_entry_hashes = list(map(hashlib.sha256, map(getattr, current_entries, [key])))
```
sabaimran, line 69: Hey @mzamini92 , could you actually explain this change a bit more? We are currently using a function that wraps around `hashlib.md5`, which should provide deterministic output. Could you explain the rationale for this change, and what you've observed?

mzamini92: I couldnt follow where issue #190 reproduction error is. Can you give me the link to check if it can be solved here or not? But the periodic index corruption is caused by the fact that the hash function is not deterministic. This means that the same entry can hash to different values depending on the order in which the hash function is called. This can lead to entries being incorrectly identified as new or existing, which can cause the index to become corrupted. The `hashlib.sha256` function is deterministic, which means that it will always return the same hash value for the same entry. This is important because it ensures that the `mark_entries_for_update` method correctly identifies new and existing entries. However, the `hashlib.sha256` function is also collision-resistant, which means that it is very difficult to find two different inputs that produce the same hash value. This is important because it reduces the likelihood of errors when identifying new entries. In short, both determinism and collision resistance are important properties of a hash function for the `mark_entries_for_update` method. The `hashlib.sha256` function has both of these properties, which is why it is a good choice for this method.

sabaimran: Ah, I don't have instructions for reproduction of that error. It seems to happen periodically and it's difficult to track the root cause. It just *smells* like these could've been related.

I understand on surface level the argument you're making, but I  have the following questions:
1. What is the basis for the claim that the current function is not deterministic? Have you observed that with some sample data yourself?
2. We are currently using the `md5` hashing function. Why would using `sha256` specifically solve this problem over using `md5`? According to documentation, both should be deterministic.

mzamini92: It seems both MD5 and SHA256 are deterministic hash functions(except previously I saw somewhere which I cant find it someone had such a problem in his code and mentioned SHA256 was more robust). However according to documentation, SHA256 is a stronger hash function than MD5 and it is less likely to produce collisions. MD5 is the fastest cryptographic algorithm. SHA-256 is about 20% slower. SHA-256 is not perfect, but is a bit safer than MD5. It produces a 256-bit hash, which is twice longer as MD5. Also, the collision probability with SHA-256 is lower than with MD5. Have you also been experiencing the index corruption issue? Is that what prompted the code change? And is your corruption consistent (e.g. at every time the index is updated)? I had but as you said it randomly happened twice but so far as long as I changed it to sha256 I hadn't face it.

sabaimran: Ah, I see. This is very interesting data to have! How long has it been since you switched to using `sha256` in your local repository? Do you have your Khoj instance always on? Is your data source frequently updating? Sorry for the array of questions, but this will better help diagnose the issue, especially is switching the hashing algorithm has solved it for you.

The rationale behind going with `md5` is that speed is relevant here from an application perspective. Especially as the size of the data source increases, we need a performant mechanism to retain freshness. If, in theory, `md5` and `sha256` should both provide us deterministic hashing over the raw content, then it makes sense to go with whichever is faster, especially as security is not a concern here.

sabaimran: Hey @mzamini92 ! I'm going to close this one out. We made a couple of improvements that should mitigate related indexing issues in this PR https://github.com/khoj-ai/khoj/pull/325. Changes required ended up being quite a bit deeper than expected. Let me know if you continue to see other issues and I can follow-up.

Thank you for the PR! 💛

PR #304, opened 2023-07-12
Migrate from PyQT6 to PySide6

debanjum:


src/khoj/main.py
```diff
@@ -15,8 +15,8 @@
 # External Packages
 import uvicorn
 from fastapi import FastAPI
-from PyQt6 import QtWidgets
-from PyQt6.QtCore import QThread, QTimer
+from PySide6 import QtWidgets
+from PySide6.QtCore import QThread, QTimer
```
sabaimran, line 19: Is this really all you have to do? 😮

sabaimran (approved):

src/khoj/main.py
debanjum: Surprisingly, yes. It seems like an actual drop in replacement for our use-case

PR #320, opened 2023-07-14
Start Khoj even when config is not valid

### LINKED ISSUE #313: Allow users to start Khoj even if config or the underlying data is in a bad state
This should particularly be helpful for users when they've botched a configuration, especially in the scenario where they're using the desktop application. For most users, it'll be difficult to go into the .yml and update the underlying configuration.

sabaimran:
Currently, Khoj forces the configuration to be valid and all the embeddings to load before it starts the server. This is an unfeasible model for the Desktop application model, where the user may certainly leave their configuration in a bad state on last run. In addition to that, user generally won't be technical enough to manually edit a yaml file.

Add an icon as well that indicates that a content type either hasn't had embeddings generated yet or has an invalid configuration.

Any configuration errors will still be logged to the terminal, which should help with debugging.

Closes #313


debanjum (approved):
Looks good!

PR #329, opened 2023-07-18
Create Schema Migrator and Reindex to Apply Index Corruption Fixes

debanjum:
### Updates
- 83e1088 Manage `khoj.yml` config migrations on app start. Version the `khoj.yml` schema
- 429e1b4 Regenerate index to apply corruption fixes on first run of this khoj version
   Otherwise users would need to manually re-index their contents with khoj


sabaimran (approved):
Longer term, we'll want to think about how we can standardize schema migrations in a way where they're easy to extend and always backwards compatible with the different versions users are on. For now, this is a good first migration 🙂

PR #330, opened 2023-07-19
Add support for our first Local LLM 🤖🏠

### LINKED ISSUE #201: Support for Offline Chat Models
Hello! Could you provide support for a way to replace openai llm with locally hosted custom model calls? This would be great! (we just need a way to provide string with a response to your plugin)

Cheers!

### LINKED ISSUE #242: Offline Chat
https://github.com/tatsu-lab/stanford_alpaca#data-generation-process

https://github.com/eugeneyan/open-llms

### LINKED ISSUE #346: Rename brand name variables
- for ex: `open-ai` -> `openai`, `gpt4_all` -> `gpt4all`

sabaimran:
# Incoming
Add support for our first local LLM. This required a fairly large lift to incorporate, with schema changes required in the `/conversation` codepath. It should be easier to support more LLMs in the future with the refactor.

Start with support for Falcon7B via Nomic's GPT4All: https://github.com/nomic-ai/gpt4all. As the ecosystem grows, it should be easy to swap out the model and use a better one down the road. Falcon is the best because it's available for commercial use and can still run on consumer-grade hardware. While fantastic that it can run completely offline, its ability to assist falls far short of the GPT suite.

Add a migration script for using the new schema. By default, do not enable any LLMs. Have the user explicitly opt-in to using Falcon.

Closes #242, Closes #201, Closes #346.

### Notes

We are not going to support a bring your own model configuration for the time being, because we want to maintain a high bar for assistance/usefulness of Khoj. Of course, if you want to fork and plug in a different model, that is perfectly okay.


src/khoj/processor/conversation/utils.py
```diff
@@ -1,35 +1,19 @@
...
 
 # Internal Packages
-from khoj.utils.helpers import merge_dicts
 import queue
-
+from khoj.utils.helpers import merge_dicts
 
 logger = logging.getLogger(__name__)
-max_prompt_size = {"gpt-3.5-turbo": 4096, "gpt-4": 8192}
+max_prompt_size = {"gpt-3.5-turbo": 4096, "gpt-4": 8192, "text-davinci-001": 910}
```
debanjum, line 16: why not call use "falcon" as the key for the max_prompt size for falcon?
```suggestion
max_prompt_size = {"gpt-3.5-turbo": 4096, "gpt-4": 8192, "falcon": 910}
```

src/khoj/processor/conversation/prompts.py
```diff
@@ -18,6 +18,33 @@
 """.strip()
 )
 
+general_conversation_falcon = PromptTemplate.from_template(
+    """
+Using your general knowledge and our past conversations as context, answer the following question.
+### Instruct:\n{query}\n
```
debanjum, line 24: Given the prompt template is triple quoted shouldn't `{query}` just be put on the next line directly instead of using `\n`? Similar comment for the other prompt templates for falcon

```suggestion
### Instruct:
{query}

```

src/khoj/migrations/migrate_processor_config_openai.py
```diff
@@ -0,0 +1,63 @@
...
+
+    current_open_ai_api_key = raw_config["processor"]["conversation"].get("openai-api-key", None)
+    current_chat_model = raw_config["processor"]["conversation"].get("chat-model", None)
+    if current_open_ai_api_key is None and current_chat_model is None:
+        return args
+
+    conversation_logfile = raw_config["processor"]["conversation"].get("conversation-logfile", None)
+
+    raw_config["processor"]["conversation"] = {
+        "openai-ai-model": {"chat-model": current_chat_model, "openai-api-key": current_open_ai_api_key},
```
debanjum, line 59: Shouldn't this just be `openai` or `api-key`? If true, other parts of this migration script will need to be updated as well

```suggestion
        "openai": {"chat-model": current_chat_model, "api-key": current_open_ai_api_key},
```

src/khoj/processor/conversation/gpt4all/chat_model.py
```diff
@@ -0,0 +1,137 @@
...
+                if chat["by"] == "khoj"
+            ]
+        )
+
+    prompt = prompts.extract_questions_falcon.format(
+        chat_history=chat_history,
+        text=text,
+    )
+    message = prompts.general_conversation_falcon.format(query=prompt)
+    response = gpt4all_model.generate(message, max_tokens=200, top_k=2)
```
debanjum, line 54: Wouldn't a `top_k=2` slow the generation down? Is this done to improve the quality of the question extraction response?

src/khoj/processor/conversation/gpt4all/chat_model.py
```diff
@@ -0,0 +1,137 @@
...
+def converse_falcon(
+    references,
+    user_query,
+    conversation_log={},
+    model: str = "ggml-model-gpt4all-falcon-q4_0.bin",
+    loaded_model: Union[GPT4All, None] = None,
+    completion_func=None,
+) -> ThreadedGenerator:
+    """
+    Converse with user using OpenAI's ChatGPT
```
debanjum, line 85: The function comment should be fixed
```suggestion
    Converse with user using Falcon
```

src/khoj/processor/conversation/gpt4all/chat_model.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+from khoj.processor.conversation.utils import ThreadedGenerator, generate_chatml_messages_with_context
+from khoj.processor.conversation import prompts
+from khoj.utils.constants import empty_escape_sequences
+
+logger = logging.getLogger(__name__)
+
+
+def extract_questions_falcon(
```
debanjum, line 19: Is Falcon able to extract questions from the user's message? And in valid json format? It maybe hard as GPT-3.5 has trouble with passing back well formated json as well (which is one of the things OpenAI's [function calling](https://platform.openai.com/docs/guides/gpt/function-calling) was created to fix)

sabaimran: It's not able to do so. I've left the method in because I think the interface of our different models should implement these functions, but by default it just does some minimal string manipulation to extract the literal questions from the user's query. For example, `"Where did I go to school? Which team did I play on?"` would be separated into `["Where did I go to school?", "Which team did I play on?"]`.

src/khoj/interface/web/config.html
```diff
@@ -200,26 +200,51 @@ <h3 class="card-title">
...
+            <div class="card">
+                <div class="card-title-row">
+                    <img class="card-icon" src="/static/assets/icons/chat.svg" alt="Chat">
+                    <h3 class="card-title">
+                        Offline Chat
+                        <img id="configured-icon-conversation-enable-local-llm" class="configured-icon {% if current_config.processor and current_config.processor.conversation and current_config.processor.conversation.enable_local_llm %}enabled{% else %}disabled{% endif %}" src="/static/assets/icons/confirm-icon.svg" alt="Configured">
+                    </h3>
+                </div>
+                <div class="card-description-row">
+                <p class="card-description">Setup offline chat with Falcon</p>
```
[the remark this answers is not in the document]
sabaimran: It's not strictly required, but I think it's a nice hint for our more technical users who can give feedback around models.

src/khoj/routers/api.py
```diff
@@ -233,21 +244,26 @@ async def set_content_config_data(
...
         client: Optional[str] = None,
     ):
         _initialize_config()
 
-        state.config.processor = ProcessorConfig(conversation=updated_config)
+        if not state.config.processor or not state.config.processor.conversation:
+            conversation_logfile = resolve_absolute_path("~/.khoj/processor/conversation/conversation_logs.json")
+            state.config.processor = ProcessorConfig(conversation=ConversationProcessorConfig(conversation_logfile=conversation_logfile))  # type: ignore
+
+        assert state.config.processor.conversation is not None
```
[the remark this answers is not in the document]
sabaimran: It does, but this gets rid of the red squiggles in my IDE 👀

src/khoj/processor/conversation/utils.py
sabaimran: The `tiktoken` library has a specific set of models it works with. I just used this as a hack so we can specify a single model name when doing prompt truncation with Falcon.

src/khoj/processor/conversation/prompts.py
sabaimran: Fair point. I thought these were literally the same, but perhaps it's not with string literals. In any case, I'll just standardize the formatting.

src/khoj/processor/conversation/gpt4all/chat_model.py
sabaimran: See documentation for `top_k` parameter [here](https://docs.gpt4all.io/gpt4all_python.html#chatting-with-gpt4all).

It samples from next `k` predicted tokens in the response. I do think it helps with output quality, though I can't say for certain.

debanjum:
Just a note: Offline chat would only be available via Web interface, not from within Emacs or Obsidian with this change. Another PR will be required to enable that

debanjum (approved):

debanjum (approved):
Looks great! Excited to try the offline chat out

PR #343, opened 2023-07-26
Fix import issues in desktop image builds

sabaimran:
# Incoming
We started seeing a couple of issues in desktop image builds. I tested out the following changes on a new remote Windows machine for building an .exe file. I'm not entirely sure why this stopped working, but hypothesis is that it's related to `Pyside` or moving down the `import` statements for `Pyside`-related libraries.

Wasn't able to test using GH actions due to https://github.com/actions/upload-artifact/issues/417


Khoj.spec
```diff
@@ -19,6 +19,9 @@ datas += copy_metadata('filelock')
 datas += copy_metadata('numpy')
 datas += copy_metadata('tokenizers')
 datas += copy_metadata('pillow')
+datas += copy_metadata('huggingface_hub')
+datas += copy_metadata('safetensors')
+datas += copy_metadata('pyyaml')
```
debanjum, line 24: So once we `copy_metadata` of these dependencies, the khoj exe for windows runs fine now?

debanjum (approved):
Thanks for fixing this and testing it out!

Khoj.spec
sabaimran: Correct - after adding these lines, both Windows and Linux desktop builds were unblocked.

src/khoj/main.py
```diff
@@ -75,7 +75,8 @@ def run():
         start_server(app, host=args.host, port=args.port, socket=args.socket)
     else:
         from PySide6 import QtWidgets
-        from PySide6.QtCore import QThread, QTimer
+        from PySide6.QtCore import QTimer
```
sabaimran, line 78: Removed unused import.

Issue #347, opened 2023-07-27
Use llamav2 instead of falcon as open source model for Khoj


sabaimran:
With some adjustment as described in https://github.com/nomic-ai/gpt4all/issues/1227, it shouldn't be too hard to get [llamav2](https://about.fb.com/news/2023/07/llama-2/) up and working in place of Falcon as the first-party OS model supported by Khoj.

This should be a much better first addition as the quality and performance is much better with llamav2.

PR #352, opened 2023-07-27
Replace Falcon 🦅 model with Llama V2 🦙 for offline chat

sabaimran:
# Incoming
- Rather than release the first local LLM with Falcon, use Llama with the existing `GPT4All` setup.
- Rework the `extract_questions` flow. While Falcon's output wasn't high enough quality to actually use this functionality, Llama generally provides good enough responses (at least able to pass ~50% of the benchmark tests). That being said, query times might be too long to justify using this.
- Rename all references from `Falcon` -> `Llama`
- Update relevant unit tests to match Llama capability

Closes https://github.com/khoj-ai/khoj/issues/347

### DIFF
--- a/src/khoj/routers/api.py
+++ b/src/khoj/routers/api.py
@@ -716,7 +716,7 @@ async def extract_references_and_questions(
             else:
                 loaded_model = state.processor_config.conversation.gpt4all_model.loaded_model
                 inferred_queries = extract_questions_llama(
-                    q, loaded_model=loaded_model, conversation_log=meta_log, use_history=True
+                    q, loaded_model=loaded_model, conversation_log=meta_log, should_extract_questions=False
                 )
 
         # Collate search results as context for GPT
--- a/src/khoj/routers/api.py
+++ b/src/khoj/routers/api.py
@@ -38,7 +38,7 @@ from khoj.utils.yaml import save_config_to_file_updated_state
 from fastapi.responses import StreamingResponse, Response
 from khoj.routers.helpers import perform_chat_checks, generate_chat_response, update_telemetry_state
 from khoj.processor.conversation.openai.gpt import extract_questions
-from khoj.processor.conversation.gpt4all.chat_model import extract_questions_llama, converse_llama
+from khoj.processor.conversation.gpt4all.chat_model import extract_questions_llama
 from fastapi.requests import Request
 
 
--- a/src/khoj/routers/api.py
+++ b/src/khoj/routers/api.py
@@ -715,7 +715,9 @@ async def extract_references_and_questions(
                 inferred_queries = extract_questions(q, model=chat_model, api_key=api_key, conversation_log=meta_log)
             else:
                 loaded_model = state.processor_config.conversation.gpt4all_model.loaded_model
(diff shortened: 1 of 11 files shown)


tests/test_gpt4all_chat_actors.py
```diff
@@ -73,25 +91,26 @@ def test_extract_question_with_date_filter_from_relative_year(loaded_model):
...
-    response = extract_questions_falcon("Is Morpheus taller than Neo?", loaded_model=loaded_model, run_extraction=True)
+    response = extract_questions_llama("Is Morpheus taller than Neo?", loaded_model=loaded_model)
 
     # Assert
     expected_responses = [
-        ("morpheus", "neo"),
+        ("morpheus", "neo", "height", "taller", "shorter"),
     ]
-    assert len(response) == 2
+    assert len(response) == 3
```
debanjum, line 113: Shouldn't this extract just two questions to search for in the knowledge base and arrive at final answer?
1. What is Morpheus's height?
2. What is Neo's height?

tests/test_gpt4all_chat_actors.py
```diff
@@ -137,17 +157,24 @@ def test_generate_search_query_using_answer_from_chat_history(loaded_model):
...
+        "Vader",
+        "daughter",
+    ]
+
     # Assert
-    assert len(response) == 1
-    assert "Leia" in response[0]
+    assert len(response) >= 1
+    assert any([expected_response in response[0] for expected_response in expected_responses]), (
+        "Expected chat actor to ask for clarification in response, but got: " + response[0]
```
debanjum, line 176: Is this assert failure reason correct?

src/khoj/routers/api.py
```diff
@@ -715,7 +715,9 @@ async def extract_references_and_questions(
                 inferred_queries = extract_questions(q, model=chat_model, api_key=api_key, conversation_log=meta_log)
             else:
                 loaded_model = state.processor_config.conversation.gpt4all_model.loaded_model
-                inferred_queries = extract_questions_falcon(q, loaded_model=loaded_model, conversation_log=meta_log)
+                inferred_queries = extract_questions_llama(
```
debanjum, line 718: should we just call this `extract_questions_offline` instead of changing it to name of every new offline chat model we may migrate to?
```suggestion
                inferred_queries = extract_questions_offline(
```

src/khoj/routers/helpers.py
```diff
@@ -111,7 +111,7 @@ def _save_to_conversation_log(
                 )
             else:
                 loaded_model = state.processor_config.conversation.gpt4all_model.loaded_model
-                chat_response = converse_falcon(
+                chat_response = converse_llama(
```
debanjum, line 114: should we just call this converse_offline instead of changing it to name of every new offline chat model we may migrate to?
```suggestion
                chat_response = converse_offline(
```

debanjum (approved):
Nice!! Feels like Llama v2 may just make offline chat useful enough 🤞🏾🏕️

tests/test_gpt4all_chat_actors.py
sabaimran: To compensate for the model's unreliability, I have it always returning the user's original question in the list of extracted questions. so the results will always be `1 + len(extracted_questions)`.

PR #354, opened 2023-07-28
Fix #353: Remove references to localhost:8000

### LINKED ISSUE #353: Remove references to localhost:8000
Default port was switched to localhost:42110 a while back, so some old references in documentation need to be updated.

ducksblock:
#### This fixes #353 

Updated the localhost:8000 port references in docs to the new default port localhost:42110.


debanjum:
Thanks for raising a PR to fix this!

PR #359, opened 2023-07-28
Add Offline chat to Obsidian

sabaimran:
# Incoming
- Add support for configuring/using offline chat (with Llama V2) from within Obsidian
- Fix type checking for search type

Closes https://github.com/khoj-ai/khoj/issues/357

PR #361, opened 2023-07-28
Configure using Offline Chat from Emacs

### LINKED ISSUE #358: Add offline chat support to Emacs
We added support for Llama V2 in Khoj with #352 , but it's still not accessible through Obsidian/Emacs. Add support for in-app experience.

Relates to #357

debanjum:
Allow enabling, disabling offline chat from Emacs during khoj setup

- Chat Options
  - Enable offline chat with `(setq khoj-chat-offline t)` during khoj setup
    - Offline chat models are better for privacy but not great at answering questions
  - Enable online chat by setting your OpenAI API key with `(setq khoj-openai-api-key <your-openai-api-key)`
    - Online chat models are better at answering but require internet

- Note: *Enabling offline chat will result in large download of Llama v2 7B model on your Khoj backend*

Related #330, #352, #359
Closes #358

### DIFF
--- a/src/interface/emacs/khoj.el
+++ b/src/interface/emacs/khoj.el
@@ -231,6 +231,11 @@ for example), set this to the full interpreter path."
   :type 'string
   :group 'khoj)
 
+(defcustom khoj-chat-offline nil
+  "Use offline model to chat with khoj."
+  :type 'boolean
+  :group 'khoj)
+
 (defcustom khoj-auto-setup t
   "Automate install, configure and start of khoj server.
 Auto invokes setup steps on calling main entrypoint."
@@ -375,6 +380,7 @@ CONFIG is json obtained from Khoj config API."
          (default-chat-dir (khoj--get-directory-from-config default-config '(processor conversation conversation-logfile)))
          (chat-model (or khoj-chat-model (alist-get 'chat-model (alist-get 'openai (alist-get 'conversation (alist-get 'processor default-config))))))
          (default-model (alist-get 'model (alist-get 'conversation (alist-get 'processor default-config))))
+         (enable-offline-chat (or khoj-chat-offline (alist-get 'enable-offline-chat (alist-get 'conversation (alist-get 'processor default-config)))))
          (config (or current-config default-config)))
 
     ;; Configure content types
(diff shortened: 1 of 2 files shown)


src/interface/emacs/khoj.el
```diff
@@ -428,40 +434,44 @@ CONFIG is json obtained from Khoj config API."
              (openai (assoc 'openai conversation)))
         (when openai
           ;; Unset the `openai' field in the khoj conversation processor config
-          (message "khoj.el: disable Khoj Chat using OpenAI as your OpenAI API key got removed from config")
+          (message "khoj.el: Disable Chat using OpenAI as your OpenAI API key got removed from config")
           (setcdr conversation (delq openai (cdr conversation)))
```
sabaimran, line 438: Do you have to do anything to set the new value of `conversation` in `processor`? Or will this transfer automatically?

src/interface/emacs/khoj.el
```diff
@@ -231,6 +231,11 @@ for example), set this to the full interpreter path."
   :type 'string
   :group 'khoj)
 
+(defcustom khoj-chat-offline nil
```
sabaimran, line 234: Can you set a boolean for this field to `False` by default?

sabaimran (approved):
🏮

src/interface/emacs/khoj.el
debanjum: Yes, when `khoj-chat-offline` is `nil`, we set `enable-offline-chat` to `false` in `khoj.yml` 

**Background**:
- While elisp has a bi-state boolean (i.e nil in elisp equates to false). In the khoj config we support a 3 states even for booleans (null, true and false)
- The elisp to json convertors therefore use :false, :null to differentiate between the two possible meanings of nil in a tri-state scenario

src/interface/emacs/khoj.el
sabaimran: Nevermind, clarified offline.

debanjum: We just set `api-key` field under processor > conversation > openai section to null in this section. The next section adds this updated conversation section to processor

PR #402, opened 2023-08-02
Add support for specifying llama model files.

Ellen7ions:
# Features
In this PR, we have removed the original Enable Offline button and replaced it with an input box for the user to specify a different Llama model file.

~~You can put your models in the folder `~/user/.cache/gpt4all/` and input the model name in the box.~~
You can put your llama model anywhere and input your model path in the box.
图片替换文本
7bed42880e740bbe9ee93907bdb4ddd8

# Changes
1. Add a new API for specifying model name.
```python
@api.post("/config/data/processor/conversation/offline", status_code=200)
````
2. The original Enable Offline Chat API no longer enables, just disables offline mode.
```python
@api.post("/config/data/processor/conversation/enable_offline_chat", status_code=200)
```


sabaimran:
Hey @Ellen7ions , thanks for raising the PR! It would be great if we could discuss some of the motivations and in this discussion #408

sabaimran:
Closing this, as custom model support may require some more investigation and discussion over use cases. Discussed in #408.

Issue #379, opened 2023-07-31
Failure to download model llama-2 (Invalid model file)


JakeStevenson:
Running on macOS Ventura, M2 processor with 32G RAM.

From a fresh configuration of Khoj, I attempt to enable local chat, but after downloading the model, it errors, saying it is unable to instantiate it.

```
[07:59:56] INFO     🌘 Starting Khoj                                                                                            main.py:66
           INFO     💬 Setting up conversation processor                                                                  configure.py:286
llama-2-7b-chat.ggmlv3.q4_K_S.bin: 100%|█████████████████████████████████████████████████████████████| 3.06G/3.06G [02:15<00:00, 24.2MB/s]
Found model file at  /Users/jakes/.cache/gpt4all/llama-2-7b-chat.ggmlv3.q4_K_S.bin
Invalid model file
[08:02:12] ERROR    Failed to download model llama-2-7b-chat.ggmlv3.q4_K_S.bin from                                            utils.py:40
                    https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGML/resolve/main/llama-2-7b-chat.ggmlv3.q3_K_M.bin to
                    /Users/jakes/.cache/gpt4all/llama-2-7b-chat.ggmlv3.q4_K_S.bin. Error: Unable to instantiate model
```

I've tried a couple of times, but so far cannot get local chat to work at all.

briancunningham6:
I get the same on an M1 macbook. It does download the model file but the application fails to start stating Jakes error. I tried manually deleting the model file and kicking off Khoj by activating the local chat in the web interface but same error occurs

sabaimran:
Interesting. I wonder whether something corrupted the download in between and now there might be a persistent bad file. Could be it timed out, there was a network error, or something else.

Could one of you try the following:
1. Delete the file: `rm ~/.cache/gpt4all/llama-2-7b-chat.ggmlv3.q4_K_S.bin`
2. Download the file manually: open https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGML/resolve/main/llama-2-7b-chat.ggmlv3.q3_K_M.bin
3. Save the file to the right directory: `mv ~/Downloads/llama-2-7b-chat.ggmlv3.q3_K_M.bin ~/.cache/gpt4all/`

@briancunningham6 , does your stack trace also say that it failed to download the model?

sabaimran:
[...]

PR #405, opened 2023-08-02
Add checksums to verify the correct model is downloaded as expected

sabaimran:
- This should help debug issues related to corrupted model download
- If download fails, let the application continue

This should also help mitigate issues, like #379. Should help debug #406


debanjum (approved):
Looks good!

PR #414, opened 2023-08-04
Fixed zsh error no matches found

Muftawo:
Running pip install -e .[dev] in zsh results in following error:

```shell
pip install -e .[dev] 
zsh: no matches found: .[dev]
```

> This is because zsh uses [square brackets for globbing / pattern matching](http://zsh.sourceforge.net/Guide/zshguide05.html#l137).


sabaimran:
Closing as this was merged in #417.

PR #446, opened 2023-08-17
Sign mac app and gracefully exit GUI

### LINKED ISSUE #440: Sign DMG packages for Mac App
You can sign a DMG package using the codesign command in the Terminal. Here's an example command:
```
codesign --deep --force --verbose --sign "Developer ID Application: Your Name (ABC123DEF456)" /path/to/your/package.dmg
```

Replace "Developer ID Application: Your Name (ABC123DEF456)" with your own signing identity. You can find your signing identity in the Apple Developer portal.

Replace /path/to/your/package.dmg with the path to your DMG package.

Here's what each option in the codesign command does:

--deep: Sign all nested code within the package.
--force: Replace any existing signature.
--verbose: Print verbose output.
--sign: Specify the signing identity.
After running the codesign command, you can verify the signature using the spctl command:

This command should print accepted if the signature is valid.
```
spctl -a -vv /path/to/your/package.dmg
```

Note that you need to have a valid signing identity and a Developer ID certificate from Apple to sign your DMG package.

### LINKED ISSUE #442: Khoj GUI Crashes on Quit with QThread Killed Error
Kill signal crashes app because Qthread is not gracefully closed before app shutdown. This does not have functional consequences but is a bad user experience

### LINKED ISSUE #443: Improve Multi-threading in GUI/Desktop App
The Khoj server should be made to run multi-threaded in the background even in GUI mode. Currently the server thread is spun-off into just a single QThread to separate it's functioning from the GUI running in a foreground QThread.

This negatively impacts the performance of Khoj in GUI, Desktop mode. 

We could use a QThreadpool for the khoj server running in the background to improve the performance

sabaimran:
- Attach the parent to the server thread, allowing the kill signal to trigger a graceful exit. Closes #442
- Investigated, Close #443 as part of this effort as well. Upon further reading, it doesn't seem necessary to use `QThreadPools`. ([see docs](https://doc.qt.io/qtforpython-5/PySide2/QtCore/QThreadPool.html)). These allow you to spin up multiple threads in the GUI layer, but the server spun up on the additional server thread should be able to create as many python threads as it needs. I'll continue evaluating this.
- I also added in support for signing packages locally in my machine, as per #440. There are no code changes, required, but documenting it here for completeness. The command to sign the package is `codesign --force --deep --sign "Developer ID Application: (Name) (ID)" "khoj_<version>_arm64.dmg"`. This will have to be manually run for each release. Close #440.

### DIFF
--- a/src/khoj/interface/desktop/main_window.py
+++ b/src/khoj/interface/desktop/main_window.py
@@ -11,8 +11,8 @@ from PySide6.QtCore import QThread
 
 
 class ServerThread(QThread):
-    def __init__(self, start_server_func):
-        super(ServerThread, self).__init__()
+    def __init__(self, start_server_func, parent=None):
+        super(ServerThread, self).__init__(parent)
         self.start_server_func = start_server_func
 
     def __del__(self):
--- a/src/khoj/main.py
+++ b/src/khoj/main.py
@@ -100,7 +100,7 @@ def run():
         # Setup Server
         initialize_server(args.config, args.regenerate, required=False)
         configure_routes(app)
-        server = ServerThread(start_server_func=lambda: start_server(app, host=args.host, port=args.port))
+        server = ServerThread(start_server_func=lambda: start_server(app, host=args.host, port=args.port), parent=gui)
 
         url = f"http://{args.host}:{args.port}"
         logger.info(f"🌗 Khoj is running at {url}")


debanjum (approved):
Yay! We can finally get rid of the annoying GUI app crash warning on quitting Khoj in GUI mode (and have at least our Mac app feel more proper, with no untrusted app warnings)

PR #457, opened 2023-08-23
Move to a push-first model for retrieving embeddings from local files

sabaimran:
# Incoming
- Use a push-first model to sync data from local files for generating embeddings
- This applies only for the local, PDF and plaintext files to start with, but should also include images as well eventually
- GitHub and Notion plugins will still work the same way, but should also eventually be converted to using a push-first model by passing required configuration parameters (as opposed to raw files over the wire)
- This is part of a larger effort to make it easier to provide people with cloud-syncing options


src/khoj/configure.py
```diff
@@ -80,46 +68,56 @@ def configure_server(config: FullConfig, regenerate: bool, search_type: Optional
...
     # Initialize Content from Config
     if state.search_models:
         try:
-            state.config_lock.acquire()
-            state.content_index = configure_content(
-                state.content_index, state.config.content_type, state.search_models, regenerate, search_type
-            )
+            logger.info("📬 Updating content index Scheduler")
+            all_files = collect_files(state.config.content_type)
+            url = f"http://0.0.0.0:42110/indexer/batch?regenerate={regenerate}"
```
debanjum, line 77: The `host` and `port` shouldn't be statically set. It is user configurable during app start `khoj --host <host-ip> --port <port_number>`

tests/test_client.py
```diff
@@ -52,41 +51,48 @@ def test_update_with_invalid_content_type(client):
...
 
 # ----------------------------------------------------------------------------------------------------
-def test_regenerate_with_invalid_content_type(client):
+def test_index_batch(client):
+    # Arrange
+    request_body = {
+        "org": {"additionalProp1": "string", "additionalProp2": "string", "additionalProp3": "string"},
+        "pdf": {"additionalProp1": "string", "additionalProp2": "string", "additionalProp3": "string"},
+        "plaintext": {"additionalProp1": "string", "additionalProp2": "string", "additionalProp3": "string"},
+        "markdown": {"additionalProp1": "string", "additionalProp2": "string", "additionalProp3": "string"},
```
debanjum, line 69: Should we populate the test data with something closer to what we expect in the request body? Such tests help understand the code better IMO.

See e.g below
```suggestion
        "org": {"/path/to/file1.org": "content of file 1 ", "/path/to/file2.org": "content of file 2", "/path/to/file3.org": "content of file 3"},
        "pdf": {"additionalProp1": "string", "additionalProp2": "string", "additionalProp3": "string"},
        "plaintext": {"additionalProp1": "string", "additionalProp2": "string", "additionalProp3": "string"},
        "markdown": {"additionalProp1": "string", "additionalProp2": "string", "additionalProp3": "string"},
```

src/khoj/routers/indexer.py
```diff
@@ -0,0 +1,237 @@
...
+        logger.warning(f"🚨 No files to process for {t.value} search.")
+        return None
+
+    try:
+        # Initialize Org Notes Search
+        if (
+            (t == None or t.value == state.SearchType.Org.value)
+            and content_config.org
+            and search_models.text_search
+            and files["org"]
```
debanjum, line 105: Wouldn't this fail in the scenario where no org-mode files exist (anymore) in the `files` dictionary? If so, this would affect the configuration of other content types below as well

src/khoj/search_type/text_search.py
```diff
@@ -174,6 +174,7 @@ def collate_results(hits, entries: List[Entry], count=5) -> List[SearchResponse]
 
 def setup(
     text_to_jsonl: Type[TextToJsonl],
+    files: dict[str, str],
```
debanjum, line 177: Note: Should we type the `files` argument into a class, at least once these changes have stabilized?

sabaimran: Certainly, once the types have been stabilized, it makes sense to move them into a class with `pydantic`.

src/khoj/routers/indexer.py
sabaimran: Sure, let me use `.get` instead.

PR #463, opened 2023-08-26
Use Slash Commands and Add Notes Slash Command

debanjum:
- c6b90f7 Move to slash commands instead of using @ to specify general commands
- 01f60ed Add /notes command to respond using only knowledge base as context


sabaimran (approved):
Pushed some changes, including:
- Fixed `conftest` for offline chat
- Added command hints in the chat.html
- Removed `default` conversation command

looks good!

PR #464, opened 2023-08-26
Fix plaintext HTML parsing and rendering

sabaimran:
# Incoming
- Use beautifulsoup to pull out just the relevant text from an HTMl page
- When rendering, strip any tags that might be present
- Save the conversations in the conversation history even if no relevant notes were found


debanjum (approved):
Looks good

PR #468, opened 2023-08-28
Add Default Command back to Chat Commands

debanjum:
### Changes
- 613c64e Test general and notes slash commands in offline chat director tests
- 45f6959 Use general prompt when no references found and not in Notes mode
- 28d9abf Make Khoj ask clarifying questions when answer not in provided context
- 6b3e5f9 Show filtered list of commands available with the currently input text
- 204c13e Add default conversation command to auto switch b/w general, notes modes

### DIFF
--- a/src/khoj/processor/conversation/openai/gpt.py
+++ b/src/khoj/processor/conversation/openai/gpt.py
@@ -119,11 +119,11 @@ def converse(
     compiled_references = "\n\n".join({f"# {item}" for item in references})
 
     # Get Conversation Primer appropriate to Conversation Type
-    if conversation_command == ConversationCommand.General:
-        conversation_primer = prompts.general_conversation.format(current_date=current_date, query=user_query)
-    elif conversation_command == ConversationCommand.Notes and is_none_or_empty(compiled_references):
+    if conversation_command == ConversationCommand.Notes and is_none_or_empty(compiled_references):
         completion_func(chat_response=prompts.no_notes_found.format())
         return iter([prompts.no_notes_found.format()])
+    elif conversation_command == ConversationCommand.General or is_none_or_empty(compiled_references):
+        conversation_primer = prompts.general_conversation.format(current_date=current_date, query=user_query)
     else:
         conversation_primer = prompts.notes_conversation.format(
             current_date=current_date, query=user_query, references=compiled_references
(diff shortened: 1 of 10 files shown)


src/khoj/processor/conversation/openai/gpt.py
```diff
@@ -119,11 +119,11 @@ def converse(
...
     # Get Conversation Primer appropriate to Conversation Type
-    if conversation_command == ConversationCommand.General:
-        conversation_primer = prompts.general_conversation.format(current_date=current_date, query=user_query)
-    elif conversation_command == ConversationCommand.Notes and is_none_or_empty(compiled_references):
+    if conversation_command == ConversationCommand.Notes and is_none_or_empty(compiled_references):
         completion_func(chat_response=prompts.no_notes_found.format())
         return iter([prompts.no_notes_found.format()])
+    elif conversation_command == ConversationCommand.General or is_none_or_empty(compiled_references):
+        conversation_primer = prompts.general_conversation.format(current_date=current_date, query=user_query)
     else:
```
sabaimran, line 127: This should explicitly use `ConversationCommand.Default`. If we add a new `ConversationCommand`, and forget to update the behavior here, it should fail.

src/khoj/utils/helpers.py
```diff
@@ -214,13 +214,15 @@ def log_telemetry(
...
     General = "general"
     Notes = "notes"
     Help = "help"
 
 
 command_descriptions = {
-    ConversationCommand.General: "This command allows you to search talk with the LLM without including context from your knowledge base.",
-    ConversationCommand.Notes: "This command allows you to search talk with the LLM while including context from your knowledge base.",
-    ConversationCommand.Help: "This command displays a help message with all available commands and other metadata.",
+    ConversationCommand.General: "Only talk about information that relies on Khoj's general knowledge, not your personal knowledge base.",
```
nickanderson, line 224: In edffbad you refer to "my" notes, here you refer to khoj's knowledge base. If this misalignment in ownership context significant? Or should it be Claridge that my notes are part of your knowledge badge?

sabaimran: @nickanderson , thanks for pointing this out -- I think this might be way Llama sort of loses the thread sometimes and messed up its person. I do think this is significant; good catch!

PR #496, opened 2023-10-10
Changed globbing.  Now doesn't clobber a users glob if they want to a…

spott:
…dd it, but will (if just given a directory), add a recursive glob.

Note: python's glob engine doesn't support `{}` globing, a future option is to warn if that is included.

Closes https://github.com/khoj-ai/khoj/issues/493

### DIFF
--- a/src/khoj/interface/web/content_type_input.html
+++ b/src/khoj/interface/web/content_type_input.html
@@ -116,8 +116,8 @@
             suffixes = [".pdf"]
         else if ('{{content_type}}' === "plaintext")
             suffixes = ['*', '*.*']
-        //
-        let globs = suffixes.map(x => `${globformat}${x}`)
+
+        let globs = suffixes.map(x => `${globFormat}${x}`)
         var inputFileNodes = document.getElementsByName("input-files");
         var inputFiles = getValidInputNodes(inputFileNodes).map(node => node.value);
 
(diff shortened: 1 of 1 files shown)


src/khoj/interface/web/content_type_input.html
```diff
@@ -106,28 +106,38 @@ <h2 class="section-title">
...
+            suffixes = [".md", ".markdown"]
         else if ('{{content_type}}' == "org")
-            suffixes = ["org"]
+            suffixes = [".org"]
         else if ('{{content_type}}' === "pdf")
-            suffixes = ["pdf"]
+            suffixes = [".pdf"]
         else if ('{{content_type}}' === "plaintext")
-            suffixes = ['*']
+            suffixes = ['*', '*.*']
```
debanjum, line 118: Shouldn't the plaintext suffix just be `.*`, so that the plaintext glob will look like `**/*.*` (like it does currently)?
 
```suggestion
            suffixes = ['.*']
```

debanjum (commented):
PR looks good (apart from the one comment)!

src/khoj/interface/web/content_type_input.html
sabaimran: Hey @spott! I'm going to commit @debanjum 's suggestion because I want this to make it for our next release. If you do have time to look at it again and remember why you configured the plaintext pattern in this way, don't hesitate to leave a comment.

sabaimran (approved):

Issue #243, opened 2023-06-25
Khoj.el calls `khoj-server-url` on package load during Emacs init


debanjum:
You'll see a call to http://localhost:8000 (or whatever `khoj-server-url` is set to) when you start Emacs. This is not intentional/required and should be removed.

PR #499, opened 2023-10-13
Push Files to Index from Obsidian, Emacs & Desktop Clients using Multi-Part Forms Method

### LINKED ISSUE #471: Update Emacs, Obsidian clients to push files to the Khoj backend
With changes to support the Khoj backend directly reading streamed data from respective syncing clients (https://github.com/khoj-ai/khoj/pull/457), Emacs and Obsidian plugins should also be modified to push the raw data. This should include scheduled jobs that occasionally sync between the client and server.

With that, the Khoj backend should gradually lose its dependency on reading from the file system for any data.

debanjum:
### Overview
- Add ability to push data to index from the Emacs, Obsidian clients
- Switch to standard mechanism of syncing files via HTTP multi-part/form. Previously we were streaming the data as JSON
  - Benefits of new mechanism
    - No manual parsing of files to send or receive on clients or server is required as most have in-built mechanisms to send multi-part/form requests
    - The whole response is not required to be kept in memory to parse content as JSON. As individual files arrive they're automatically pushed to disk to conserve memory if required
    - Binary files don't need to be encoded on client and decoded on server

### Major
- 60e9a61 Use multi-part form to receive files to index on server
- 68018ef Use multi-part form to send files to index on desktop client
- fc99431 Send files to index on server from the khoj.el emacs client
  - 292f042 Send content for indexing on server at a regular interval from khoj.el
- f2e293a: Send files to index on server from the khoj obsidian client
- bed3aff 6baaaaf Update tests to test multi-part/form method of pushing files to index

#### Minor
- 6aa69da Put indexer API endpoint under /api path segment
- bea196a Explicitly make GET request to /config/data from khoj.el:khoj-server-configure method
- 9ba173b Improve emoji, message on content index updated via logger
- 7b9161e Don't call khoj server on khoj.el load, only once khoj invoked explicitly by user
- Improve indexing of binary files
  - 541cd59 Let fs_syncer pass PDF files directly as binary before indexing
  - d27dc71 Use encoding of each file set in indexer request to read file 
- 99a2c93 Add CORS policy to khoj server. Allow requests from khoj apps, obsidian & localhost
- 84654ff Update indexer API endpoint URL to` index/update` from `indexer/batch`

Resolves #471 #243

### DIFF
--- a/src/interface/desktop/main.js
+++ b/src/interface/desktop/main.js
@@ -163,7 +163,7 @@ function pushDataToKhoj (regenerate = false) {
         const headers = {
             'x-api-key': 'secret'
         };
-        axios.post(`${hostURL}/api/v1/index/update?regenerate=${regenerate}`, formData, { headers })
+        axios.post(`${hostURL}/api/v1/index/update?force=${regenerate}&client=desktop`, formData, { headers })
             .then(response => {
                 console.log(response.data);
                 const win = BrowserWindow.getAllWindows()[0];
--- a/src/interface/desktop/main.js
+++ b/src/interface/desktop/main.js
@@ -93,9 +93,9 @@ function filenameToMimeType (filename) {
         case 'png':
             return 'image/png';
         case 'jpg':
-            return 'image/jpeg';
         case 'jpeg':
             return 'image/jpeg';
+        case 'md':
         case 'markdown':
             return 'text/markdown';
         case 'org':
--- a/src/interface/desktop/main.js
+++ b/src/interface/desktop/main.js
@@ -169,7 +169,7 @@ function pushDataToKhoj (regenerate = false) {
 
     const hostURL = store.get('hostURL') || KHOJ_URL;
 
-    axios.post(`${hostURL}/v1/indexer/batch?regenerate=${regenerate}`, stream, { headers })
+    axios.post(`${hostURL}/api/v1/indexer/batch?regenerate=${regenerate}`, stream, { headers })
         .then(response => {
             console.log(response.data);
             const win = BrowserWindow.getAllWindows()[0];
--- a/src/interface/desktop/main.js
+++ b/src/interface/desktop/main.js
@@ -8,7 +8,6 @@ const {dialog} = require('electron');
 
 const cron = require('cron').CronJob;
 const axios = require('axios');
-const { Readable } = require('stream');
 
 const KHOJ_URL = 'http://127.0.0.1:42110'
 
@@ -65,7 +64,7 @@ const schema = {
 
 var state = {}
 
-const store = new Store({schema});
+const store = new Store({ schema });
 
 console.log(store);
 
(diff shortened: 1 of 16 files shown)


src/khoj/routers/indexer.py
```diff
@@ -103,9 +85,9 @@ async def index_batch(
                 dict_to_update = plaintext_files
 
             if dict_to_update is not None:
-                dict_to_update[file.path] = file.content
+                dict_to_update[file.filename] = file.file.read().decode("utf-8")
```
sabaimran, line 88: Does this work even if the file is binary encoded? Does it work for PDFs, and would it work for images as well?

src/interface/desktop/main.js
```diff
@@ -151,44 +153,37 @@ function pushDataToKhoj (regenerate = false) {
 
     for (const syncedFile of lastSync) {
         if (!filesToPush.includes(syncedFile.path)) {
-            data.files.push({
-                path: syncedFile.path,
-                content: ""
-            });
+            fileObj = new Blob([""], { type: filenameToMimeType(syncedFile.path) });
```
sabaimran, line 156: Files are currently being sent as `UploadFile` objects. Let's say we wanted to update this spec and allow the clients to send operation "types". For example, the operation actually being communicated here is a `delete`. It would be much cleaner, for intent, to relay that information by specifying it to be a `delete` op. Similarly, indexing a brand new file would be an `add`, an existing file an `update`. Would that be a heavy lift to modify the API specification here, or is that doable? Asking as the request format is already being updated.

src/khoj/routers/indexer.py
```diff
@@ -58,7 +57,7 @@ class IndexerInput(BaseModel):
 
 @indexer.post("/batch")
 async def index_batch(
-    request: Request,
```
sabaimran, line 61: Could you add telemetry here as well? We'll need to retain the `request` object in that case.

sabaimran (commented):
nice, cleaner standardization of the API spec!

src/interface/desktop/main.js
debanjum: You can pass other fields in the request apart from the files to update. So while you can't pass the data in the UploadFile object directly you can pass in a dictionary that maps `filepath` to `operation-type`. Is that sufficient for these API spec updates?

src/khoj/routers/indexer.py
debanjum: Umm, not sure but valid concern. Let me test that

src/interface/desktop/main.js
sabaimran: Yeah, that makes sense. We can add that as a follow-on change later.

src/khoj/routers/indexer.py
debanjum: Added telemetry in e347823

src/interface/desktop/main.js
debanjum: Sounds good

src/khoj/routers/indexer.py
debanjum: Fixed up and validated indexing of PDF files
  - 541cd59 Let fs_syncer pass PDF files directly as binary before indexing
  - d27dc71 Use encoding of each file set in indexer request to read file

PR #503, opened 2023-10-17
[Multi-User Part 2]: Add login pages and gate access to application behind login wall

sabaimran:
# Incoming
- Make most routes conditional on authentication *if anonymous mode is not enabled*. If anonymous mode is enabled, it scaffolds a default user and uses that for all application interactions.
- Add a basic login page and add routes for redirecting the user if logged in


src/khoj/configure.py
```diff
@@ -67,7 +67,7 @@ async def authenticate(self, request):
             user = await self.khojuser_manager.filter(email=current_user.get("email")).afirst()
             if user:
```
sabaimran, line 68: In the `authenticate` method of the middleware, we should also check whether the `current_user`'s session object has an `exp` timestamp that exceeds the current time.

src/khoj/routers/api.py
```diff
@@ -127,13 +127,15 @@ def _initialize_config():
             state.processor_config = configure_processor(state.config.processor)
 
     @api.get("/config/data", response_model=FullConfig)
+    @requires(["authenticated"], redirect="login_page")
```
debanjum, line 130: These API endpoints (apart from Web client endpoints) have been updated in next PRs, to not redirect to login page but return unauthorized http status code in response. So no change required here

src/khoj/routers/indexer.py
```diff
@@ -156,7 +157,6 @@ async def update(
         host=host,
     )
 
-    logger.info(f"📪 Content index updated via API call by {client} client")
```
debanjum, line 159: Shouldn't this stay here?

src/khoj/main.py
```diff
@@ -14,13 +14,16 @@
...
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
+from rich.logging import RichHandler
 import schedule
-import django
 
 from fastapi.staticfiles import StaticFiles
 from rich.logging import RichHandler
+import schedule
+
```
debanjum, line 26: Why are these showing up as new imports? None of the new code in this file seems to need these packages

src/khoj/interface/web/login.html
```diff
@@ -0,0 +1,197 @@
+<html>
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0 maximum-scale=1.0">
+        <title>Khoj - Search</title>
```
debanjum, line 5: Should the title say "Khoj - Config"?

```suggestion
        <title>Khoj - Config</title>
```

src/khoj/interface/web/login.html
```diff
@@ -0,0 +1,197 @@
...
+                    }),
+                    headers: {
+                        "Content-Type": "application/json"
+                    }
+                }).then(function(response) {
+                    return response.json();
+                }).then(function(data) {
+                    console.log(data);
+                    if (data.user != null) {
+                        document.getElementById("khoj-banner").innerHTML = "Thanks for signing up. We'll be in touch soon! 🚀";
```
debanjum, line 185: Is this the text to be shown post signup?

debanjum (approved):
Looks good! Some of the changes suggested for this PR may already be addressed in next PRs for the rearchitecture work. Feel free to ignore them in this PR

src/khoj/interface/web/login.html
sabaimran: Oh, that's only if it's a demo instance. We actually don't need this here -- Probably won't have a demo instance + sign-in.

PR #511, opened 2023-10-19
[Multi-User Part 3]: Separate chat sesssions based on authenticated users

sabaimran:
# Incoming
- Add a data model which allows us to store Conversations with users. This does a minimal lift over the current setup, where the underlying data is stored in a JSON file. This maintains parity with that configuration.
- There does _seem_ to be some regression in chat quality, which is most likely attributable to search results.

This will help us with #275. It should become much easier to maintain multiple Conversations in a given table in the backend now. We will have to do some thinking on the UI.


src/khoj/routers/api.py
```diff
@@ -341,24 +320,26 @@ async def set_processor_enable_offline_chat_config_data(
...
-            state.config.processor.conversation.offline_chat.chat_model = offline_chat_model
-        state.processor_config = configure_processor(state.config.processor, state.processor_config)
+            offline_chat = await ConversationAdapters.get_offline_chat(user)
+            chat_model = offline_chat.chat_model
+            if state.gpt4all_processor_config is None:
+                state.gpt4all_processor_config = GPT4AllProcessorModel(chat_model=chat_model)
+
+        else:
+            await sync_to_async(ConversationAdapters.clear_offline_chat_conversation_config)(user)
+            state.gpt4all_processor_config = None
```
debanjum, line 342: This should be a user specific state variable now?

debanjum (approved):
Did a quick review, but looks good apart from some of the comments I've left. They can be addressed in followup PRs if required

src/khoj/routers/api.py
sabaimran: In our self-hosted model, we can't really support multiple offline models on the server. May need to think more about how to make this apparent in the UX, but it makes sense for it to be shared across the application state when self-hosted.

PR #524, opened 2023-10-29
Autofocus to chat input when openning chat

tuan3w:
This merge enable autofocus when chat model is open. This allow the user to chat with Khoj without extra action.


sabaimran (approved):

sabaimran:
Tested locally! Looks good 👍🏾

PR #528, opened 2023-11-01
[Multi-User Part 7]: Improve Sign-In UX & Rename DB Models for Readability

debanjum:
### ✨ New
- f585a71 Create profile pic drop-down menu in navigation pane
  Put settings page, logout action under drop-down menu

### ⚙️ Fix
- f77336ba Add Key icon for API keys table on Web Client's settings page

### 🧪 Improve
- 87e6b1ea Rename `TextEmbeddings` to `TextEntries` for improved readability
- bcbee05a Rename `Db.Models` `Embeddings`, `EmbeddingsAdapter` to `Entry`, `EntryAdapter`
- 58a71719 Show truncated API key for identification & restrict table width for config page responsiveness

### Screenshots

| Before | After
| --- | --- |
|   |  |


src/khoj/routers/web_client.py
```diff
@@ -34,19 +34,52 @@
...
+        },
+    )
 
 
 @web_client.get("/chat", response_class=FileResponse)
 @requires(["authenticated"], redirect="login_page")
 def chat_page(request: Request):
-    return templates.TemplateResponse("chat.html", context={"request": request, "demo": state.demo})
+    user = request.user.object
+    user_picture = request.session.get("user", {}).get("picture")
```
sabaimran, line 72: Is the `picture` field always populated? What happens if the user hasn't set a profile picture? Can we supply a generic/default avatar?

src/khoj/interface/web/chat.html
```diff
@@ -281,10 +282,17 @@
                 </a>
             {% endif %}
             <nav class="khoj-nav">
-                <a class="khoj-nav khoj-nav-selected" href="/chat">Chat</a>
-                <a class="khoj-nav" href="/">Search</a>
+                <a class="khoj-nav khoj-nav-selected" href="/chat">💬 Chat</a>
+                <a class="khoj-nav" href="/">🔎 Search</a>
                 {% if not demo %}
-                    <a class="khoj-nav" href="/config">Settings</a>
+                    <!-- Dropdown Menu -->
```
sabaimran, line 288: Could it be worthwhile to use Jinja Macros to define this account component? That way we won't have to copy/paste as much and we'll get better re-use. https://jinja.palletsprojects.com/en/3.0.x/templates/#macros

src/khoj/processor/text_to_jsonl.py
```diff
@@ -12,14 +12,14 @@
 from khoj.utils.rawconfig import Entry
```
sabaimran, line 12: All these `x_to_jsonl` files also need to be renamed to something like `x_to_entries` or `x_to_embeddings`. Do you want to take care of that as part of this PR as well?

src/khoj/routers/web_client.py
```diff
@@ -34,19 +34,52 @@
...
-    return templates.TemplateResponse("index.html", context={"request": request, "demo": state.demo})
+    user = request.user.object
+    user_picture = request.session.get("user", {}).get("picture")
+
+    return templates.TemplateResponse(
+        "index.html",
+        context={
+            "request": request,
+            "demo": state.demo,
+            "username": user.username if user else None,
```
sabaimran, line 45: `user` should never be `None` here. Same with the `index_post` method below.
```suggestion
            "username": user.username,
```

sabaimran:
Can you also include a screenshot? 😁

debanjum:
> Can you also include a screenshot? 😁

Added before, after screenshot

src/khoj/routers/web_client.py
debanjum: Updated code to use username initial as default profile picture. This matches behavior in other applications for default profile picture

src/khoj/processor/text_to_jsonl.py
debanjum: Good idea. Updated `s/x_to_jsonl/x_to_entries` for files and `s/XToJsonl/XToEntries/` for class names

src/khoj/routers/web_client.py
debanjum: Valid point. Fixed

src/khoj/interface/web/chat.html
debanjum: Great idea, didn't know that existed. Reduced duplication of HTML by creating a `utils.html` with the macro to generate the navigation pane

PR #537, opened 2023-11-07
Create Billing integration. Improve Settings pages on Desktop, Web apps

debanjum:
### Major
- Expose Billing via Stripe on Khoj Web app for Khoj Cloud subscription
  - f466751f Expose card on web app config page to manage subscription to Khoj cloud
  - 9aaf475c Create API webhook, endpoints for subscription payments using Stripe
- Put Computer files to index into Card under Content section
  - 156421d3 Show file type icons for each indexed file in config card of web app
  - 7c424e0d Enable deleting all indexed desktop files from Khoj via Desktop app
  - 6e957584 Create config page on web app to manage computer files indexed by Khoj
- Track data source (computer, github, notion) of each entry
  - d527b644 Update content by source via API. Make web client use this API for config
  - 9ab327a2 Store the data source of each entry in database

### Cleanup
- 045c2252 Set content enabled status on update via config buttons on web app
- c82cd086 Delete deprecated content config pages for local files from web client
- 97cf8339 Rename Sync button, Force Sync toggle to Save, Save All buttons

### Fixes
- 779fa531 Prevent Desktop app triggering multiple simultaneous syncs to server
- 9f47fc8e Upgrade langchain version since adding support for OCR-ing PDFs
- 404d47f1 Bubble up content indexing errors to notify user on client apps


src/khoj/routers/api.py
```diff
@@ -740,3 +724,61 @@ async def extract_references_and_questions(
         compiled_references = [item.additional["compiled"] for item in result_list]
 
     return compiled_references, inferred_queries, defiltered_query
+
+
+# Stripe integration for Khoj Cloud Subscription
+stripe.api_key = os.getenv("STRIPE_API_KEY")
+endpoint_secret = os.getenv("STRIPE_SIGINING_SECRET")
```
sabaimran, line 731: Similar to `auth.py`, we can move this into a separate route file, and add a warning for users if they don't have these environment variables set.

And if they are not set, we probably want to return 404s for any of the below endpoints?

src/khoj/routers/api.py
```diff
@@ -740,3 +724,61 @@ async def extract_references_and_questions(
...
+    ):
+        # Retrieve the customer's details
+        customer_id = event["data"]["object"]["customer"]
+        customer = stripe.Customer.retrieve(customer_id)
+        customer_email = customer["email"]
+        # Mark the customer as subscribed
+        user = await adapters.set_user_subscribed(customer_email)
+        if not user:
+            success = False
+    elif event["type"] == "customer.subscription.updated" or event["type"] == "customer.subscription.deleted":
```
sabaimran, line 764: What kind of `updates` might there be here? Would we need to change something, like the `subscription_renewal_date`, for certain changes?

src/database/models/__init__.py
```diff
@@ -14,7 +14,15 @@ class Meta:
 
 
 class KhojUser(AbstractUser):
+    class SubscriptionType(models.TextChoices):
+        TRIAL = "trial"
+        STANDARD = "standard"
+
     uuid = models.UUIDField(models.UUIDField(default=uuid.uuid4, editable=False))
+    subscription_type = models.CharField(
```
sabaimran, line 22: Do you think `Subscription` would better serve as a separate, isolated table? I imagine it would be safer to separate that from the core `User` object in order to keep it less cluttered over time. We can also use the `BaseModel` in that case. Open to hear your thoughts.

```python
class Subscription(BaseModel):
   class SubscriptionType(models.TextChoices):
      TRIAL = "trial"
      STANDARD = "standard"
   user = models.OneToOneField(KhojUser, on_delete=models.CASCADE)
   subscription_type = models.CharField(max_length=20, choices=SubscriptionType.choices, default=SubscriptionType.TRIAL)
   subscription_renewal_date = models.DateTimeField(null=True, default=None)
```

sabaimran (commented):
I have a couple of questions/suggestions, but otherwise looks good! Thanks for the major clean-up 🚀 .

src/khoj/interface/web/config.html
```diff
@@ -122,13 +158,39 @@ <h3 class="card-title">API Keys</h3>
...
+                            src="/static/assets/icons/confirm-icon.svg"
+                            alt="Configured">
+                    </h3>
+                </div>
+                <div class="card-description-row">
+                    <p class="card-description">Manage your subscription to Khoj Cloud</p>
+                </div>
+                {% if not is_subscribed %}
+                <div class="card-action-row">
+                    <a class="card-button happy" href="https://buy.stripe.com/28ocQb7kb4iD0cEdQQ?email={{ username }}" target="_blank">
```
sabaimran, line 181: nit: maybe use `happy` for after subscription is active?

Also, when I click on the link, it doesn't autopopulate my email address in the checkout page.

sabaimran: And when the `stripe` settings are not configured, this card shouldn't appear.

sabaimran: Should we drive the link here through an environment variable and pass it to the template so we can update it dynamically when required?

debanjum: - Fixed auto-population for my email address
- Hid the billing card if strip environment not setup
- Now using environment variable `KHOJ_CLOUD_SUBSCRIPTION_URL` to set the Stripe payment link
- Letting the subscribe card be happy colored for now :)

src/khoj/routers/api.py
debanjum: Moved the subscription routes to separate route file and disabled the API endpoints when stripe environment not setup. 

Showing when billing is enabled in logs instead of warning as default self-hosted users can't do anything about it (unlike the Google OAuth warning where `--anonymous-mode` should be turned on so warning is a good idea)

PR #541, opened 2023-11-12
Add test separators

asim-shrestha:


tests/test_word_filter.py
```diff
@@ -2,7 +2,8 @@
 from khoj.search_filter.word_filter import WordFilter
 from khoj.utils.rawconfig import Entry
 
-
+# Test
```
debanjum, line 5: Black formatting check failed. I think it's because of the deleted newline here? 

```suggestion

# Test
```

PR #548, opened 2023-11-16
Index Parent Headings of Org-Mode Entries to Improve Search Context

### LINKED ISSUE #85: Consider outline hierarchy when searching?
I am playing around with the search and comparing the results with my https://github.com/yantar92/org-ql/ setup.

One obvious point that is not handled by khoj is matching the outline path for nodes:
Consider the following outline structure:

* Topics
** Emacs
*** debanjum [Github] debanjum/khoj: Natural Language Search Engine for your Org-Mode and Markdown notes, Beancount transactions and Photos

A natural way to search for the deepest node is "emacs org search engine", but khoj does not yield the match here.

debanjum:
### Overview
The parent hierarchy of org-mode entries can store important context. 

This change updates OrgNode to track parent headings for each org entry and adds the parent outline for each entry to the index

### Details
- ddb07def Test search uses ancestor headings as context for improved results
- 74403e35 Add ancestor headings of each org-mode entry to their compiled form
- 305c25ae Track ancestor headings for each org-mode entry in org-node parser

Resolves #85

### DIFF
--- a/src/khoj/processor/org_mode/org_to_entries.py
+++ b/src/khoj/processor/org_mode/org_to_entries.py
@@ -110,7 +110,7 @@ class OrgToEntries(TextToEntries):
 
             compiled = heading
             if state.verbose > 2:
-                logger.debug(f"Title: {parsed_entry.heading}")
+                logger.debug(f"Title: {heading}")
 
             if parsed_entry.tags:
                 tags_str = " ".join(parsed_entry.tags)
--- a/src/khoj/processor/org_mode/org_to_entries.py
+++ b/src/khoj/processor/org_mode/org_to_entries.py
@@ -100,12 +100,13 @@ class OrgToEntries(TextToEntries):
                 continue
 
             todo_str = f"{parsed_entry.todo} " if parsed_entry.todo else ""
-            # Prepend filename as top heading to entry
-            filename = Path(entry_to_file_map[parsed_entry]).stem
+
+            # Prepend ancestor headings, filename as top heading to entry for context
+            ancestors_trail = " / ".join(parsed_entry.ancestors) or Path(entry_to_file_map[parsed_entry])
             if parsed_entry.heading:
-                heading = f"* {filename}\n** {todo_str}{parsed_entry.heading}."
+                heading = f"* Path: {ancestors_trail}\n** {todo_str}{parsed_entry.heading}."
             else:
-                heading = f"* {filename}."
+                heading = f"* Path: {ancestors_trail}."
 
             compiled = heading
             if state.verbose > 2:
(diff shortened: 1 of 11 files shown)


src/khoj/processor/org_mode/org_to_entries.py
```diff
@@ -100,12 +100,13 @@ def convert_org_nodes_to_entries(
                 continue
 
             todo_str = f"{parsed_entry.todo} " if parsed_entry.todo else ""
-            # Prepend filename as top heading to entry
-            filename = Path(entry_to_file_map[parsed_entry]).stem
+
+            # Prepend ancestor headings, filename as top heading to entry for context
```
sabaimran, line 104: It would be either ancestor headings OR filename, right? It looks like `filename` won't be included if it has ancestors, but correct me if I'm wrong. We should make sure to include them.

src/khoj/processor/org_mode/orgnode.py
```diff
@@ -126,6 +129,17 @@ def makelist(file, filename):
                     for parsedtag in parsedtags.split(":"):
                         if parsedtag != "":
                             tags.append(parsedtag)
+
+            # Add previous heading to ancestors if current heading is deeper than previous level
```
sabaimran, line 133: Neat! This will make it a lot more useful, and hopefully will make it easier to retrieve sibling data in a given query.

sabaimran (approved):

yantar92:
sabaimran ***@***.***> writes:

> +            # Prepend ancestor headings, filename as top heading to entry for context
>
> It would be either ancestor headings OR filename, right? It looks like `filename` won't be included if it has ancestors, but correct me if I'm wrong. We should make sure to include them.

Another candidate for indexing that is equivalent to the filename is the file #+TITLE.

-- 
Ihor Radchenko // yantar92,
Org mode contributor,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

src/khoj/processor/org_mode/org_to_entries.py
debanjum: Yeah. filename is included in the ancestor headings

src/khoj/processor/org_mode/orgnode.py
debanjum: Yeah, this should improve search quality for nested org-mode entry trees by providing relevant, additional context

debanjum:
>> Another candidate for indexing that is equivalent to the filename is the file #+TITLE.

Hey @yantar92, yeah good idea. I've updated the code to use #+TITLE, when present, instead of filename as the root ancestor of all entries in an org-mode file

PR #551, opened 2023-11-17
Delete Conversation History from Web, Desktop, Obsidian Clients

### LINKED ISSUE #523: Feature request: Ability to clear chat
I use khoj plugin in Obsidian. And many time, I want to clear the current chat session. I know I can delete `conversation.json` to clear chat history. However, I think it would be helpful to have a button to clear chat history explicitly. Thanks.

debanjum:
Add delete button to clear conversation history from Web, Desktop and Obsidian Khoj clients

Resolves #523


src/khoj/routers/api.py
```diff
@@ -603,7 +603,13 @@ async def chat(
         request, meta_log, q, (n or 5), (d or math.inf), conversation_command
     )
 
-    if conversation_command == ConversationCommand.Default and is_none_or_empty(compiled_references):
+    if conversation_command == ConversationCommand.Clear:
+        await ConversationAdapters.adelete_conversation_by_user(user)
+        return StreamingResponse(
+            iter(["Cleared conversation history"]), media_type="text/event-stream", status_code=200
+        )
```
sabaimran, line 610: Shouldn't the UI update when this occurs? We probably want to handle it in a special way on the client side.

src/khoj/processor/conversation/prompts.py
```diff
@@ -253,6 +253,7 @@
 **/notes**: Chat using the information in your knowledge base.
 **/general**: Chat using just Khoj's general knowledge. This will not search against your notes.
 **/default**: Chat using your knowledge base and Khoj's general knowledge for context.
+**/clear**: Clear the chat session.
```
sabaimran, line 256: Hmm, I'm not really a fan of having this in the `/` commands. I think it should be a UI button that allows the user to clear the chat. That will be more extensible as we go into a multi-session design and allow users to clear/manage multiple chats simultaneously. We'll want it to do be a specific API call that deletes that specific conversation by ID. For now, deleting all by the user is ok, but it should be eventually be granular on the specific chat.

`/` commands IMO are more about modality, less about management.

src/khoj/routers/api.py
sabaimran: This will be easier if we create a separate UI button for this action; it would clear the specific chat in the UI and send a delete request on the server side.

src/khoj/processor/conversation/prompts.py
debanjum: Makes sense. Moved to using a delete button, instead of using a slash command to clear conversation

src/khoj/routers/api.py
debanjum: Updated code to clear the conversation history displayed on the clients after the clear history action has successfully executed

PR #556, opened 2023-11-20
Add support for suggesting base questions to users

sabaimran:
- Add support for reflective questions to prompt the user
- The admin can set a couple of good enough default questions, but we can also extend this into providing more context-aware suggestions depending on the user's conversations and content


src/database/adapters/__init__.py
```diff
@@ -341,6 +342,25 @@ async def get_openai_chat():
...
+        all_questions = await sync_to_async(ReflectiveQuestion.objects.filter(user=None).values_list)(
+            "question", flat=True
+        )
+
+        max_results = 3
+        all_questions = await sync_to_async(list)(all_questions)
+        if len(all_questions) < max_results:
+            return all_questions
+
+        return random.sample(all_questions, max_results)
```
debanjum, line 362: I think you need to import `random` package here

debanjum (approved):
Super useful start to provide guidance to users, reduce cognitive load to try Khoj out

PR #559, opened 2023-11-21
fix: Fix local server usage

### LINKED ISSUE #558: [Regression] Obsidian Plugin connectivity check no longer works with local server
https://github.com/khoj-ai/khoj/commit/0a950d9382ef0794cbf85463ab48943879434284
https://github.com/khoj-ai/khoj/blob/master/src/interface/obsidian/src/main.ts#L69-L91

I think this PR made the connectivity check only work against https://app.khoj.dev, which means that indexing continues to work, but none of the commands are registered onLoad.

dtkav:
fixes #558

### DIFF
--- a/src/interface/obsidian/src/main.ts
+++ b/src/interface/obsidian/src/main.ts
@@ -73,21 +73,19 @@ export default class Khoj extends Plugin {
         // Check if khoj backend is configured, note if cannot connect to backend
         let headers = { "Authorization": `Bearer ${this.settings.khojApiKey}` };
 
-        if (this.settings.khojUrl === "https://app.khoj.dev") {
-            if (this.settings.khojApiKey === "") {
-                new Notice(`❗️Khoj API key is not configured. Please visit https://app.khoj.dev/config#clients to get an API key.`);
-                return;
-            }
-
-            await request({ url: this.settings.khojUrl ,method: "GET", headers: headers })
-                .then(response => {
-                    this.settings.connectedToBackend = true;
-                })
-                .catch(error => {
-                    this.settings.connectedToBackend = false;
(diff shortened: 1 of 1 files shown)


debanjum (approved):
Thanks for raising the issue and creating a fix for it! Looks good

dtkav:
Hey @debanjum , thanks for the review.
While this fix makes it _possible_ to use local servers again (I am using it in my fork via [BRAT](https://github.com/TfTHacker/obsidian42-brat) ), I don't think it is ideal.

AFAICT this is still the only network check that updates the connection status setting, and command registration only runs when you load the plugin. This means that you need to configure the plugin via the settings page and then restart obsidian for it to take effect. This is an unexpected flow compared to other obsidian plugins I have used.

Here are some ideas to improve the UX:
1. Automatically test the connection on-edit of the url/api key fields
2. update `settings.connectedToBackend` to `true` on any successful network request to the API.
3. Test the connection in the command handler, perhaps with some debounce rather than relying on a `settings.connectedToBackend` value which could be stale.
4. Allow the search and chat commands to be registered regardless of the flag value and then show the connectivity error message in the search/chat modal.

IMHO it makes sense to:
Immediate term -- merge this so that there is a workaround for the regression.
Short term -- test the network connection and set `settings.connectedToBackend` on edit to the url/api key fields (or with an explicit "test" button).
Longer term --
1. Move the connectivity state management into a wrapper around the network layer
2. Surface the connectivity state in the various modals rather than to inexplicably block command registration/invocation which makes it seem like the installation is broken.

I have only skimmed the code base, so please take the above with a grain of salt.

Thanks for your work on Khoj.

debanjum:
Hey @dtkav, thanks for the well thought-out suggestions 👌🏾. Agree with the immediate, short and long term improvement suggestions for managing backend connectivity state. I'll merge this fix of yours in for now. And look into the short term solution in a bit

PR #560, opened 2023-11-21
Fix ranking search results on Obsidian

debanjum:
This bug was causing the search results on the Obsidian client to be shown in the reverse order of their actual relevance.

It reversed since entry scores returned by Khoj server are a distance metric since the move to Postgres. So lesser distance is better. Previously higher score was better.

### DIFF
--- a/src/interface/obsidian/src/search_modal.ts
+++ b/src/interface/obsidian/src/search_modal.ts
@@ -106,7 +106,7 @@ export class KhojSearchModal extends SuggestModal<SearchResult> {
 
         // Combine markdown and PDF results and sort them by score
         let results = mdData.concat(pdfData)
-            .sort((a: any, b: any) => b.score - a.score)
+            .sort((a: any, b: any) => a.score - b.score)
             .map((result: any) => { return { entry: result.entry, file: result.file } as SearchResult; })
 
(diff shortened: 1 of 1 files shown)


src/interface/obsidian/src/search_modal.ts
```diff
@@ -106,7 +106,7 @@ export class KhojSearchModal extends SuggestModal<SearchResult> {
 
         // Combine markdown and PDF results and sort them by score
         let results = mdData.concat(pdfData)
-            .sort((a: any, b: any) => b.score - a.score)
+            .sort((a: any, b: any) => a.score - b.score)
```
sabaimran, line 109: I think we can remove this & make the query return `All`, rather than pdf/markdown separately. The results should come sorted on the server side.

debanjum: Yeah, valid point. 
It was an initial crude attempt to only show content from current vault. But this hasn't scaled well as other clients also support indexing PDF, Markdown file now. 

I've removed doing search queries for just markdown, pdf content for now.

A more proper solution, for later, would be a user configurable property to limit results to current vault and then actually limiting it via file/directory filters on server/client side.

PR #561, opened 2023-11-21
Clean Logs, Improve Message Rendering and Make Khoj Trusted Host Configurable

debanjum:
- 90d463c1 Append chat message to chat logs as TextNodes in web, desktop clients

- Simplify Code to Identify Files from Github, Notion on Web, Desktop Client
  - befcbcdd Use file source to find entries from github, notion on web, desktop client
  - 3f0de45e Pass file source to clients via text search API response

- Make Django Logs Follow Khoj Log Format, Verbosity
  - 4aec5813 Handle image search setup related warning
  - b06628ee Format Django initializing outputs using Khoj logger format

- 76d041f6 Use `KHOJ_HOST` env var to set allowed/trusted domains to host Khoj

### DIFF
--- a/src/khoj/app/settings.py
+++ b/src/khoj/app/settings.py
@@ -26,13 +26,13 @@ SECRET_KEY = os.getenv("KHOJ_DJANGO_SECRET_KEY")
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.getenv("KHOJ_DEBUG") == "True"
 
-ALLOWED_HOSTS = [".khoj.dev", "localhost", "127.0.0.1", "[::1]", "beta.khoj.dev"]
+# All Subdomains of KHOJ_DOMAIN are trusted
+KHOJ_DOMAIN = os.getenv("KHOJ_DOMAIN", "khoj.dev")
+ALLOWED_HOSTS = [f".{KHOJ_DOMAIN}", "localhost", "127.0.0.1", "[::1]"]
 
 CSRF_TRUSTED_ORIGINS = [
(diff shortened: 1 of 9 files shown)


src/khoj/app/settings.py
```diff
@@ -24,24 +24,24 @@
...
+# All Subdomains of KHOJ_DOMAIN are trusted
+KHOJ_DOMAIN = os.getenv("KHOJ_DOMAIN", "khoj.dev")
+ALLOWED_HOSTS = [f".{KHOJ_DOMAIN}", "localhost", "127.0.0.1", "[::1]"]
 
 CSRF_TRUSTED_ORIGINS = [
-    "https://app.khoj.dev",
-    "https://beta.khoj.dev",
-    "https://khoj.dev",
-    "https://*.khoj.dev",
+    f"https://*.{KHOJ_DOMAIN}",
```
sabaimran, line 34: IIRC, the `*` subdomain didn't work, so I had to explicitly add in `app` and `beta. We should keep `https://app.khoj.dev` as an explicit setting for now.

src/khoj/search_type/image_search.py
```diff
@@ -26,9 +25,6 @@
 
 
 def initialize_model(search_config: ImageSearchConfig):
```
sabaimran, line 27: We should really just delete this until we build better support for this feature 🙈

sabaimran (approved):

src/khoj/search_type/image_search.py
debanjum: Yeah, we really should!

PR #566, opened 2023-11-22
Speak to Khoj via Desktop, Web or Obsidian Client

debanjum:
- cc77bc40 Create speech to text API endpoint
- Use OpenAI Whisper for ASR offline (by downloading Whisper model) or online (via OpenAI API)
- 63675b32  Speak to Khoj from the Desktop client
- 2951fc92 Speak to Khoj from the Web client
- 877532a Speak to Khoj from Obsidian client
- 1ca99b6e Add speech to text model configuration to Database


src/interface/desktop/chat.html
```diff
@@ -377,6 +377,62 @@
...
+            const headers = { 'Authorization': `Bearer ${khojToken}` };
+
+            const sendToServer = (audioBlob) => {
+                const formData = new FormData();
+                formData.append('file', audioBlob);
+
+                fetch(url, { method: 'POST', body: formData, headers})
+                    .then(response => response.ok ? response.json() : Promise.reject(response))
+                    .then(data => { chatInput.value += data.text; })
+                    .catch(err => err.status == 422 ? console.error("Configure speech-to-text model on server.") : console.error("Failed to transcribe audio"));
```
sabaimran, line 399: Should we also update the UI here to indicate to the user that transcript was not successful?

src/interface/desktop/chat.html
```diff
@@ -377,6 +377,62 @@
...
+                const audioChunks = [];
+                const recordingConfig = { mimeType: 'audio/webm' };
+                mediaRecorder = new MediaRecorder(stream, recordingConfig);
+
+                mediaRecorder.addEventListener("dataavailable", function(event) {
+                    if (event.data.size > 0) audioChunks.push(event.data);
+                });
+
+                mediaRecorder.addEventListener("stop", function() {
+                    const audioBlob = new Blob(audioChunks, { type: 'audio/webm' });
```
sabaimran, line 412: do you have to do anything to handle the `mediaRecorder`'s life cycle? like should it be closed or the listener removed?

src/interface/desktop/chat.html
```diff
@@ -377,6 +377,62 @@
...
+                .getUserMedia({ audio: true })
+                .then(handleRecording)
+                .catch((e) => {
+                    console.error(e);
+                });
+            } else if (mediaRecorder.state === 'recording') {
+                mediaRecorder.stop();
+                speakButtonImg.src = './assets/icons/microphone-solid.svg';
+                speakButtonImg.alt = 'Speak';
+            }
```
sabaimran, line 433: Should `mediaRecorder` be defined in one level outside of the scope of the `handleRecording` function? Otherwise it's possible that it would be null here.

src/khoj/database/models/__init__.py
```diff
@@ -120,14 +120,23 @@ class OfflineChatProcessorConversationConfig(BaseModel):
...
+
 class ChatModelOptions(BaseModel):
     class ModelType(models.TextChoices):
         OPENAI = "openai"
         OFFLINE = "offline"
 
     max_prompt_size = models.IntegerField(default=None, null=True, blank=True)
     tokenizer = models.CharField(max_length=200, default=None, null=True, blank=True)
-    chat_model = models.CharField(max_length=200, default=None, null=True, blank=True)
+    chat_model = models.CharField(max_length=200, default="gpt-4")
```
sabaimran, line 139: I don't think the default should be `gpt-4`. It's not certain the admin will set an open AI API key. The default could be the open source model, however, with a default of `ModelType.Offline` below. Otherwise, I'd just continue to leave this field mandatory but without a default.

src/khoj/interface/web/chat.html
```diff
@@ -369,6 +369,56 @@
...
+                });
+
+                mediaRecorder.start();
+                speakButtonImg.src = '/static/assets/icons/stop-solid.svg';
+                speakButtonImg.alt = 'Stop Speaking';
+            };
+
+            // Toggle recording
+            if (!mediaRecorder || mediaRecorder.state === 'inactive') {
+                navigator.mediaDevices
```
sabaimran, line 410: I'm assuming this asks for permissions as well? Should we add any error checking here in case it's not supported by the user's browser & send an alert?

src/khoj/routers/api.py
```diff
@@ -553,6 +556,54 @@ async def chat_options(
...
+            audio_file_writer.write(audio_data)
+        audio_file = open(audio_filename, "rb")
+
+        # Send the audio data to the Whisper API
+        speech_to_text_config = await ConversationAdapters.get_speech_to_text_config()
+        openai_chat_config = await ConversationAdapters.get_openai_chat_config()
+        if not openai_chat_config or not speech_to_text_config:
+            # If the user has not configured a speech to text model, return an unprocessable entity error
+            status_code = 422
+        elif speech_to_text_config.model_type == ChatModelOptions.ModelType.OPENAI:
```
sabaimran, line 580: What happens if the `model_type` is offline? Should we load the whisper model directly in code?

I think this one also makes sense to abstract into separate files, perhaps under `search/khoj/speech` or equivalent? Then add one file for handling offline and one for online, similar to what we do with `chat`.

src/khoj/routers/api.py
```diff
@@ -586,9 +588,9 @@ async def chat_options(
     return Response(content=json.dumps(cmd_options), media_type="application/json", status_code=200)
 
 
-@api.post("/speak")
+@api.post("/transcribe")
 @requires(["authenticated"])
-async def transcribe_audio(request: Request, common: CommonQueryParams, file: UploadFile = File(...)):
+async def transcribe(request: Request, common: CommonQueryParams, file: UploadFile = File(...)):
```
sabaimran, line 593: Can you add a small decorator/dependency here that limits the size of the incoming audio packet? The limit we use in Flint is 10 MB: https://github.com/khoj-ai/flint/commit/b32c16be48be6f90c28f406416e8740b1ce0cd19#diff-8af27642228981ffa9a2c39239d7cbdfaa9f969f9c187be165c4cedc57b8c5f1R182-R194

src/khoj/utils/config.py
```diff
@@ -80,7 +80,7 @@ class GPT4AllProcessorConfig:
 class GPT4AllProcessorModel:
     def __init__(
         self,
-        chat_model: str = "llama-2-7b-chat.ggmlv3.q4_0.bin",
+        chat_model: str = "mistral-7b-instruct-v0.1.Q4_0.gguf",
```
sabaimran, line 83: good catch

sabaimran (approved):

src/khoj/interface/web/chat.html
```diff
@@ -543,20 +543,52 @@
             }
         }
 
+        function flashStatusInChatInput(message) {
+            // Get chat input element and original placeholder
+            let chatInput = document.getElementById("chat-input");
+            let originalPlaceholder = chatInput.placeholder;
+            // Set placeholder to message
+            chatInput.placeholder = message;
```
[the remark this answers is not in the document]
debanjum: Yeah, good point.. I'll let this be for now, we can update it to updat a different status `div` later

src/khoj/routers/api.py
debanjum: Good point. Added 10Mb file size limit allowed on the transcribe API endpoint

PR #577, opened 2023-12-03
Fix null exception when user does not exist for subscription

sabaimran:
If a user subscribes to Khoj with an email address that's not present in the DB, create an account with the corresponding email address and then update the Subscription object.

Bear in mind that this flow will cause some hiccups for users whose email addresses are not associated with Google accounts. So when they try to log in, their account may not be associated with their subscription. We should find better workarounds to avoid this.


debanjum:
To reduce chances of users paying via a non-supported email account is to link to https://app.khoj.dev/config?#billing from the https://khoj.dev/pricing page
1. This would require users to create their account via a supported email account on khoj.dev before trying to pay
2. The email address associated with their payment by default will be set to their Khoj associated email account

But for that to work, we need to fix post-login redirects on app.khoj.dev. Created #578 to track that

PR #587, opened 2023-12-20
Support multiple search models, with ability for custom user config

sabaimran:
# Incoming
- Add support for using multiple search models from huggingface within Khoj. 
- Change the `VectorField` field in the `Entry` model to have a `None` dimension in order to support variable generated embeddings
- Add a cute temporary notification banner to relay messages on the config page
- In Khoj Cloud's use case, this will enable users to use a multilingual model, hence why the configuration section is labeled "Language"


src/khoj/routers/api.py
```diff
@@ -332,6 +332,31 @@ async def update_chat_model(
...
+    update_telemetry_state(
+        request=request,
+        telemetry_type="api",
+        api="set_search_model",
+        client=client,
+        metadata={"search_model": new_config.setting.name},
+    )
+
+    if new_config is None:
+        return {"status": "error", "message": "Model not found"}
```
debanjum, line 355: Should the `update_telemetry_state` be done only after/if new_config is not None

src/khoj/routers/api.py
```diff
@@ -410,13 +435,10 @@ async def search(
...
 
     encoded_asymmetric_query = None
-    if t == SearchType.All or t != SearchType.Image:
-        text_search_models: List[TextSearchModel] = [
-            model for model in state.search_models.__dict__.values() if isinstance(model, TextSearchModel)
-        ]
-        if text_search_models:
-            with timer("Encoding query took", logger=logger):
-                encoded_asymmetric_query = state.embeddings_model.embed_query(defiltered_query)
+    if t != SearchType.Image:
```
debanjum, line 438: Don't we still need to set the `encoded_asymmetric_query` when `t == SearchType.All`?

debanjum (approved):

src/khoj/routers/api.py
sabaimran: That should still be true with the removed condition, but yes.

debanjum: Ah, I see. Makes sense

Issue #570, opened 2023-11-26
Add a health check to Khoj


sabaimran:
In `api.py`, add a simple `/health` endpoint that simply returns `200`.

Sanket-Arekar:
Hey @sabaimran I would like to work on this issue. Can you Please assign me this issue?

sabaimran:
Awesome! Thanks for volunteering, @Sanket-Arekar . Let me know if you need any guidance with the dev setup.

I would recommend cloning the repository, creating a virtual environment, and then running `pip install -e '.[dev]` to get started. We also need to add better documentation for this.

Sanket-Arekar:
Okay. Thank You for assigning me this issue, I will start working on it.

ayushjha119:
@sabaimran please check now, i have raised new PR according to your requirement

sabaimran:
Thank for taking care of this @ayushjha119 ! I'm updating the developer documentation based on your experience in this PR: https://github.com/khoj-ai/khoj/pull/596. Let me know if there's anything you'd like to add.

ayushjha119:
@sabaimran Thank you soo much, for helping in my first Open source contribution

PR #590, opened 2023-12-20
fix:#570 added health endpoint that simply returns 200

ayushjha119:
Fixed this
In api.py, add a simple /health endpoint that simply returns 200.

### DIFF
--- a/src/khoj/routers/api.py
+++ b/src/khoj/routers/api.py
@@ -862,3 +862,8 @@ async def extract_references_and_questions(
         compiled_references = [item.additional["compiled"] for item in result_list]
 
     return compiled_references, inferred_queries, defiltered_query
+
+    # health check
+@app.get("/health", status_code=200)
+async def health_check():
+    return {"status": 200}


src/khoj/routers/api.py
```diff
@@ -862,3 +862,8 @@ async def extract_references_and_questions(
         compiled_references = [item.additional["compiled"] for item in result_list]
 
     return compiled_references, inferred_queries, defiltered_query
+
+    # health check
```
sabaimran, line 866: nit: can you remove this comment? the api definition is pretty self-explanatory 🙂 .

src/khoj/routers/api.py
```diff
@@ -862,3 +862,8 @@ async def extract_references_and_questions(
...
+@app.get("/health", status_code=200)
+async def health_check():
+    return {"status": 200}
```
sabaimran, line 869: Could you use the `Response` class here from `fastapi.responses` and set the input for `status_code` to 200?

sabaimran (commented):
Looks good for the most part! Just a couple of clean-up comments.

sabaimran:
Closing as you've created a new PR #591 .

PR #607, opened 2024-01-11
Short-circuit API rate limiter for unauthenticated users

debanjum:
### Major
- ba99089a Short-circuit API rate limiter for unauthenticated user
  Calls by unauthenticated users were failing at API rate limiter as it
  failed to access user info object. This is a bug.
  
  API rate limiter should short-circuit for unauthenicated users so a
  proper Forbidden response can be returned by API
  
  Add regression test to verify that unauthenticated users get 403
  response when calling the /chat API endpoint
  
### Minor
- b1269fda Remove trailing slash to normalize khoj url in obsidian plugin settings
- bb1c1b39 Move /api/config API controllers into separate module
- 5f97357f Delete unused /api/beta API endpoint
- ffdb291f Fix error message rendering in khoj.el, khoj obsidian chat
- 7dfbcd2e Handle deprecation warnings for subscribe renew date, langchain, pydantic & logger.warn


src/khoj/routers/api_config.py
```diff
@@ -0,0 +1,359 @@
...
+def get_config_data(request: Request):
+    user = request.user.object
+    EntryAdapters.get_unique_file_types(user)
+
+    return state.config
+
+
+@api_config.post("/data")
+@requires(["authenticated"])
+async def set_config_data(
```
sabaimran, line 111: I think we actually don't need this method anymore. This was from earlier when we had an application design which assumed the server had access to local files, so clients would directly update the configuration for running the local indexing job. Method can be removed.

src/khoj/routers/api_config.py
```diff
@@ -0,0 +1,359 @@
...
+        telemetry_type="api",
+        api="set_content_config",
+        client=client,
+        metadata={"content_type": "notion"},
+    )
+
+    return {"status": "ok"}
+
+
+@api_config.delete("/data/content-source/{content_source}", status_code=200)
```
sabaimran, line 206: This also isn't being used anymore.

src/khoj/routers/api_config.py
```diff
@@ -0,0 +1,359 @@
...
+            api="set_search_model",
+            client=client,
+            metadata={"search_model": new_config.setting.name},
+        )
+
+    return {"status": "ok"}
+
+
+# Create Routes
+@api_config.get("/data/default")
```
sabaimran, line 325: Not sure if this is still being used by Obsidian, Emacs but I don't think so.

tests/test_client.py
```diff
@@ -461,6 +461,20 @@ def test_user_no_data_returns_empty(client, sample_org_data, api_user3: KhojApiU
     assert response.json() == []
 
 
+@pytest.mark.django_db(transaction=True)
+def test_chat_with_unauthenticated_user(chat_client_with_auth, api_user2: KhojApiUser):
+    # Arrange
+    headers = {"Authorization": f"Bearer {api_user2.token}"}
```
sabaimran, line 467: This seems to be failing with an internal server 500 error

src/khoj/routers/api_config.py
debanjum: This is being used in the web interface to delete data for each content-source (desktop, github or notion)

tests/test_client.py
debanjum: Yeah, investigating that. Not sure why it's failing in the workflow but not locally on my machine

src/khoj/routers/api_config.py
debanjum: This was being used by a deprecated method in the khoj emacs client. I've deleted that code and this `/api/config/data` API endpoint

tests/test_client.py
debanjum: Never mind, the issue was trivial. We weren't checking for required API key environment variable before trying to run the test. It's resolves now

PR #609, opened 2024-01-13
Add option to use HuggingFace's inference endpoint for generating embeddings

sabaimran:
# Incoming
- To make it easier for users to have an alternative to using on-edge computing for generating embeddings, provide the option of using one of the hugging face inference endpoints for each search model config.
- You can see documentation for huggingface endpoints [here](https://huggingface.co/docs/inference-endpoints/index)
- We're prioritizing endpoints for huggingface over openai, for instance, because the open source models provide many high quality options


debanjum (approved):
Left some comments but looks great apart from that!

PR #616, opened 2024-01-17
Allow custom inference endpoint for the crossencoder model

sabaimran:
# Incoming
- Permit use of a custom hugging face inference endpoint for generating scores from the `CrossEncoder`. I had to setup a custom model/handler to support this exact use case. See here: https://huggingface.co/sabaimran/ms-marco-MiniLM-L-6-v2.
- Update some langchain imports


debanjum (approved):
Looks good!

PR #617, opened 2024-01-17
Improve server status check and message in Obsidian client

debanjum:
- 870af19b Update health API to pass authenticated users their info
- fe8b7f18 Improve Khoj server status check in Khoj Obsidian client
- 8d4155be Show Khoj Obsidian commands even if no connection to server
- 461a925f Show Khoj chat by default in Obsidian side pane instead of search

### DIFF
--- a/src/interface/obsidian/src/utils.ts
+++ b/src/interface/obsidian/src/utils.ts
@@ -1,4 +1,4 @@
-import { FileSystemAdapter, Notice, Vault, Modal, TFile } from 'obsidian';
+import { FileSystemAdapter, Notice, Vault, Modal, TFile, request } from 'obsidian';
 import { KhojSetting } from 'src/settings'
 
 export function getVaultAbsolutePath(vault: Vault): string {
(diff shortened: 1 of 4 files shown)


src/interface/obsidian/src/utils.ts
```diff
@@ -123,3 +123,50 @@ export async function createNoteAndCloseModal(query: string, modal: Modal, opt?:
...
+}
+
+export function getBackendStatusMessage(
+    connectedToServer: boolean,
+    userEmail: string,
+    khojUrl: string,
+    khojApiKey: string
+): string {
+    // Welcome message with default settings. Khoj cloud always expects an API key.
+    if (khojApiKey === '' && khojUrl === 'https://app.khoj.dev')
```
sabaimran, line 160: You'll also want to make sure `khojApiKey` is not `null`, `undefined`, in addition to being empty string. Below should take care of it.

```suggestion
    if (khojApiKey && khojUrl === 'https://app.khoj.dev')
```

sabaimran (approved):

src/interface/obsidian/src/utils.ts
debanjum: Fixed that by forcing it to be a boolean with `!!`

PR #626, opened 2024-01-24
Improve subqueries for online search and prompt generation for image

sabaimran:
- Include conversation history so that subqueries or intermediate prompts are generated with the appropriate context


debanjum (approved):
Great to have image gen and online search use conversation history

PR #627, opened 2024-01-26
Scrape results from Serper results using Olostep

sabaimran:
- Use [Olostep](https://www.olostep.com/) to go through top results from Serper's google search
- Add an intermediate step that summarizes extracted web results for further response downstream
- If `answerBox` is present in the query results, no need to do the additional scraping
- Add truncation logic to the model wrapper

### DIFF
--- a/src/khoj/processor/conversation/utils.py
+++ b/src/khoj/processor/conversation/utils.py
@@ -15,8 +15,10 @@ from khoj.utils.helpers import merge_dicts
 
 logger = logging.getLogger(__name__)
 model_to_prompt_size = {
-    "gpt-3.5-turbo": 4096,
-    "gpt-4": 8192,
+    "gpt-3.5-turbo": 3000,
+    "gpt-4": 7000,
+    "gpt-4-1106-preview": 7000,
+    "gpt-4-turbo-preview": 4000,
     "llama-2-7b-chat.ggmlv3.q4_0.bin": 1548,
     "gpt-3.5-turbo-16k": 15000,
     "mistral-7b-instruct-v0.1.Q4_0.gguf": 1548,
@@ -200,7 +202,7 @@ def truncate_messages(
         logger.debug(
             f"Truncate current message to fit within max prompt size of {max_prompt_size} supported by {model_name} model:\n {truncated_message}"
         )
-        messages = [ChatMessage(content=truncated_message + original_question, role=messages[0].role)]
+        messages = [ChatMessage(content=truncated_message + "\n" + original_question, role=messages[0].role)]
 
     return messages + [system_message]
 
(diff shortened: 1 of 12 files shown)


src/khoj/processor/conversation/utils.py
```diff
@@ -194,6 +196,7 @@ def truncate_messages(
         assert type(system_message.content) == str
         current_message = "\n".join(messages[0].content.split("\n")[:-1]) if type(messages[0].content) == str else ""
         original_question = "\n".join(messages[0].content.split("\n")[-1:]) if type(messages[0].content) == str else ""
+        original_question = f"\n{original_question}"
```
debanjum, line 199: Why add a new line before the `original_question`?

src/khoj/routers/helpers.py
```diff
@@ -175,35 +195,41 @@ async def generate_better_image_prompt(q: str, conversation_history: str) -> str
...
             state.gpt4all_processor_config = GPT4AllProcessorModel(conversation_config.chat_model)
 
         loaded_model = state.gpt4all_processor_config.loaded_model
         return send_message_to_model_offline(
-            message=message,
+            message=truncated_messages[-1].content,
             loaded_model=loaded_model,
             model=conversation_config.chat_model,
             streaming=False,
+            system_message=truncated_messages[0].content,
```
debanjum, line 219: Is system message always the first of the `truncated_messages`? Why not just set system_message to system_message?

```suggestion
            system_message=system_message,
```

debanjum:
Can we add chat actor, director tests to verify the functioning of this /online with website crawl flows? It's getting a little complicated. It'd be good to see what kind of stuff Khoj can do with this vs without it (It'll improve insight into our chatquality benchmarking via pytests)

src/khoj/processor/tools/online_search.py
```diff
@@ -51,4 +70,41 @@ def _search_with_google(subquery: str):
...
+    for subquery in response_dict:
+        # If a high quality answer is not found, search the web pages of the first 3 organic results
+        if is_none_or_empty(response_dict[subquery].get("answerBox")):
+            extracted_content[subquery] = []
+            for result in response_dict[subquery].get("organic")[:1]:
+                logger.info(f"Searching web page of '{result['link']}'")
+                try:
+                    extracted_content[subquery].append(search_with_olostep(result["link"]).strip())
+                except Exception as e:
+                    logger.error(e, exc_info=True)
```
debanjum, line 87: Maybe good to prefix a description of the error for context
```suggestion
                    logger.error(f'Failed to use Olostep to search through {result["link"]}.\nException: {e}", exc_info=True)
```

debanjum (approved):
This change will be super useful for deeper insights from online results! It'd be great if we can add some chat quality tests to evaluate Khoj's improved capabilities with these changes

src/khoj/processor/conversation/utils.py
sabaimran: It's getting concatenated at the end without a new line if it's not explicitly added, which is why I'm adding it in here (before truncation).

src/khoj/routers/helpers.py
sabaimran: We would probably want to leave an option for the truncation logic to act as an underlying abstraction layer to handle longer messages. I wouldn't want to override its decision if/when we apply some truncation logic to the system prompt.

PR #628, opened 2024-01-28
Add retries in case the embeddings API fails

sabaimran:
# Incoming
Sometimes the embeddings API fails if the server is busy. Add retries in case it doesn't work the first time.

### DIFF
--- a/src/khoj/processor/embeddings.py
+++ b/src/khoj/processor/embeddings.py
@@ -4,6 +4,14 @@ from typing import List
 import requests
 import tqdm
 from sentence_transformers import CrossEncoder, SentenceTransformer
+from tenacity import (
+    before_sleep_log,
+    retry,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_exponential,
+    wait_random_exponential,
+)
 from torch import nn
 
 from khoj.utils.helpers import get_device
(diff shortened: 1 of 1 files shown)


src/khoj/processor/embeddings.py
```diff
@@ -28,13 +36,30 @@ def __init__(
...
-            target_url = f"{self.inference_endpoint}"
-            payload = {"inputs": [query]}
-            headers = {"Authorization": f"Bearer {self.api_key}", "Content-Type": "application/json"}
-            response = requests.post(target_url, json=payload, headers=headers)
-            return response.json()["embeddings"][0]
+            return self.embed_with_api([query])[0]
         return self.embeddings_model.encode([query], show_progress_bar=False, **self.encode_kwargs)[0]
 
+    @retry(
+        retry=retry_if_exception_type(Exception),
```
debanjum, line 43: Do we want to limit retry to when there is an HTTPError? This would avoid retrying on non-recoverable errors (like a bug in our code, or a change in their API's response format)

```suggestion
        retry=retry_if_exception_type(requests.exceptions.HTTPError),
```

debanjum (approved):
Good idea to safeguard against the embedding service API failures. Left some comments but changes looks good otherwise!

PR #629, opened 2024-01-28
Store rate limiter-related metadata in the database for more resilience

sabaimran:
# Incoming
- Add a new database model `UserRequests` to keep state of the number of requests made per API/window per user
- This helps maintain state even between server restarts
- Allows you to scale up workers on your service without having to implement sticky routing
- We could pair this with an in-memory cache if latency becomes an issue


src/khoj/configure.py
```diff
@@ -328,3 +329,9 @@ def upload_telemetry():
         logger.error(f"📡 Error uploading telemetry: {e}", exc_info=True)
     else:
         state.telemetry = []
+
+
+@schedule.repeat(schedule.every(31).minutes)
```
debanjum, line 334: Why not repeat this every ~1 day (22-26 hours)?

src/khoj/database/migrations/0029_userrequests.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="UserRequests",
+            fields=[
+                ("id", models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name="ID")),
+                ("created_at", models.DateTimeField(auto_now_add=True)),
+                ("updated_at", models.DateTimeField(auto_now=True)),
+                ("request", models.CharField(max_length=200)),
```
debanjum, line 20: What is the purpose of the `request` Char field and the `updated_at` DateTimeField ? They don't seem to be used anywhere

debanjum (approved):
Looks good!

src/khoj/configure.py
sabaimran: We'd want to flush it more frequently to give as accurate a model of usage as we can. For example, if a user made a request just after the schedule job ran to clear the cache, they would have to wait almost 48 hours for that request to clear from the DB.

src/khoj/database/migrations/0029_userrequests.py
sabaimran: `created_at` and `updated_at` are added to all of our models and managed independently. But you're right, the `request` field is superfluous. Will remove that.

Issue #632, opened 2024-01-30
[FIX] chat


Anandsangat:
Sir, why is there no chat from today, when will the problem be solved?

PR #633, opened 2024-01-30
Fix subscription state detection for users based on phone numbers, emails

sabaimran:
# Incoming
- Correctly use the code to get the current subscription state


sabaimran:
Closes #632

PR #639, opened 2024-02-06
Add support for multiple chat sessions in the desktop application

sabaimran:
# Incoming
Given #638 , add the same treatment to the desktop application. Allow users to create multiple chat sessions to maintain parallel conversations with Khoj.


debanjum (approved):
Looks good! Can you attach screenshots of the chat view with the chat sessions pane, would be good to document/look at

PR #641, opened 2024-02-07
Have Khoj dynamically select conversation command(s) in chat

sabaimran:
# Incoming
- Intercept the commands if in default mode, and have Khoj dynamically guess which tools would be the most relevant for answering the user's query
- Choose from `online`, `general`, `notes`, and `default` to start with, as these would be for information getting, while `image` is more for response modality
- Add unit tests


src/khoj/utils/helpers.py
```diff
@@ -282,6 +282,13 @@ class ConversationCommand(str, Enum):
     ConversationCommand.Help: "Display a help message with all available commands and other metadata.",
 }
 
+tool_descriptions_for_llm = {
+    ConversationCommand.Default: "Use this if there might be a mix of general and personal knowledge in the question",
+    ConversationCommand.General: "Use this when you can answer the question without any outside information or personal knowledge",
```
debanjum, line 287: Minor suggestion to improve the general command. Unsure if it's better
```suggestion
    ConversationCommand.General: "Use this when you can answer the question without needing any additional online or personal information",
```

src/khoj/processor/conversation/prompts.py
```diff
@@ -285,6 +290,54 @@
 """.strip()
 )
 
+pick_relevant_tools = PromptTemplate.from_template(
+    """
+You are Khoj, a smart and helpful personal assistant. You have access to a variety of tools to help you answer the user's question. You can use the following tools to collect more relevant information. You can use any combination of these tools to answer the user's question. Tell me which tools you would like to use to answer the user's question.
```
debanjum, line 295: Should we frame this more specifically as pick relevant data sources?

```suggestion
pick_relevant_information_sources = PromptTemplate.from_template(
    """
You are Khoj, a smart and helpful personal assistant. You have access to a variety of data sources to help you answer the user's question. You can use the data sources listed below to collect more relevant information. You can use any combination of these data sources to answer the user's question. Tell me which data sources you would like to use to answer the user's question.
```

src/khoj/routers/helpers.py
```diff
@@ -128,15 +141,50 @@ async def agenerate_chat_response(*args):
     return await loop.run_in_executor(executor, generate_chat_response, *args)
 
 
+async def aget_relevant_tools(query: str, conversation_history: dict):
+    """
+    Given a query, determine which of the available tools the agent should use in order to answer appropriately.
+    """
```
debanjum, line 147: Should we reframe this more specifically, as get information from relevant data sources (tools sounds too generic IMO)

```suggestion
async def aget_relevant_data_sources(query: str, conversation_history: dict):
    """
    Given a query, determine which of the available data sources the agent should use in order to answer appropriately.
    """
```

src/khoj/routers/helpers.py
```diff
@@ -105,6 +109,15 @@ def update_telemetry_state(
     ]
 
 
+def construct_chat_history(conversation_history: dict, n: int = 4) -> str:
+    chat_history = ""
+    for chat in conversation_history.get("chat", [])[-n:]:
+        if chat["by"] == "khoj" and chat["intent"].get("type") == "remember":
```
debanjum, line 115: We should (at some point) set the intent type based on command used (e.g "online-information" vs "remember" vs "draw"). The schema hasn't aged well (wasn't great earlier either) but the point is to store the user intent inferred by Khoj for each chat message to respond appropriately

debanjum (approved):
Super excited to have Khoj be able to dynamically choose the data sources (and output format soon?). Having to type in the slash command every time was getting fairly annoying 😅

src/khoj/processor/conversation/offline/chat_model.py
```diff
@@ -148,27 +148,24 @@ def converse_offline(
...
         for result in online_results:
             if online_results[result].get("extracted_content"):
                 simplified_online_results[result] = online_results[result]["extracted_content"]
 
-        conversation_primer = prompts.online_search_conversation.format(
-            query=user_query, online_results=str(simplified_online_results)
-        )
-    elif conversation_command == ConversationCommand.General or is_none_or_empty(compiled_references_message):
-        conversation_primer = user_query
-    else:
```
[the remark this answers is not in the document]
sabaimran: It just doesn't need to be differentiated, because `conversation_primer` is being set to the `user_query` already, then conditionally being built on top of.

PR #644, opened 2024-02-10
Add additional personalization in Chat via Location, Username

sabaimran:
# Incoming
- Inject location data and username data into the Khoj chat context for greater personalization
- Make username configurable in the profile page


src/khoj/interface/web/config.html
```diff
@@ -3,6 +3,25 @@
...
+        <div class="section-cards">
+            <div class="card">
+                <div class="card-title-row">
+                    <img class="card-icon" src="/static/assets/icons/user-silhouette.svg" alt="Chat">
+                    <h3 class="card-title">
+                        Name
+                    </h3>
+                </div>
+                <div class="card-description-row">
+                    <input type="text" id="profile_given_name" class="form-control" placeholder="Michael Jordan" value="{{ given_name }}">
```
debanjum, line 16: Should placeholder be set to their current/default username (=username associated with gmail address). It can also be an instruction like "Enter your name"?
```suggestion
                    <input type="text" id="profile_given_name" class="form-control" placeholder="Michael Jordan" value="{{ given_name }}">
```

src/khoj/processor/conversation/openai/gpt.py
```diff
@@ -32,6 +33,8 @@ def extract_questions(
     def _valid_question(question: str):
         return not is_none_or_empty(question) and question != "[]"
 
+    location = f"{location_data.city}, {location_data.state}, {location_data.country}" if location_data else "Unknown"
```
debanjum, line 36: Should this be set to empty string if `location_data` is not available like what's been done for offline chat? The user may provide their location via a chat message or notes. Having "location = unknown" in the prompt may confuse (some) chat models

src/khoj/processor/conversation/prompts.py
```diff
@@ -122,6 +122,9 @@
     """
 You are a talented creator. Generate a detailed prompt to generate an image based on the following description. Update the query below to improve the image generation. Add additional context to the query to improve the image generation. Make sure to retain any important information from the query. Use the conversation log to inform your response.
 
+Today's Date: {current_date}
+Location: {location}
```
debanjum, line 126: How do you see today's date and user's location being used by the image generator?

src/khoj/processor/conversation/prompts.py
```diff
@@ -215,6 +218,7 @@
 What searches, if any, will you need to perform to answer the users question?
 Provide search queries as a JSON list of strings
 Current Date: {current_date}
+Location: {location}
```
debanjum, line 221:
```suggestion
User's Location: {location}
```

src/khoj/processor/conversation/prompts.py
```diff
@@ -349,6 +353,7 @@
 What Google searches, if any, will you need to perform to answer the user's question?
 Provide search queries as a list of strings
 Current Date: {current_date}
+Location: {location}
```
debanjum, line 356:
```suggestion
User's Location: {location}
```

src/khoj/routers/api_chat.py
```diff
@@ -219,6 +220,9 @@ async def chat(
     stream: Optional[bool] = False,
     slug: Optional[str] = None,
     conversation_id: Optional[int] = None,
+    city: Optional[str] = None,
+    region: Optional[str] = None,
+    country: Optional[str] = None,
```
debanjum, line 225: Should we just pass a single location parameter from the clients? It can be set to `location = city, region, country`?

src/khoj/routers/web_client.py
```diff
@@ -166,6 +168,7 @@ def config_page(request: Request):
             "current_model_state": successfully_configured,
             "anonymous_mode": state.anonymous_mode,
             "username": user.username,
```
debanjum, line 170: Isn't the username currently always set to the user's email address? If so, should we just use that for the user name (instead of adding the new given_name field) or populate the `username` field with the username from gmail to make the field more useful?

src/khoj/routers/api_config.py
```diff
@@ -290,6 +290,38 @@ async def get_indexed_data_size(request: Request, common: CommonQueryParams):
...
+def set_user_name(
+    request: Request,
+    name: str,
+    client: Optional[str] = None,
+):
+    user = request.user.object
+
+    split_name = name.split(" ")
+
+    if len(split_name) > 2:
```
debanjum, line 304: What about folks with a middle name(s)? 😄 I think in that case the `first_name = ' '.join(split_name[:-1]) `and `last name = split_name[-1]` or vice versa (i.e `first_name = split_name[0]` and `last_name = ' '.join(split_name[1:])`)

src/khoj/interface/web/chat.html
```diff
@@ -345,7 +361,7 @@
             }
 
             // Generate backend API URL to execute query
-            let url = `/api/chat?q=${encodeURIComponent(query)}&n=${resultsCount}&client=web&stream=true&conversation_id=${conversationID}`;
+            let url = `/api/chat?q=${encodeURIComponent(query)}&n=${resultsCount}&client=web&stream=true&conversation_id=${conversationID}&region=${region}&city=${city}&country=${countryName}`;
```
debanjum, line 364: Maybe should send location data from Obsidian as well?

debanjum (commented):
Left some comments. Looking forward for the added username, location context to make the responses more personalized!

src/khoj/interface/web/config.html
sabaimran: Given name is a computed value which would be the default user name if an explicit one isn't provided. See method named `get_user_name`.

src/khoj/processor/conversation/prompts.py
sabaimran: Something like `paint a picture of where i am now` might be relevant? Like depicting current city (e.g., `Shanghai`) in the current season (e.g., `winter`)? Edge case, but I can imagine it being relevant.

src/khoj/routers/api_chat.py
sabaimran: I think it's better to keep separate as some clients wouldn't have access to all three parameters. For example, the WhatsApp client would only potentially have access to `country`. I think it makes the format easier for the clients to have each param separate.

src/khoj/routers/web_client.py
sabaimran: The usage of `username` is semantically more for unique identification, rather than common name. I'll leave as-is.

PR #645, opened 2024-02-10
Update Text Chunking Strategy to Improve Search Context

### LINKED ISSUE #620: [FIX] Splitting the entry into words solely by spaces results in the omission of content
## Describe the bug

Splitting the entry into words solely by spaces and discarding long words can result in the omission of meaningful content from the corpus, especially when dealing with CJK languages.

## To Reproduce

Create a text file, in every line, except for the first line, write a long word without any spaces. Repeat these lines to make the total length of the file longer than 500 characters.

```
First Line
dog=5
cat=10
apple=3
car=7
book=2
tree=9
pen=4
chair=8
phone=6
lamp=1
...more_lines_without_space...
```

Index this file, the `compiled` column contain only a single word "First".

## Additional context

It's unusual to have such text in Latin languages, but it's possible in Chinese/Japanese where words are not separated by spaces. There is no space in the following text:

```
Khoj是一个开源的AI个人助手，通过索引你的个人文本或图像数据来工作。你可以自行托管Khoj，并开始搜索和与你的组织、Markdown、PDF笔记聊天。Khoj正在努力帮助你充分利用你的数据。

需求人群：
用于搜索和管理个人知识库，支持搜索和与你的组织、Markdown、PDF笔记聊天等功能

产品特色：
快速语义搜索；
聊天功能；
自定义扩展插件；
```

The issue is caused by the code in [khoj/src/khoj/processor/content/text_to_entries.py](https://github.com/khoj-ai/khoj/blob/039ed78253f03b7c780b26a9712a84375e5c0726/src/khoj/processor/content/text_to_entries.py)
```py
            # Split entry into words
            compiled_entry_words = [word for word in entry.compiled.split(" ") if word != ""]

            # Drop long words instead of having entry truncated to maintain quality of entry processed by models
            compiled_entry_words = [word for word in compiled_entry_words if len(word) <= max_word_length]
```

A possible workaround could involve splitting the entry into words not only by spaces but also by new lines. A good solution for this could be using langchain's `RecursiveCharacterTextSplitter`.

debanjum:
## Major
- 0336107f Parse markdown parent entries as single entry if fit within max tokens
- 43aaaf20 Parse markdown file as single entry if it fits with max token limits
- c093a46b Add parent heading ancestory to extracted markdown entries for context
- 73c1cc18 Chunk text in preference order of para, sentence, word, character

## Minor
- 83801f83 Create wrapper function to get entries from org, md, pdf & text files
- ed6e5d1b Remove unused Entry to Jsonl converter from text to entry class, tests
- 3a103ea2 Dedupe code by using single func to process an org file into entries

Resolves #620

### DIFF
--- a/src/khoj/processor/content/markdown/markdown_to_entries.py
+++ b/src/khoj/processor/content/markdown/markdown_to_entries.py
@@ -87,7 +87,7 @@ class MarkdownToEntries(TextToEntries):
 
         # If content is small or content has no children headings, save it as a single entry
         if len(TextToEntries.tokenizer(markdown_content_with_ancestry)) <= max_tokens or not re.search(
-            rf"^#{{{len(ancestry)+1},}}\s", markdown_content, re.MULTILINE
+            rf"^#{{{len(ancestry)+1},}}\s", markdown_content, flags=re.MULTILINE
         ):
             entry_to_file_map += [(markdown_content_with_ancestry, markdown_file)]
             entries.extend([markdown_content_with_ancestry])
@@ -98,7 +98,7 @@ class MarkdownToEntries(TextToEntries):
         sections: List[str] = []
         while len(sections) < 2:
             next_heading_level += 1
-            sections = re.split(rf"(\n|^)(?=[#]{{{next_heading_level}}} .+\n?)", markdown_content, re.MULTILINE)
+            sections = re.split(rf"(\n|^)(?=[#]{{{next_heading_level}}} .+\n?)", markdown_content, flags=re.MULTILINE)
 
         for section in sections:
             # Skip empty sections
--- a/src/khoj/processor/content/markdown/markdown_to_entries.py
+++ b/src/khoj/processor/content/markdown/markdown_to_entries.py
@@ -32,10 +32,8 @@ class MarkdownToEntries(TextToEntries):
             deletion_file_names = None
 
         # Extract Entries from specified Markdown files
-        with timer("Parse entries from Markdown files into dictionaries", logger):
-            current_entries = MarkdownToEntries.convert_markdown_entries_to_maps(
-                *MarkdownToEntries.extract_markdown_entries(files)
-            )
+        with timer("Extract entries from specified Markdown files", logger):
+            current_entries = MarkdownToEntries.extract_markdown_entries(files)
 
         # Split entries by max tokens supported by model
         with timer("Split entries by max token size supported by model", logger):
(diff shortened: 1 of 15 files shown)


src/khoj/processor/content/org_mode/org_to_entries.py
```diff
@@ -57,93 +54,175 @@ def process(
...
+                if parsed_entry.scheduled:
+                    compiled += f'\n Scheduled for {parsed_entry.scheduled.strftime("%Y-%m-%d")}.'
+                    if state.verbose > 2:
+                        logger.debug(f'Scheduled: {parsed_entry.scheduled.strftime("%Y-%m-%d")}')
+
+                if parsed_entry.hasBody:
+                    compiled += f"\n {parsed_entry.body}"
+                    if state.verbose > 2:
+                        logger.debug(f"Body: {parsed_entry.body}")
+
```
sabaimran, line 211: Do you think all these debug lines are useful to keep?

Is the reason you're explicitly adding the closed, scheduled, etc lines so that it can be preserved through all chunks of the entry?

tests/test_markdown_to_entries.py
```diff
@@ -79,19 +66,139 @@ def test_multiple_markdown_entries_to_jsonl(tmp_path):
...
+    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=3)
 
     # Assert
-    assert len(jsonl_data) == 2
+    assert len(entries) == 2
     # Ensure entry compiled strings include the markdown files they originate from
     assert all([tmp_path.stem in entry.compiled for entry in entries])
 
 
+def test_extract_entries_with_different_level_headings(tmp_path):
```
sabaimran, line 77: Love having complete unit tests! It can be really hard to pinpoint indexing-related issues otherwise.

sabaimran (commented):
Will do some local testing with generated data, but looks good!

src/khoj/processor/content/org_mode/org_to_entries.py
debanjum: No, they're not required. They require extreme verbose mode to trigger (i.e `-vvv` flag). Removing it for now.

> Is the reason you're explicitly adding the closed, scheduled, etc lines so that it can be preserved through all chunks of the entry?

Yes, exactly.

src/khoj/processor/content/markdown/markdown_to_entries.py
```diff
@@ -31,15 +30,14 @@ def process(
         else:
             deletion_file_names = None
 
+        max_tokens = 256
```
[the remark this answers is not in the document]
debanjum: Yeah, ideally should be coming from model, more so now that there are search models with larger context size (previously most supported 512 max tokens because that's default for BERT). But the models don't currently expose this. We should add this to the `SearchModelConfig` to make it configurable by admins

Issue #637, opened 2024-02-05
[FIX] notion import not working


danrthompson:
## Describe the bug
A clear and concise description of what the bug is. Please include what you were expecting to happen vs. what actually happened.

when i try to setup notion by adding my token, adding the connection to a page (i tried a demo page with basically no contents and no databases), then hitting reinitialize, the loading of the notion contents fails

the demo page has title "Test" and contents "Testing!"

## To Reproduce
install the server on a mac with docker compose (using the 1.5.0, 1.4.0, or 1.1.0 backend version), then create notion integration, add notion connection to a page, save token to backend, and reinitialize

## Platform
- Server:
    - [ ] Cloud-Hosted (https://app.khoj.dev)
    - [x] Self-Hosted Docker
    - [ ] Self-Hosted Python package
    - [ ] Self-Hosted source code
- Client:
    - [ ] Obsidian
    - [ ] Emacs
    - [ ] Desktop app
    - [x] Web browser
    - [ ] WhatsApp
- OS:
    - [ ] Windows
    - [x] macOS
    - [ ] Linux
    - [ ] Android
    - [ ] iOS

### If self-hosted
- Server Version [e.g. 1.0.1]: 1.5.0 (also tried 1.4.0 and 1.1.0)

## Additional context

stack trace:
khoj-server-1    | [20:59:13] INFO     📬 Initializing content index...            configure.py:236
khoj-server-1    |            DEBUG    🚨 No Content configuration available.        indexer.py:338
khoj-server-1    |            INFO     🌖 Khoj is ready to use                          main.py:157
khoj-server-1    |            INFO     Started server process [1]                      server.py:75
khoj-server-1    |            INFO     Waiting for application startup.                    on.py:45
khoj-server-1    |            INFO     Application startup complete.                       on.py:59
khoj-server-1    |            INFO     Uvicorn running on http://0.0.0.0:42110 (Press server.py:206
khoj-server-1    |                     CTRL+C to quit)
khoj-server-1    | [21:00:27] INFO     192.168.65.1:49632 - "GET                    h11_impl.py:431
khoj-server-1    |                     /config/content-source/notion HTTP/1.1" 200
khoj-server-1    |            INFO     192.168.65.1:49632 - "GET                    h11_impl.py:431
khoj-server-1    |                     /static/assets/pico.min.css HTTP/1.1" 200
khoj-server-1    |            INFO     192.168.65.1:49633 - "GET                    h11_impl.py:431
[...]

PR #648, opened 2024-02-19
Fix process for generating embeddings for Notion entries

sabaimran:
We weren't previously generating the embeddings correctly for the Notion entries. Update to resolve the embeddings logic.

Resolves https://github.com/khoj-ai/khoj/issues/637

### DIFF
--- a/src/khoj/processor/content/notion/notion_to_entries.py
+++ b/src/khoj/processor/content/notion/notion_to_entries.py
@@ -112,7 +112,11 @@ class NotionToEntries(TextToEntries):
                             page_entries = self.process_page(p_or_d)
                             current_entries.extend(page_entries)
 
-        return self.update_entries_with_ids(current_entries, user)
+        current_entries = TextToEntries.split_entries_by_max_tokens(current_entries, max_tokens=256)
+
+        current_entries = self.update_entries_with_ids(current_entries, user=user)
+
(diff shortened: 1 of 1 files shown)


src/khoj/processor/content/notion/notion_to_entries.py
```diff
@@ -112,7 +112,9 @@ def process(
                             page_entries = self.process_page(p_or_d)
                             current_entries.extend(page_entries)
 
-        return self.update_entries_with_ids(current_entries, user)
+        current_entries = TextToEntries.split_entries_by_max_tokens(current_entries, max_tokens=256)
+
+        return self.update_entries_with_ids(current_entries, user=user)
```
debanjum, line 117: Oh, was the issue just that we weren't fixing the entries by max token? Why was that breaking indexing of Notion entries?

debanjum (approved):
Left one comment (confused about how this fixes the issue) but otherwise looking forward to having the notion integration fixed!

src/khoj/processor/content/notion/notion_to_entries.py
sabaimran: Ah, that method also generally handles logic for splitting up the data into chunks and creating the `Entry` items that make it easy to write to the db.

PR #656, opened 2024-02-23
Retrieve, create, and save conversations differently for ClientApplications

sabaimran:
# Incoming
- Change the way conversations are being managed for requests coming from client applications
- Not all of our client apps will necessarily maintain state over the conversation IDs available to a user. For some (single-threaded conversations), it should just use a single conversation. Fix the code to do so.

### DIFF
--- a/src/khoj/database/adapters/__init__.py
+++ b/src/khoj/database/adapters/__init__.py
@@ -396,8 +396,12 @@ class ConversationAdapters:
     def get_conversation_by_user(
         user: KhojUser, client_application: ClientApplication = None, conversation_id: int = None
     ):
+        if client_application:
+            return Conversation.objects.filter(user=user, client=client_application).first()
+
         if conversation_id:
             conversation = Conversation.objects.filter(user=user, client=client_application, id=conversation_id)
+
         if not conversation_id or not conversation.exists():
             conversation = Conversation.objects.filter(user=user, client=client_application).order_by("-updated_at")
         if conversation.exists():
@@ -433,6 +437,9 @@ class ConversationAdapters:
     async def aget_conversation_by_user(
         user: KhojUser, client_application: ClientApplication = None, conversation_id: int = None, slug: str = None
     ):
+        if client_application:
+            return await Conversation.objects.filter(user=user, client=client_application).afirst()
+
         if conversation_id:
             conversation = Conversation.objects.filter(user=user, client=client_application, id=conversation_id)
         else:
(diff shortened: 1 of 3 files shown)


src/khoj/database/adapters/__init__.py
```diff
@@ -520,6 +527,7 @@ def save_conversation(
             conversation = Conversation.objects.filter(user=user, client=client_application, id=conversation_id)
         else:
             conversation = Conversation.objects.filter(user=user, client=client_application)
+
         if conversation.exists():
             conversation.update(conversation_log=conversation_log, slug=slug, updated_at=datetime.now(tz=timezone.utc))
```
debanjum, line 532: What happens if the chat API is called with no `conversation_id` or `client_application` set?

debanjum (approved):
Looks good!

src/khoj/database/adapters/__init__.py
```diff
@@ -435,10 +440,14 @@ async def aget_conversation_by_user(
...
         if conversation_id:
             conversation = Conversation.objects.filter(user=user, client=client_application, id=conversation_id)
-        else:
+        elif slug:
             conversation = Conversation.objects.filter(user=user, client=client_application, slug=slug)
+        else:
+            conversation = Conversation.objects.filter(user=user, client=client_application)
+
         if await conversation.aexists():
             return await conversation.afirst()
```
debanjum, line 449: Should we ensure that we return the most recently updated conversation here as well (like we're doing in `get_converation_by_user` method above)?
```suggestion
        if await conversation.aexists():
            return await conversation.order_by("-updated_at").afirst()
```

src/khoj/database/adapters/__init__.py
debanjum: Shouldn't we updated the most recently updated conversation from the retrieved conversation list 
```suggestion
        else:
            conversation = Conversation.objects.filter(user=user, client=client_application)

        if conversation.exists():
            conversation.order_by("-updated-at").first().update(conversation_log=conversation_log, slug=slug, updated_at=datetime.now(tz=timezone.utc))
```

src/khoj/database/adapters/__init__.py
sabaimran: Updated to add sorting in the filtering step

src/khoj/database/adapters/__init__.py
sabaimran: Great point! Shouldn't generally happen, but let's get the first conversation last updated.

PR #657, opened 2024-02-24
Improve Syncing Obsidian Vault

debanjum:
### Improve
- Only send files modified since their last sync for indexing on server from the Obsidian client

### Fix
- Invalidate static asset browser cache in Web client when Khoj version changes

### DIFF
--- a/src/interface/obsidian/src/utils.ts
+++ b/src/interface/obsidian/src/utils.ts
@@ -28,17 +28,42 @@ function fileExtensionToMimeType (extension: string): string {
     }
 }
 
-export async function updateContentIndex(vault: Vault, setting: KhojSetting, lastSyncedFiles: TFile[], regenerate: boolean = false): Promise<TFile[]> {
+function filenameToMimeType (filename: TFile): string {
+    switch (filename.extension) {
+        case 'pdf':
+            return 'application/pdf';
+        case 'png':
+            return 'image/png';
+        case 'jpg':
+        case 'jpeg':
+            return 'image/jpeg';
+        case 'md':
+        case 'markdown':
+            return 'text/markdown';
+        case 'org':
+            return 'text/org';
+        default:
+            return 'text/plain';
+    }
+}
+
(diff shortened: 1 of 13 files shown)


src/interface/obsidian/src/utils.ts
```diff
@@ -28,17 +28,42 @@ function fileExtensionToMimeType (extension: string): string {
...
+        case 'jpg':
+        case 'jpeg':
+            return 'image/jpeg';
+        case 'md':
+        case 'markdown':
+            return 'text/markdown';
+        case 'org':
+            return 'text/org';
+        default:
+            return 'text/plain';
```
sabaimran, line 46: hmm, is this a fair assumption? there would be a bunch of file types we can't process, and we wouldn't necessarily want them to slip through if they're interpreted as `text/plain`. like docx, xml, odt, ppt, etc.

src/khoj/interface/web/search.html
```diff
@@ -4,13 +4,13 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0 maximum-scale=1.0">
         <title>Khoj - Search</title>
 
-        <link rel="icon" type="image/png" sizes="128x128" href="/static/assets/icons/favicon-128x128.png">
-        <link rel="manifest" href="/static/khoj.webmanifest">
-        <link rel="stylesheet" href="/static/assets/khoj.css">
+        <link rel="icon" type="image/png" sizes="128x128" href="/static/assets/icons/favicon-128x128.png?v={{ khoj_version }}">
```
sabaimran, line 7: Is the rationale behind including the `khoj_version` to limit the caching to the current version for static files? Neat solution.

sabaimran (approved):
Main question just about how to handle unknown file types, but all else looks good.

src/khoj/interface/web/search.html
debanjum: Yes, exactly!

src/interface/obsidian/src/utils.ts
debanjum: Yeah, this is the same function as the one we use in desktop `main.js`. I've updated the code to `console.warn` when we fallback to the default option for now

PR #660, opened 2024-02-26
Include additional user context in the image generation flow

sabaimran:
# Incoming
- Add online context and user notes in the image generation flow
- Allow the agent to dynamically determine whether to return the response as an image or as text
- Render references when included as part of image generation in web, desktop chat views

### DIFF
--- a/src/khoj/processor/conversation/prompts.py
+++ b/src/khoj/processor/conversation/prompts.py
@@ -299,7 +299,7 @@ Collate the relevant information from the website to answer the target query.
 """.strip()
 )
 
-pick_relevant_mode = PromptTemplate.from_template(
+pick_relevant_output_mode = PromptTemplate.from_template(
     """
 You are Khoj, an excellent analyst for selecting the correct way to respond to a user's query. You have access to a limited set of modes for your response. You can only use one of these modes.
 
--- a/src/khoj/processor/conversation/prompts.py
+++ b/src/khoj/processor/conversation/prompts.py
@@ -125,9 +125,11 @@ You are a talented creator. Generate a detailed prompt to generate an image base
 Today's Date: {current_date}
 User's Location: {location}
 
-User's Notes: {references}
+User's Notes:
+{references}
 
-Online References: {online_results}
+Online References:
+{online_results}
 
 Conversation Log:
 {chat_history}
--- a/src/khoj/processor/conversation/prompts.py
+++ b/src/khoj/processor/conversation/prompts.py
@@ -120,16 +120,21 @@ User's Notes:
 
 image_generation_improve_prompt = PromptTemplate.from_template(
     """
-You are a talented creator. Generate a detailed prompt to generate an image based on the following description. Update the query below to improve the image generation. Add additional context to the query to improve the image generation. Make sure to retain any important information from the query. Use the conversation log to inform your response.
+You are a talented creator. Generate a detailed prompt to generate an image based on the following description. Update the query below to improve the image generation. Add additional context to the query to improve the image generation. Make sure to retain any important information originally from the query. You are provided with the following information to help you generate the prompt:
 
 Today's Date: {current_date}
 User's Location: {location}
+
+User's Notes: {references}
+
+Online References: {online_results}
 
 Conversation Log:
 {chat_history}
 
 Query: {query}
 
+Remember, now you are generating a prompt to improve the image generation. Add additional context to the query to improve the image generation. Make sure to retain any important information originally from the query. Use the additional context from the user's notes, online references and conversation log to improve the image generation.
 Improved Query:"""
 )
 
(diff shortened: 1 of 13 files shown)


src/khoj/processor/conversation/prompts.py
```diff
@@ -120,16 +120,21 @@
...
-You are a talented creator. Generate a detailed prompt to generate an image based on the following description. Update the query below to improve the image generation. Add additional context to the query to improve the image generation. Make sure to retain any important information from the query. Use the conversation log to inform your response.
+You are a talented creator. Generate a detailed prompt to generate an image based on the following description. Update the query below to improve the image generation. Add additional context to the query to improve the image generation. Make sure to retain any important information originally from the query. You are provided with the following information to help you generate the prompt:
 
 Today's Date: {current_date}
 User's Location: {location}
 
+User's Notes: {references}
+
+Online References: {online_results}
+
```
debanjum, line 131: nit: Should references be on a new line as they're multi-line strings like the conversation log

```suggestion
User's Notes:
{references}

Online References:
{online_results}

```

src/khoj/utils/helpers.py
```diff
@@ -289,6 +289,11 @@ class ConversationCommand(str, Enum):
     ConversationCommand.Online: "Use this when you would like to look up information on the internet",
 }
 
+mode_descriptions_for_llm = {
+    ConversationCommand.Image: "Use this if you think the user is requesting an image or visual representation of the query.",
```
debanjum, line 293:
```suggestion
    ConversationCommand.Image: "Use this if you think the user is requesting an image or visual response to their query.",
```

src/khoj/utils/helpers.py
```diff
@@ -289,6 +289,11 @@ class ConversationCommand(str, Enum):
...
+    ConversationCommand.Default: "Use this if the other modes don't seem to fit the query.",
```
debanjum, line 294:
```suggestion
    ConversationCommand.Default: "Use this if the other response modes don't seem to fit the query.",
```

tests/test_openai_chat_director.py
```diff
@@ -509,3 +509,31 @@ async def test_get_correct_tools_with_chat_history(chat_client):
...
+
+# ----------------------------------------------------------------------------------------------------
+@pytest.mark.anyio
+@pytest.mark.django_db(transaction=True)
+async def test_use_default_response_mode(chat_client):
+    # Arrange
+    user_query = "What's the latest in the Israel/Palestine conflict?"
+
+    # Act
+    mode = await aget_relevant_modes(user_query, {})
```
debanjum, line 522: These current tests should be moved into the `test_openai_chat_actor.py` tests

Chat director tests should test the response when calling the chat API. See `test_answer_using_file_filter` as example. We can check if the API response contains an image or text for the auto output tool selection tests you've added here.

tests/test_gpt4all_chat_director.py
```diff
@@ -533,3 +533,31 @@ async def test_get_correct_tools_with_chat_history(client_offline_chat):
     # Assert
     tools = [tool.value for tool in tools]
     assert tools == ["online"]
+
+
+# ----------------------------------------------------------------------------------------------------
+@pytest.mark.anyio
+@pytest.mark.django_db(transaction=True)
+async def test_use_default_response_mode(chat_client):
```
debanjum, line 541: These current tests should be moved into the `test_gpt4all_chat_actor.py` tests

Chat director tests should test the response when calling the chat API. See `test_answer_using_file_filter` as example. We can check if the API response contains an image or text for the auto output tool selection tests you've added here.

src/khoj/routers/helpers.py
```diff
@@ -185,6 +189,42 @@ async def aget_relevant_information_sources(query: str, conversation_history: di
         return [ConversationCommand.Default]
 
 
+async def aget_relevant_modes(query: str, conversation_history: dict):
```
debanjum, line 192: nit: We should make the function name more specific

```suggestion
async def aget_relevant_output_mode(query: str, conversation_history: dict):
```

src/khoj/processor/conversation/prompts.py
```diff
@@ -294,6 +299,40 @@
 """.strip()
 )
 
+pick_relevant_mode = PromptTemplate.from_template(
```
debanjum, line 302: nit: We should make the name more specific to its purpose

```suggestion
pick_relevant_output_mode = PromptTemplate.from_template(
```

src/khoj/routers/api_chat.py
```diff
@@ -250,6 +251,9 @@ async def chat(
 
     if conversation_commands == [ConversationCommand.Default]:
         conversation_commands = await aget_relevant_information_sources(q, meta_log)
+        mode = await aget_relevant_modes(q, meta_log)
```
debanjum, line 254: We should (at some point?) parallelize calls to `aget_relevant_output_modes` and `aget_relevant_information_sources` to reduce response latency

src/khoj/routers/helpers.py
```diff
@@ -185,6 +189,42 @@ async def aget_relevant_information_sources(query: str, conversation_history: di
...
+    """
+    Given a query, determine which of the available tools the agent should use in order to answer appropriately.
+    """
+
+    mode_options = dict()
+
+    for mode, description in mode_descriptions_for_llm.items():
```
debanjum, line 199: Can we just use `mode_descriptions_for_llms` directly instead of creating the new `mode_options` dictionary? Given `mode_descriptions_for_llms` is already a dictionary

debanjum (approved):
Left a few comments. Psyched to have the output mode be automatically selected by Khoj. It'll make the interaction more fluid than using `/slash` commands manually!

src/khoj/routers/api_chat.py
sabaimran: Yes, I'm tracking a ticket for this separately.

src/khoj/routers/helpers.py
sabaimran: Not quite. This formats it differently so that it uses the value of the `ConversationCommand` enum is used in the input prompt.

PR #666, opened 2024-03-05
Support upload for files via drag/drop in the web UI

sabaimran:
# Incoming
- Allow users to drag/drop files into the main chat web UI to upload
- Allow users to click on the attachment icon and select a file for upload

Export-1709721982870

### DIFF
--- a/src/khoj/interface/web/chat.html
+++ b/src/khoj/interface/web/chat.html
@@ -700,7 +700,6 @@ To get started, just start typing below. You can also type / to see a list of co
 
             dropzone.ondrop = function(event) {
                 event.preventDefault();
-                this.classList.remove('dragover');
 
                 var file = event.dataTransfer.files[0];
                 uploadDataForIndexing(file);
--- a/src/khoj/interface/web/chat.html
+++ b/src/khoj/interface/web/chat.html
@@ -567,6 +567,10 @@ To get started, just start typing below. You can also type / to see a list of co
         function uploadDataForIndexing(file) {
             if (!allowedExtensions.includes(file.type)) {
                 alert("Sorry, that file type is not yet supported");
+                var overlayText = document.getElementById("dropzone-overlay");
+                if (overlayText != null) {
+                    overlayText.remove();
+                }
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/khoj/interface/web/chat.html
```diff
@@ -563,6 +564,148 @@
...
+                // Display loading spinner
+                var loadingSpinner = document.createElement("div");
+                overlayText.innerHTML = "Uploading file for indexing";
+                loadingSpinner.className = "spinner";
+                overlayText.appendChild(loadingSpinner);
+            }
+
+            reader.onload = function (event) {
+                fileContents = event.target.result;
+                console.log(fileContents);
```
debanjum, line 638: nit: can we remove/dedupe all these `console.log` statements? I'm guessing it was only used to debug file uploads for indexing

src/khoj/interface/web/chat.html
```diff
@@ -1372,7 +1557,7 @@
         }
         #input-row {
             display: grid;
-            grid-template-columns: auto 32px 40px;
+            grid-template-columns: auto 32px 40px 32px;
```
debanjum, line 1560: Should we move the file upload button to the left of the chat input text field? Idea being that buttons on the right side are the more used actions like the current send message or voice note buttons (as closer to thumb on phone screen, so easier to click), the buttons on the left can be the less used actions

```suggestion
            grid-template-columns: 32px auto 32px 40px;
```

src/khoj/interface/web/chat.html
```diff
@@ -563,6 +564,148 @@
...
 
+        function openFileBrowser() {
+            event.preventDefault();
+            var overlayText = document.getElementById("dropzone-overlay");
+            var dropzone = document.getElementById('chat-body');
+
+            if (overlayText == null) {
+                dropzone.classList.add('dragover');
+                var overlayText = document.createElement("div");
+                overlayText.innerHTML = "Select a file to share it with Khoj";
```
debanjum, line 575: Can we add a hint that you can also just drag and drop files onto the chat interface to upload them for indexing?

src/khoj/interface/web/chat.html
```diff
@@ -563,6 +564,148 @@
...
+                overlayText.className = "dropzone-overlay";
+                overlayText.id = "dropzone-overlay";
+                dropzone.appendChild(overlayText);
+            }
+
+            const fileInput = document.createElement('input');
+            fileInput.type = 'file';
+            fileInput.addEventListener('change', function() {
+                const selectedFile = fileInput.files[0];
```
debanjum, line 584: Does this file attachment button only support uploading 1 file at a time?

debanjum (approved):
Looks great! This will make it super easy for folks wanting to upload static files to chat with, without needing to download any client app (e.g  desktop, obsidian or emacs app)

src/khoj/interface/web/chat.html
sabaimran: Yes, that's the default button of the component. It can be updated to take multiple as well. We can pick that up later if there's demand.

src/khoj/interface/web/chat.html
sabaimran: I suspect this text won't really be visible to most people once the file picker pops up, but sure, doesn't hurt.

PR #668, opened 2024-03-09
Rerank Search Results by Default on GPU machines

debanjum:
**Trigger**: SentenceTransformer Cross Encoder models now run fast on GPU enabled machines, including Mac ARM devices since UKPLab/sentence-transformers#2463

### Details
- Only call search API when pause in typing search query on web, desktop apps (44c8d093)
- Use cross-encoder to rerank search results by default on GPU machines (1105d881) and when using an inference server (53d40248)

### DIFF
--- a/src/khoj/interface/web/search.html
+++ b/src/khoj/interface/web/search.html
@@ -193,16 +193,17 @@
                 });
         }
 
+        let debounceTimeout;
         function incrementalSearch(event) {
-            type = document.getElementById("type").value;
-            // Search with reranking on 'Enter'
-            if (event.key === 'Enter') {
-                search(rerank=true);
-            }
-            // Limit incremental search to text types
-            else if (type !== "image") {
-                search(rerank=false);
-            }
+            // Run incremental search only after waitTime passed since the last key press
+            let waitTime = 300;
+            clearTimeout(debounceTimeout);
+            debounceTimeout = setTimeout(() => {
+                type = document.getElementById("type").value;
+                // Search with reranking on 'Enter'
+                let should_rerank = event.key === 'Enter';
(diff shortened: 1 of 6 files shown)


src/khoj/interface/web/search.html
```diff
@@ -193,16 +193,22 @@
...
+            debounceTimeout = setTimeout(() => {
+                type = document.getElementById("type").value;
+                // Search with reranking on 'Enter'
+                if (event.key === 'Enter') {
+                    search(rerank=true);
+                }
+                // Limit incremental search to text types
+                else if (type !== "image") {
+                    search(rerank=false);
+                }
```
sabaimran, line 210: This portion isn't necessary since we don't have images at the moment.

src/khoj/routers/api.py
```diff
@@ -146,9 +146,12 @@ async def search(
                     # Collate results
                     results += text_search.collate_results(hits, dedupe=dedupe)
 
+                    # Rerank results if explicitly requested or if device has GPU
+                    rerank = r or state.device.type != "cpu" or is_env_var_true("KHOJ_RERANK")
+
                     # Sort results across all content types and take top results
                     results = text_search.rerank_and_sort_results(
-                        results, query=defiltered_query, rank_results=r, search_model_name=search_model.name
+                        results, query=defiltered_query, rank_results=rerank, search_model_name=search_model.name
```
sabaimran, line 154: You can see in `embeddings.py` how we determine whether to make an API query for the result or to use on-device compute. We should take that into account. Perhaps always run cross encoder if the API is being used, as one would assume the service is production-ready?

debanjum: Fair point, fixed that!

PR #669, opened 2024-03-09
Do not create new chat session when an old chat session is deleted

debanjum:
### Issue
Previously deleting a chat session from the side panel on desktop, web app would sometimes result in also creating a new chat session

### Fix
  `get_conversation_by_user` shouldn't return new conversation if
  conversation with requested id not found.

  It should only return new conversation if no specific conversation
  is requested and no conversations found for user at all

### Miscellaneous Improvements
  - Chat history load should be logged as call to that chat_history api,
    not the "chat" api
  - Show status updates of clearing conversation history in chat input
  - Simplify web, desktop client code by removing unnecessary new variables

### Repro
  - Delete a new chat, this calls loadChat via window.onload which
    calls server /chat/history API endpoint with conversationId set to
    that of just deleted conversation sporadically

    The call to GET chat/history API with conversationId set occurs
    when window.onload triggers before the conversationId is deleted
    by the delete button after the DELETE /chat/history API call (via race)

  - In such a scenario, get_conversation_by_user called by
    chat/history API with conversationId of deleted conversation
    returns a new conversation

### DIFF
--- a/src/interface/desktop/chat.html
+++ b/src/interface/desktop/chat.html
@@ -626,10 +626,9 @@
 
             let chatBody = document.getElementById("chat-body");
             chatBody.innerHTML = "";
-            let conversationId = chatBody.dataset.conversationId;
             let chatHistoryUrl = `/api/chat/history?client=desktop`;
-            if (conversationId) {
-                chatHistoryUrl += `&conversation_id=${conversationId}`;
+            if (chatBody.dataset.conversationId) {
+                chatHistoryUrl += `&conversation_id=${chatBody.dataset.conversationId}`;
             }
 
             fetch(`${hostURL}${chatHistoryUrl}`, { headers })
@@ -650,6 +649,8 @@
                         // Disable chat input field and update placeholder text
                         document.getElementById("chat-input").setAttribute("disabled", "disabled");
                         document.getElementById("chat-input").setAttribute("placeholder", "Configure Khoj to enable chat");
+                    } else if (data.status != "ok") {
+                        throw new Error(data.message);
                     } else {
                         // Set welcome message on load
                         renderMessage("Hey 👋🏾, what's up?", "khoj");
@@ -657,12 +658,9 @@
                     return data.response;
                 })
                 .then(response => {
-                    conversationId = response.conversation_id;
-                    const conversationTitle = response.slug || `New conversation 🌱`;
-
                     let chatBody = document.getElementById("chat-body");
-                    chatBody.dataset.conversationId = conversationId;
-                    chatBody.dataset.conversationTitle = conversationTitle;
+                    chatBody.dataset.conversationId = response.conversation_id;
+                    chatBody.dataset.conversationTitle = response.slug || `New conversation 🌱`;
 
                     const fullChatLog = response.chat || [];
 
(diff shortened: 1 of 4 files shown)


src/interface/desktop/chat.html
```diff
@@ -626,10 +626,9 @@
 
             let chatBody = document.getElementById("chat-body");
             chatBody.innerHTML = "";
-            let conversationId = chatBody.dataset.conversationId;
             let chatHistoryUrl = `/api/chat/history?client=desktop`;
-            if (conversationId) {
-                chatHistoryUrl += `&conversation_id=${conversationId}`;
+            if (chatBody.dataset.conversationId) {
+                chatHistoryUrl += `&conversation_id=${chatBody.dataset.conversationId}`;
```
sabaimran, line 631: 🤔 why not use the computed variable?

src/khoj/database/adapters/__init__.py
```diff
@@ -405,9 +405,9 @@ def get_conversation_by_user(
         else:
             conversation = (
                 Conversation.objects.filter(user=user, client=client_application).order_by("-updated_at").first()
-            )
+            ) or Conversation.objects.create(user=user, client=client_application)
 
-        return conversation or Conversation.objects.create(user=user, client=client_application)
+        return conversation
```
sabaimran, line 410: The conversation returned would be `None` then if the `conversation_id` did not match any. Is that what we want to do?

src/khoj/routers/api_chat.py
```diff
@@ -75,16 +75,21 @@ def chat_history(
...
+    meta_log = conversation.conversation_log
     meta_log.update(
         {"conversation_id": conversation.id, "slug": conversation.title if conversation.title else conversation.slug}
     )
 
     update_telemetry_state(
         request=request,
         telemetry_type="api",
-        api="chat",
+        api="chat_history",
```
sabaimran, line 92: 👌🏾

sabaimran (approved):
Awesome to have a fix for this. Just one minor question about how conversationId is being interpreted in the html pages, but looks good.

src/khoj/database/adapters/__init__.py
debanjum: Yes, this is the core fix being used for the bug here. Also it's a bit strange if I want a get conversation with specific id but can instead get a conversation with a different id when a conversation with requested id doesn't exist

src/interface/desktop/chat.html
debanjum: Not any important reason, just didn't seem necessary to introduce a new variable given the value is just being used in 2 places

sabaimran:
I took a second look at the code. The method `aget_conversation_by_user`, which is in the hot path for the `/chat/` endpoint hasn't been updated. For that one, it'll still create a new conversation if/when the `conversation_id` doesn't come back with a match.

debanjum:
> I took a second look at the code. The method aget_conversation_by_user, which is in the hot path for the /chat/ endpoint hasn't been updated. For that one, it'll still create a new conversation if/when the conversation_id doesn't come back with a match.

Great catch! Fixed in https://github.com/khoj-ai/khoj/pull/677/files

PR #670, opened 2024-03-09
Improve Online Search Speed and Context

debanjum:
### Major
- 38238a50 Read web pages in parallel to improve chat response time
- 094f3c93 Read web pages directly when Olostep proxy not setup
- bd35ff38 Include search results & web page content in online context for chat response

### Minor
- 14c1d688 Simplify, modularize and add type hints to online search functions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ classifiers = [
     "Topic :: Text Processing :: Linguistic",
 ]
 dependencies = [
-    "bs4 >= 0.0.1",
+    "beautifulsoup4 ~= 4.12.3",
     "dateparser >= 1.1.1",
     "defusedxml == 0.7.1",
     "fastapi >= 0.104.1",
@@ -58,7 +58,6 @@ dependencies = [
     "langchain <= 0.2.0",
     "langchain-openai >= 0.0.5",
     "requests >= 2.26.0",
-    "bs4 >= 0.0.1",
     "anyio == 3.7.1",
     "pymupdf >= 1.23.5",
     "django == 4.2.10",
@@ -76,6 +75,7 @@ dependencies = [
     "openai-whisper >= 20231117",
     "django-phonenumber-field == 7.3.0",
     "phonenumbers == 8.13.27",
+    "markdownify ~= 0.11.6",
 ]
 dynamic = ["version"]
 
--- a/src/khoj/processor/tools/online_search.py
+++ b/src/khoj/processor/tools/online_search.py
@@ -6,6 +6,8 @@ from typing import Dict, Union
 
 import aiohttp
 import requests
+from bs4 import BeautifulSoup
+from markdownify import markdownify
 
 from khoj.routers.helpers import extract_relevant_info, generate_online_subqueries
 from khoj.utils.helpers import is_none_or_empty, timer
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -76,6 +76,9 @@ dependencies = [
     "openai-whisper >= 20231117",
     "django-phonenumber-field == 7.3.0",
     "phonenumbers == 8.13.27",
+    "aiohttp ~= 3.9.0",
+    "beautifulsoup4 ~= 4.12.3",
```
sabaimran, line 80: Can you also remove the `bs4` dependencies, and make sure the HTML parsing for plaintext indexing still works after removing it and using beautifulsoup4 directly? According to the [pypi project for bs4](https://pypi.org/project/bs4/),

> This is a dummy package designed to prevent namesquatting on PyPI. You should install [beautifulsoup4](https://pypi.python.org/pypi/beautifulsoup4) instead.

note: there are two bs4 dependencies for some reason.

src/khoj/processor/tools/online_search.py
```diff
@@ -34,83 +36,94 @@
...
-                try:
-                    extracted_content[subquery].append(search_with_olostep(result["link"]).strip())
-                except Exception as e:
-                    logger.error(f"Error while searching web page of '{result['link']}': {e}", exc_info=True)
-                    continue
-            extracted_relevant_content = await extract_relevant_info(subquery, extracted_content)
-            response_dict[subquery]["extracted_content"] = extracted_relevant_content
+        response_dict[subquery] = search_with_google(subquery)
+
+    # Gather distinct web pages from organic search results of each subquery without an instant answer
```
sabaimran, line 52: Specify that we're only grabbing the first link from each subquery's organic search results?

Or make the `1` used below set via variable, like `MAX_NUM_PAGES_TO_SCRAPE`.

src/khoj/processor/tools/online_search.py
```diff
@@ -34,83 +36,94 @@
...
+        if "answerBox" not in response_dict[subquery]
+    }
+
+    # Read, extract relevant info from the retrieved web pages
+    tasks = []
+    for webpage_link in webpage_links:
+        logger.info(f"Reading web page at '{webpage_link}'")
+        task = read_webpage_and_extract_content(subquery, webpage_link)
+        tasks.append(task)
+    results = await asyncio.gather(*tasks)
```
sabaimran, line 66: Nice! 🚀

tests/test_helpers.py
```diff
@@ -90,7 +90,7 @@ def test_olostep_api():
     website = "https://en.wikipedia.org/wiki/Great_Chicago_Fire"
 
     # Act
-    response = search_with_olostep(website)
+    response = read_webpage_with_olostep(website)
```
sabaimran, line 93: Let's also add a test for `read_webpage`?

src/khoj/processor/tools/online_search.py
```diff
@@ -1,23 +1,25 @@
+import asyncio
 import json
 import logging
 import os
-from typing import Dict, List, Union
+from typing import Dict, Tuple, Union
 
+import aiohttp
 import requests
+from bs4 import BeautifulSoup
```
sabaimran, line 9: Use beautifulsoup, rather than bs4

src/khoj/processor/tools/online_search.py
```diff
@@ -1,23 +1,25 @@
+import asyncio
```
sabaimran, line 1: Neat, the code's also looking a lot crisper 👌🏾

sabaimran (commented):
Just one comment around dependency resolution and one for unit tests, but looks good!

pyproject.toml
debanjum: Good catch, there were also 2 aiohttp dependencies. Cleaned both up in `pyproject.toml`

src/khoj/processor/tools/online_search.py
debanjum: Used variable (`MAX_WEBPAGES_TO_READ`) to store webpages to read per search query

src/khoj/processor/tools/online_search.py
debanjum: Even though the pip package to install is `beautifulsoup4`, the package name for import is `bs4`. So this will stay the same

tests/test_helpers.py
debanjum: Good point! Added test for `read_webpage`

PR #671, opened 2024-03-11
Part 1: Server-side changes to support agents integrated with Conversations

sabaimran:
# Incoming
- This initial set of changes allows us to have support for agents that are linked to conversations.
- Add a db model for Agents, attaching them to conversations
- When an agent is added to a conversation, override the system prompt to tweak the instructions
- Agents can be configured with prompt modification, model specification, a profile picture, and other things
- Admin-configured models will not be editable by individual users
- Add unit tests to verify agent behavior. Unit tests demonstrate imperfect adherence to prompt specifications
- Agent configuration is currently limited to the system prompt


src/khoj/database/adapters/__init__.py
```diff
@@ -391,6 +393,54 @@ async def aget_client_application_by_id(client_id: str, client_secret: str):
...
+            # If the agent is set to the default agent, then return None and let the default application code be used
+            return None
+        return agent
+
+    @staticmethod
+    def get_default_agent():
+        return Agent.objects.filter(name=AgentAdapters.DEFAULT_AGENT_NAME).first()
+
+    @staticmethod
+    def create_default_agent():
```
sabaimran, line 421: Changed this behavior in #675 to only sync the agent to the existing default prompt, rather than delete the existing agent.

src/khoj/database/models/__init__.py
```diff
@@ -69,6 +72,35 @@ class Type(models.TextChoices):
...
+        OFFLINE = "offline"
+
+    max_prompt_size = models.IntegerField(default=None, null=True, blank=True)
+    tokenizer = models.CharField(max_length=200, default=None, null=True, blank=True)
+    chat_model = models.CharField(max_length=200, default="mistral-7b-instruct-v0.1.Q4_0.gguf")
+    model_type = models.CharField(max_length=200, choices=ModelType.choices, default=ModelType.OFFLINE)
+
+
+class Agent(BaseModel):
+    creator = models.ForeignKey(KhojUser, on_delete=models.CASCADE, default=None, null=True, blank=True)
```
debanjum, line 87: Should we add a comment saying creator is to be unset only for admin managed agents? Just as a reminder of why creator is allowed to remain unset

src/khoj/database/models/__init__.py
```diff
@@ -69,6 +72,35 @@ class Type(models.TextChoices):
...
+    tools = models.JSONField(default=list)  # List of tools the agent has access to, like online search or notes search
+    public = models.BooleanField(default=False)
+    managed_by_admin = models.BooleanField(default=False)
+    chat_model = models.ForeignKey(ChatModelOptions, on_delete=models.CASCADE)
+
+
+@receiver(pre_save, sender=Agent)
+def check_public_name(sender, instance, **kwargs):
+    if instance.public:
+        if Agent.objects.filter(name=instance.name, public=True).exists():
```
debanjum, line 100: Should we just enforce unique Agent names per creator. Agent public identifiers (like url, name) can be  [creator_name]/[agent_name] similar to how github, huggingface etc do naming conventions for repositories, models. For e.g an agent called `smith` created by username = `neo` will be called `neo/smith` etc.

```suggestion
        if Agent.objects.filter(name=instance.name, creator=instance.creator).exists():
```

src/khoj/database/adapters/__init__.py
```diff
@@ -391,6 +393,54 @@ async def aget_client_application_by_id(client_id: str, client_secret: str):
...
+class AgentAdapters:
+    DEFAULT_AGENT_NAME = "khoj"
+    DEFAULT_AGENT_AVATAR = "https://khoj-web-bucket.s3.amazonaws.com/lamp-128.png"
+
+    @staticmethod
+    async def aget_agent_by_id(agent_id: int):
+        return await Agent.objects.filter(id=agent_id).afirst()
+
+    @staticmethod
+    def get_all_acessible_agents(user: KhojUser = None):
```
debanjum, line 405: nit: spelling of accessible

```suggestion
    def get_all_accessible_agents(user: KhojUser = None):
```

src/khoj/processor/conversation/offline/chat_model.py
```diff
@@ -156,6 +158,16 @@ def converse_offline(
     # Initialize Variables
     compiled_references_message = "\n\n".join({f"{item}" for item in references})
 
+    system_prompt = ""
```
debanjum, line 161: nit: this initialization isn't required as system_prompt is always set in the if/else below

src/khoj/processor/conversation/openai/gpt.py
```diff
@@ -125,6 +127,13 @@ def converse(
 
     conversation_primer = prompts.query_prompt.format(query=user_query)
 
+    system_prompt = ""
```
debanjum, line 130: nit: This initialization isn't required as system_prompt is always set in the if/else conditional below

src/khoj/routers/api_chat.py
```diff
@@ -148,6 +148,7 @@ def chat_sessions(
 async def create_chat_session(
     request: Request,
     common: CommonQueryParams,
+    agent_id: Optional[int] = None,
```
debanjum, line 151: The agent_id query param seems unused in the API endpoint. Is this expected?

src/khoj/routers/api_chat.py
```diff
@@ -341,6 +342,7 @@ async def chat(
     llm_response, chat_metadata = await agenerate_chat_response(
         defiltered_query,
         meta_log,
+        conversation,
```
debanjum, line 345: If we're passing `conversation` down to `agenerate_chat_response`, passing `meta_log` separately seems unnecessary, given `meta_log = conversation.meta_log`

src/khoj/database/adapters/__init__.py
```diff
@@ -391,6 +393,54 @@ async def aget_client_application_by_id(client_id: str, client_secret: str):
         return await ClientApplication.objects.filter(client_id=client_id, client_secret=client_secret).afirst()
 
 
+class AgentAdapters:
+    DEFAULT_AGENT_NAME = "khoj"
+    DEFAULT_AGENT_AVATAR = "https://khoj-web-bucket.s3.amazonaws.com/lamp-128.png"
+
+    @staticmethod
+    async def aget_agent_by_id(agent_id: int):
+        return await Agent.objects.filter(id=agent_id).afirst()
```
debanjum, line 402: Should we add check here to see if `user_id` is permitted to access the specified agent_id (i.e if `agent_id` is either public or created by user_id)? This will add a little more in-depth defense against unauthorized access to private agents even if you know that agent's id

src/khoj/configure.py
```diff
@@ -229,11 +230,16 @@ def configure_server(
...
         state.SearchType = configure_search_types()
         state.search_models = configure_search(state.search_models, state.config.search_type)
+        setup_default_agent()
         initialize_content(regenerate, search_type, init, user)
     except Exception as e:
         raise e
 
 
+def setup_default_agent():
+    AgentAdapters.create_default_agent()
```
debanjum, line 240: We should only `create_default_agent` if it doesn't already exist. Creating one everytime app starts seems unnecessary

debanjum (approved):
Looks great, really looking forward to having my own team of Khoj Agents on standby! 😄

src/khoj/database/models/__init__.py
sabaimran: Hmm, the tricky bit about that is that the `given_name`/`family_name` fields on the user objects are nullable, and email address wouldn't make for a good route. I get your point, but I think let's optimize this later on. It's not a concern until we expose the ability to make new agents.

src/khoj/routers/api_chat.py
sabaimran: Not necessary, but just that `meta_log` is already extracted for prior function arguments. Prevents unnecessary calls to get the same information again later on.

src/khoj/configure.py
sabaimran: This is so that it can easily keep in sync as the system prompt changes. In the method, it only creates an actual _new_ agent if it doesn't exist. Otherwise, it just updates the fields.

PR #675, opened 2024-03-14
Part 2: Add web UI updates for basic agent interactions

sabaimran:
# Incoming
- Show which agent you're talking to
- See the list of full available agents
- Add ability to inspect a specific agent

Note: agent creation is not included in this update.


src/khoj/database/adapters/__init__.py
```diff
@@ -394,20 +394,41 @@ async def aget_client_application_by_id(client_id: str, client_secret: str):
...
     async def aget_agent_by_id(agent_id: int, user: KhojUser):
         agent = await Agent.objects.filter(id=agent_id).afirst()
-        # Check if it's accessible to the user
+        if agent and (agent.public or agent.creator == user):
+            return agent
+        return None
+
+    @staticmethod
+    async def aget_agent_by_slug(agent_slug: str):
+        return await Agent.objects.filter(slug__iexact=agent_slug.lower()).afirst()
```
debanjum, line 410: Should this helper method also check if retrieved agents by slug are public or created by user?

src/khoj/database/adapters/__init__.py
```diff
@@ -423,12 +444,18 @@ def get_default_agent():
...
     @staticmethod
     def create_default_agent():
-        # First delete the existing default
-        Agent.objects.filter(name=AgentAdapters.DEFAULT_AGENT_NAME).delete()
-
         default_conversation_config = ConversationAdapters.get_default_conversation_config()
         default_personality = prompts.personality.format(current_date="placeholder")
 
+        if Agent.objects.filter(name=AgentAdapters.DEFAULT_AGENT_NAME).exists():
+            agent = Agent.objects.filter(name=AgentAdapters.DEFAULT_AGENT_NAME).first()
```
debanjum, line 451: We could use the `get_agent_by_slug` method from above here?

```suggestion
        if get_agent_by_slug(AgentAdapters.DEFAULT_AGENT_SLUG).exists():
            agent = get_agent_by_slug(AgentAdapters.DEFAULT_AGENT_SLUG).first()
```

src/khoj/database/models/__init__.py
```diff
@@ -84,23 +85,33 @@ class ModelType(models.TextChoices):
...
-    )  # Creator will only be null when the agents are managed by admin
+    creator = models.ForeignKey(KhojUser, on_delete=models.CASCADE, default=None, null=True, blank=True)
     name = models.CharField(max_length=200)
     personality = models.TextField()
     avatar = models.URLField(max_length=400, default=None, null=True, blank=True)
     tools = models.JSONField(default=list)  # List of tools the agent has access to, like online search or notes search
     public = models.BooleanField(default=False)
     managed_by_admin = models.BooleanField(default=False)
     chat_model = models.ForeignKey(ChatModelOptions, on_delete=models.CASCADE)
+    slug = models.CharField(max_length=200, default=None, null=True, blank=True)
```
debanjum, line 96: In what scenarios is the agent slug expected to be unset?

src/khoj/database/models/__init__.py
```diff
@@ -84,23 +85,33 @@ class ModelType(models.TextChoices):
...
+        if Agent.objects.filter(name=instance.name, creator=instance.creator).exists():
+            raise ValidationError(f"A private Agent with the name {instance.name} already exists.")
+
+        slug = instance.name.lower().replace(" ", "-")
+        observed_random_numbers = set()
+        while Agent.objects.filter(slug=slug).exists():
+            random_number = choice([i for i in range(0, 10000) if i not in observed_random_numbers])
+            observed_random_numbers.add(random_number)
+            slug = f"{slug}-{random_number}"
+        instance.slug = slug
```
debanjum, line 114: Can we simplify this to use `random.randint(10000)`?

```suggestion
        slug = instance.name.lower().replace(" ", "-")
        random_numbers = random.randint(10000)
        slug = f"{slug}-{random_number}"
        instance.slug = slug
```

src/khoj/interface/web/agent.html
```diff
@@ -0,0 +1,286 @@
...
+                display: block;
+                width: min(30vw, 100%);
+                margin-left: auto;
+                margin-right: auto;
+            }
+        }
+    </style>
+    <script>
+        async function openChat(agentId) {
+            let response = await fetch(`/api/chat/sessions?agent_slug=${agentId}`, { method: "POST" });
```
debanjum, line 260: nit: s/agentId/agentSlug/

```suggestion
        async function openChat(agentSlug) {
            let response = await fetch(`/api/chat/sessions?agent_slug=${agentSlug}`, { method: "POST" });
```

src/khoj/interface/web/agent.html
```diff
@@ -0,0 +1,286 @@
...
+                </div>
+                <div class="divider"></div>
+                <div id="agent-data-wrapper">
+                    <div id="agent-avatar-wrapper">
+                        <img id="agent-avatar" src="{{ agent.avatar }}" alt="Agent Avatar">
+                        <input type="text" id="agent-name-input" value="{{ agent.name }}" {% if agent.creator_not_self %} disabled {% endif %}>
+                    </div>
+                    <div id="agent-instructions">Instructions</div>
+                    <div id="agent-tuning">
+                        <p>{{ agent.tuning }}</p>
```
debanjum, line 29: Should this be updated to use `agent.personality` now?

```suggestion
                    <div id="agent-tuning">
                        <p>{{ agent.personality }}</p>
```

src/khoj/interface/web/agent.html
```diff
@@ -0,0 +1,286 @@
...
+                    </div>
+                    <div class="divider"></div>
+                    <div id="agent-public">
+                        <p>Public</p>
+                        <label class="switch">
+                            <input type="checkbox" {% if agent.public %} checked {% endif %} {% if agent.creator_not_self %} disabled {% endif %}>
+                            <span class="slider round"></span>
+                        </label>
+                    </div>
+                    <p id="agent-creator" style="display: none;">Creator: {{ agent.creator }}</p>
```
debanjum, line 39: Will this show as `Creator: null` for admin created agents? If so, should we instead do something like:
```suggestion
                    <p id="agent-creator" style="display: none;">Creator: {{ agent.creator || "admin" }}</p>
```

src/khoj/interface/web/agent.html
debanjum, line 260: Does this open the first chat session with the agent (if you already have a chat session with that agent) or is there only 1 chat session that can ever be associated with an agent? A 1-1 link between an agent and chat session seems a bit too strong, IMO

src/khoj/processor/content/markdown/markdown_to_entries.py
```diff
@@ -114,7 +114,7 @@ def convert_markdown_entries_to_maps(parsed_entries: List[str], entry_to_file_ma
             # Append base filename to compiled entry for context to model
             # Increment heading level for heading entries and make filename as its top level heading
             prefix = f"# {stem}\n#" if heading else f"# {stem}\n"
-            compiled_entry = f"{prefix}{parsed_entry}"
+            compiled_entry = f"{entry_filename}\n{prefix}{parsed_entry}"
```
debanjum, line 117: Do we need to do this here? Seems like the `entry_filename` is being used via `prefix` -> `stem` -> `Path(raw_filename).stem`

src/khoj/routers/api_agents.py
```diff
@@ -0,0 +1,43 @@
...
+    user: KhojUser = request.user.object if request.user.is_authenticated else None
+    agents = await AgentAdapters.aget_all_accessible_agents(user)
+    agents_packet = list()
+    for agent in agents:
+        agents_packet.append(
+            {
+                "slug": agent.slug,
+                "avatar": agent.avatar,
+                "name": agent.name,
+                "tuning": agent.tuning,
```
debanjum, line 33: nit: s/tuning/personality/?

```suggestion
                "personality": agent.personality,
```

src/khoj/routers/auth.py
```diff
@@ -117,9 +119,9 @@ async def auth(request: Request):
                 metadata={"user_id": str(khoj_user.uuid)},
             )
             logger.log(logging.INFO, f"New User Created: {khoj_user.uuid}")
-            RedirectResponse(url="/?status=welcome")
+            return RedirectResponse(url=f"{next_url}", status_code=HTTP_302_FOUND)
 
-    return RedirectResponse(url="/")
+    return RedirectResponse(url=f"{next_url}")
```
debanjum, line 124: Nice!

src/khoj/routers/auth.py
```diff
@@ -117,9 +119,9 @@ async def auth(request: Request):
                 metadata={"user_id": str(khoj_user.uuid)},
             )
             logger.log(logging.INFO, f"New User Created: {khoj_user.uuid}")
-            RedirectResponse(url="/?status=welcome")
+            return RedirectResponse(url=f"{next_url}", status_code=HTTP_302_FOUND)
```
debanjum, line 122: Should we still add the `status=welcome` query param here just to use when we add a first-run experience (FRE)?

```suggestion
            return RedirectResponse(url=f"{next_url}/status=welcome", status_code=HTTP_302_FOUND)
```

src/khoj/routers/api_chat.py
```diff
@@ -291,6 +308,14 @@ async def chat(
             response_obj = {"response": no_entries_found_format}
             return Response(content=json.dumps(response_obj), media_type="text/plain", status_code=200)
 
+    if conversation_commands == [ConversationCommand.Notes] and is_none_or_empty(compiled_references):
```
debanjum, line 311: Nice, were we previously not returning no notes found if chat query with `/notes` command returned no results? Do we differentiate between notes command auto-selected by Khoj vs by user? If Khoj has auto-selected just the `/notes` command and no notes are found should we not revert to general in that case instead of returning `No notes found"?

src/khoj/main.py
```diff
@@ -160,7 +160,9 @@ def start_server(app, host=None, port=None, socket=None):
     if socket:
         uvicorn.run(app, proxy_headers=True, uds=socket, log_level="debug", use_colors=True, log_config=None)
     else:
-        uvicorn.run(app, host=host, port=port, log_level="debug", use_colors=True, log_config=None)
+        uvicorn.run(
+            app, host=host, port=port, log_level="debug", use_colors=True, log_config=None, timeout_keep_alive=60
```
debanjum, line 164: What does the timeout_keep_alive do here? Is this required?

debanjum (approved):
Looks great, did a code review pass. I'll play around with the agent UX in a bit. But feel free to merge once you think it's ready!

src/khoj/routers/api_chat.py
sabaimran: I don't think we should just revert to default in that case because the tool selection process only selecting `notes` is a pretty high signal that that would be the only tool desired. Most likely, the user would be trying to extract something from their notes and not have anything relevant, so this is useful context to return back.

src/khoj/database/adapters/__init__.py
```diff
@@ -423,12 +444,18 @@ def get_default_agent():
...
+            agent.personality = default_personality
+            agent.chat_model = default_conversation_config
+            agent.slug = AgentAdapters.DEFAULT_AGENT_SLUG
+            agent.name = AgentAdapters.DEFAULT_AGENT_NAME
```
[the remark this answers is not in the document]
sabaimran: Technically speaking? Yes, none of them are read-only. Practically speaking, a code change would have to be done to change these, which is possible.

src/khoj/database/models/__init__.py
sabaimran: Hmm, shouldn't be. Good point, I'll update the model.

src/khoj/database/models/__init__.py
sabaimran: This wouldn't work in the case that there's a collision. That's why I'm using the `choice` function while extracting the values that are already seen.

src/khoj/interface/web/agent.html
sabaimran: No, it wouldn't show the field if the `creator` is not populated. There would just be a notice at the bottom calling out that it's managed by the admin.

src/khoj/interface/web/agent.html
sabaimran: It creates a new chat session with the given agent, rather than opening up the most recent chat session with that agent.

src/khoj/processor/content/markdown/markdown_to_entries.py
sabaimran: that would get only the file name, but not the whole path. i think this loses important metadata if not included (e.g., look at the research paper in my downloads folder about orangutans).

src/khoj/routers/auth.py
sabaimran: Yes, let's do that once we add the FRE. Confusing for developers otherwise if we have it in there before it's implemented and not being used.

src/khoj/main.py
sabaimran: It allows clients to maintain more long-living sessions with the server. This can help minimize the overhead of establishing a connection to the backend, therefore providing faster responses. This specific code only affects people self-hosting locally.

sabaimran: It allows clients to maintain more long-living sessions with the server. This can help minimize the overhead of establishing a connection to the backend, therefore providing faster responses. This specific code only affects people self-hosting locally.

PR #680, opened 2024-03-20
Migrate to Llama.cpp for Offline Chat

debanjum:
## Benefits
- Support all GGUF format chat models
- Support more GPUs like AMD, Nvidia, Mac, Vulcan (previously just Vulcan, Mac)
- Support more capabilities like larger context window, schema enforcement, speculative decoding etc.

### Major
- 978ebfee Use llama.cpp for offline chat models
  - Support larger context window
  - Automatically apply appropriate chat template. So offline chat models not using llama2 format are now supported
  - [shiny new default offline chat model](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B)
- f8ba541a Enable extract queries actor to improve notes search with offline chat
- aafb8785 Update documentation to use llama.cpp for offline chat in Khoj

### Minor
- e131de9c Use [NouseResearch's Hermes-2-Pro](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B) as default offline chat model in khoj.yml
- a05c0f17 Rename GPT4AllChatProcessor* to OfflineChatProcessor Config, Model
- a52d1d7d Only add location to image prompt generator when location known


src/khoj/database/models/__init__.py
```diff
@@ -160,7 +160,7 @@ class ModelType(models.TextChoices):
 
     max_prompt_size = models.IntegerField(default=None, null=True, blank=True)
     tokenizer = models.CharField(max_length=200, default=None, null=True, blank=True)
-    chat_model = models.CharField(max_length=200, default="mistral-7b-instruct-v0.1.Q4_0.gguf")
+    chat_model = models.CharField(max_length=200, default="NousResearch/Hermes-2-Pro-Mistral-7B-GGUF")
```
sabaimran, line 163: For folks who are using the old model (`mistral-7b-instruct-v0.1.Q4_0.gguf`), will the offline code path still work?

pyproject.toml
```diff
@@ -62,8 +62,7 @@ dependencies = [
     "pymupdf >= 1.23.5",
     "django == 4.2.10",
     "authlib == 1.2.1",
-    "gpt4all == 2.1.0; platform_system == 'Linux' and platform_machine == 'x86_64'",
-    "gpt4all == 2.1.0; platform_system == 'Windows' or platform_system == 'Darwin'",
+    "llama-cpp-python == 0.2.56",
```
sabaimran, line 65: Really neat being able to simplify this.

documentation/docs/get-started/setup.mdx
```diff
@@ -97,24 +97,44 @@ sudo -u postgres createdb khoj --password
...
 Run the following command in your terminal to install the Khoj backend.
 
 ```mdx-code-block
   <Tabs groupId="operating-systems">
     <TabItem value="macos" label="MacOS">
     ```shell
+# ARM/M1+ Machines
+MAKE_ARGS="-DLLAMA_METAL=on" python -m pip install khoj-assistant
+
+# Intel Machines
```
sabaimran, line 111: Hmm, could we have this automatically determine OS/GPU and do this in code during setup?

src/khoj/migrations/migrate_offline_chat_default_model_2.py
```diff
@@ -0,0 +1,71 @@
...
+            f"Upgrading config schema to {schema_version} from {previous_version} to change default (offline) chat model to mistral GGUF"
+        )
+        raw_config["version"] = schema_version
+
+        # Update offline chat model to use Nous Research's Hermes-2-Pro GGUF in path format suitable for llama-cpp
+        offline_chat_model = raw_config["processor"]["conversation"]["offline-chat"]["chat-model"]
+        if offline_chat_model == "mistral-7b-instruct-v0.1.Q4_0.gguf":
+            raw_config["processor"]["conversation"]["offline-chat"][
+                "chat-model"
+            ] = "NousResearch/Hermes-2-Pro-Mistral-7B-GGUF"
```
sabaimran, line 68: Ah, this should answer my previous question. Would other models users have configured that work with gpt4all still work? As in, if something is not the `mistral-7b-instruct-v0.1.Q4_0.gguf` default, will it continue to function?

src/khoj/processor/conversation/offline/utils.py
```diff
@@ -1,43 +1,59 @@
...
 
-def download_model(model_name: str):
+def download_model(repo_id: str, filename: str = "*Q4_K_M.gguf"):
     try:
-        import gpt4all
+        from llama_cpp.llama import Llama
     except ModuleNotFoundError as e:
-        logger.info("There was an error importing GPT4All. Please run pip install gpt4all in order to install it.")
+        logger.info(
+            "There was an error importing llama-cpp-python. Please run pip install llama-cpp-python in order to install it."
```
sabaimran, line 17: As this is going into the default packages used in `pyproject.toml`, this whole `try/except` block should be unnecessary, I believe. Installations would be caught when the package is first installed.

documentation/docs/features/chat.md
```diff
@@ -23,7 +23,7 @@ Offline chat stays completely private and works without internet using open-sour
...
 
 1. Open your [Khoj offline settings](http://localhost:42110/server/admin/database/offlinechatprocessorconversationconfig/) and click *Enable* on the Offline Chat configuration.
-2. Open your [Chat model options](http://localhost:42110/server/admin/database/chatmodeloptions/) and add a new option for the offline chat model you want to use. Make sure to use `Offline` as its type. We currently only support offline models that use the [Llama chat prompt](https://replicate.com/blog/how-to-prompt-llama#wrap-user-input-with-inst-inst-tags) format. We recommend using `mistral-7b-instruct-v0.1.Q4_0.gguf`.
+2. Open your [Chat model options](http://localhost:42110/server/admin/database/chatmodeloptions/) and add a new option for the offline chat model you want to use. Make sure to use `Offline` as its type. We support any [GGUF chat model](https://huggingface.co/models?library=gguf) for offline chat. For a balanced chat model that runs well on standard consumer hardware we recommend [Hermes-2-Pro-Mistral-7B by NousResearch](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B-GGUF).
```
sabaimran, line 26: Does this guidance need to be updated? https://docs.khoj.dev/features/chat/#offline-chat

sabaimran (commented):
Exciting getting llama.cpp running & having extract_questions working with the offline models.

src/khoj/migrations/migrate_offline_chat_default_model_2.py
debanjum: Yeah, this change wouldn't help folks who have changed the default model they use. They'll need to manually update the huggingface repo for the model they want to use

documentation/docs/get-started/setup.mdx
debanjum: That'd have been great but not sure how we can do that via the pip install mechanism given that llama.cpp builds python wheels on the client device to handle a wider range of GPUs/machines.

If our install step becomes too complex, we could ask folks to use an installer script (e.g `curl https://khoj.dev/install.sh | bash`) but for now this mechanism of using env variables should be good enough to start with as
- the self-hosted version is anyway for folks comfortable with the terminal
- other projects our using the same setup mechanism

src/khoj/database/models/__init__.py
debanjum: The migration script should help folks using the previous default model, rest of the folks will need to manually update/reset the chat model they are using

PR #695, opened 2024-04-10
Support Scheduling Tasks (Reminders: Part 1)

debanjum:
### Minimal Process
1. Detect when user intends to schedule a task, aka reminder
   - Support new `reminder` output mode to the response type chat actor
   - Show examples of selecting the reminder output mode to the response type chat actor
2. Extract schedule time (as cron timestring) and inferred query to run from user message
3. Use APScheduler to call chat with inferred query at scheduled time

### DIFF
--- a/src/khoj/main.py
+++ b/src/khoj/main.py
@@ -23,6 +23,7 @@ warnings.filterwarnings("ignore", message=r"legacy way to download files from th
 
 import uvicorn
 import django
+from apscheduler.schedulers.background import BackgroundScheduler
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.staticfiles import StaticFiles
@@ -125,6 +126,10 @@ def run(should_start_server=True):
 
     # Setup task scheduler
     poll_task_scheduler()
+
+    # Setup Background Scheduler
+    state.scheduler = BackgroundScheduler()
+    state.scheduler.start()
 
     # Start Server
     configure_routes(app)
--- a/src/khoj/processor/conversation/prompts.py
+++ b/src/khoj/processor/conversation/prompts.py
@@ -10,8 +10,7 @@ You were created by Khoj Inc. with the following capabilities:
 
 - You *CAN REMEMBER ALL NOTES and PERSONAL INFORMATION FOREVER* that the user ever shares with you.
 - Users can share files and other information with you using the Khoj Desktop, Obsidian or Emacs app. They can also drag and drop their files into the chat window.
-- You *CAN* generate images, look-up real-time information from the internet, and answer questions based on the user's notes.
-- You cannot set reminders.
+- You *CAN* generate images, look-up real-time information from the internet, set reminders and answer questions based on the user's notes.
 - Say "I don't know" or "I don't understand" if you don't know what to say or if you don't know the answer to a question.
 - Ask crisp follow-up questions to get additional context, when the answer cannot be inferred from the provided notes or past conversations.
 - Sometimes the user will share personal information that needs to be remembered, like an account ID or a residential address. These can be acknowledged with a simple "Got it" or "Okay".
@@ -300,6 +299,22 @@ AI: I can help with that. I see online that there is a new model of the Dell XPS
 Q: What are the specs of the new Dell XPS 15?
 Khoj: default
 
+Example:
+Chat History:
+User: Where did I go on my last vacation?
+AI: You went to Jordan and visited Petra, the Dead Sea, and Wadi Rum.
+
+Q: Remind me who did I go with on that trip?
+Khoj: default
+
+Example:
+Chat History:
+User: How's the weather outside? Current Location: Bali, Indonesia
+AI: It's currently 28°C and partly cloudy in Bali.
+
+Q: Share a painting using the weather for Bali every morning.
+Khoj: reminder
+
 Now it's your turn to pick the mode you would like to use to answer the user's question. Provide your response as a string.
 
 Chat History:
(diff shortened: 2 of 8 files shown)


src/khoj/processor/conversation/prompts.py
```diff
@@ -491,6 +499,35 @@
...
+User: Share the top weekly posts on Hacker News every Monday evening. Format it as a newsletter. My Current Location: Nairobi, Kenya
+Khoj: ["30 15 * * 1", "Top posts last week on Hacker News"]
+
+User: Notify me when version 2.0.0 of the Khoj python package is released. My Current Location: Mexico City, Mexico
+Khoj: ["0 16 * * *", "Check if latest Khoj python package version >= 2.0.0"]
+
+User: Share the latest news about tech in India on the first Sunday of every Month. My Current Location: Delhi, India
+Khoj: ["30 3 1-7 * 0", "Latest tech news in India"]
+
+User: {query}. My Current Location: {user_location}
```
sabaimran, line 526: Should we add an example where the intent is being inferred from the chat history? Like 
user: "summarize the tech news every morning" 
khoj: {provides summary of tech news}
user: "nice, give this to me everyday at 8 am"

khoj: ["0 8 * * *", "summarize the tech news"]

src/khoj/routers/api_chat.py
```diff
@@ -370,6 +373,55 @@ async def send_rate_limit_message(message: str):
...
+            crontime, inferred_query = await schedule_query(q, location)
+            trigger = CronTrigger.from_crontab(crontime)
+            common = CommonQueryParamsClass(
+                client=websocket.user.client_app,
+                user_agent=websocket.headers.get("user-agent"),
+                host=websocket.headers.get("host"),
+            )
+            scope = websocket.scope.copy()
+            scope["path"] = "/api/chat"
+            scope["type"] = "http"
```
sabaimran, line 386: Why is the scope type `http`?

sabaimran (commented):
Looks good for the most part, just have some comments about handling unknown time zones and prompt template. Super excited to use this once persistence and thread safety are enabled!

src/khoj/routers/api_chat.py
```diff
@@ -574,6 +629,42 @@ async def chat(
...
+            args=(partial_scheduled_chat, f"{ProcessLock.Operation.SCHEDULED_JOB}_{user.uuid}_{inferred_query}"),
+            id=f"job_{user.uuid}_{inferred_query}_{crontime}",
+            max_instances=2,  # Allow second instance to kill any previous instance with stale lock
+            jitter=30,
+        )
+
+        next_run_time = job.next_run_time.strftime("%Y-%m-%d %H:%M:%S")
+        llm_response = f"""
+        ### 🕒 Scheduled Job
+- Query: **"{inferred_query}"**
```
sabaimran, line 648: How should we handle jobs that can't actually be "run". For example, some actions wouldn't be accessible (`"Fill out the wait list for the new Taylor Swift album every morning`" or `"Tell me when the next slot becomes available for Xi's pottery class"`). How can we gracefully fail when/if that happens? How can we evaluate a failure?

src/khoj/main.py
```diff
@@ -126,6 +127,24 @@ def run(should_start_server=True):
...
+            "apscheduler.jobstores.default": {
+                "type": "sqlalchemy",
+                "url": f'postgresql://{django_db["USER"]}:{django_db["PASSWORD"]}@{django_db["HOST"]}:{django_db["PORT"]}/{django_db["NAME"]}',
+            },
+            "apscheduler.timezone": "UTC",
+            "apscheduler.job_defaults.misfire_grace_time": "60",  # Useful to run scheduled jobs even when worker delayed because it was busy or down
+            "apscheduler.job_defaults.coalesce": "true",  # Combine multiple jobs into one if they are scheduled at the same time
+        }
+    )
+    # state.scheduler.add_jobstore(DjangoJobStore(), "default")
```
sabaimran, line 145: Should this be commented out?

src/khoj/main.py
```diff
@@ -126,6 +127,24 @@ def run(should_start_server=True):
...
 
+    # Setup Background Scheduler
+    from django.conf import settings as django_settings
+
+    django_db = django_settings.DATABASES["default"]
+    state.scheduler = BackgroundScheduler(
+        {
+            "apscheduler.jobstores.default": {
+                "type": "sqlalchemy",
+                "url": f'postgresql://{django_db["USER"]}:{django_db["PASSWORD"]}@{django_db["HOST"]}:{django_db["PORT"]}/{django_db["NAME"]}',
```
sabaimran, line 138: Can you explain how this works/what this does?

src/khoj/main.py
```diff
@@ -126,6 +127,24 @@ def run(should_start_server=True):
...
     poll_task_scheduler()
 
+    # Setup Background Scheduler
+    from django.conf import settings as django_settings
+
+    django_db = django_settings.DATABASES["default"]
+    state.scheduler = BackgroundScheduler(
+        {
+            "apscheduler.jobstores.default": {
+                "type": "sqlalchemy",
```
sabaimran, line 137: Should this be django?
```suggestion
                "type": "django",
```

src/khoj/main.py
```diff
@@ -126,6 +127,24 @@ def run(should_start_server=True):
     # Setup task scheduler
     poll_task_scheduler()
 
+    # Setup Background Scheduler
+    from django.conf import settings as django_settings
```
sabaimran, line 131: Can this go to `settings.py`?

src/khoj/utils/helpers.py
```diff
@@ -325,6 +327,7 @@ class ConversationCommand(str, Enum):
 
 mode_descriptions_for_llm = {
     ConversationCommand.Image: "Use this if you think the user is requesting an image or visual response to their query.",
+    ConversationCommand.Reminder: "Use this if you think the user is requesting a response at a scheduled date, time.",
```
sabaimran, line 330:
```suggestion
    ConversationCommand.Reminder: "Use this if you think the user is requesting a response at a scheduled date or time.",
```

src/khoj/routers/api_chat.py
debanjum: That's just the type of connection, not http over https. Anyway that code has been changed now

src/khoj/main.py
debanjum: That can be removed, It isn't required for job persistence. Just gives a Django admin panel view into managing jobs

src/khoj/main.py
debanjum: APScheduler uses the postgres database (specified in `url`) as the job store. Job stores persist scheduled jobs in a DB table maintained by AP Scheduler. This code is pointing to the DB to use for maintaining scheduled job state by APScheduler

src/khoj/main.py
debanjum: APScheduler automatically comes with a SQLAlchemy job store but not a Django job store. So to avoid adding another dependency like `django-apscheduler` to Khoj I decided to stick with using the default SQLAlchemy job store to connect with the Khoj DB

src/khoj/main.py
debanjum: We import the `django_settings` here to build the DB URL to pass to APScheduler for scheduled job persistence.

src/khoj/routers/api_chat.py
debanjum: I think in the general case, this is a more complex question that we cannot automate IMHO. We can track failed job runs etc. but to track if the response satisfied the user's request, we'd need to ask the user from what I can imagine.

To track normal failures, the scheduled jobs UI can maybe show the status of the last run (pass ✅, fail ⛔️)?

src/khoj/processor/conversation/prompts.py
```diff
@@ -491,6 +499,35 @@
...
 )
 
+# Schedule task
+# --
+crontime_prompt = PromptTemplate.from_template(
+    """
+You are Khoj, an extremely smart and helpful task scheduling assistant
+- Given a user query, you infer the date, time to run the query at as a cronjob time string
+- The cron job should run in {server_timezone} time zone
+- Infer user's time zone from the current location provided in their message
```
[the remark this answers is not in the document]
debanjum: For Whatsapp users:
- Using their phone number to infer timezone maybe a good default?
- We can allow folks to configure their location, scheduled tasks via a /configure command or Whatsapp form UI at some point?
- We can't message after 24 hours since user messaged Khoj on Whatsapp, right? If so, maybe we should warn them of that when they create a scheduled task

For VPN users:
- The degradation in experience due to incorrect inferred location already exists for Khoj users. To mitigate this we should allow folks to explicitly configure their location in their settings. Khoj should use that location instead of trying to infer location based on I.P if it is explicitly set by users

In my opinion, these are good follow-on tasks to tackle after the MVP of scheduled tasks is out. What do you think?

src/khoj/main.py
debanjum: Note, this change has split out into #714 to keep PRs small for easier reviewing

src/khoj/routers/api_chat.py
debanjum: I'm merging this PR into it's feature branch but please feel free to comment here or in the Reminders UX PR at #715

src/khoj/processor/conversation/prompts.py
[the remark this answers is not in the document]
debanjum: I'm merging this PR into it's feature branch but please feel free to comment here or in the Scheduled Tasks Prompt improvement PR at #718

PR #703, opened 2024-04-13
Improve Chat Page Load Perf, Offline Chat Perf and Miscellaneous Fixes

debanjum:
### Store Generated Images as WebP 
- 78bac4ae Add migration script to convert PNG to WebP references in database
- c6e84436 Update clients to support rendering webp images inline
- d21f22ff Store Khoj generated images as webp instead of png for faster loading

### Lazy Fetch Chat Messages to Improve Time, Data to First Render
This is especially helpful for long conversations with lots of images
- 128829c4 Render latest msgs on chat session load. Fetch, render rest as they near viewport
- 9e558577 Support getting latest N chat messages via chat history API

### Intelligently set Context Window of Offline Chat to Improve Performance
- 4977b551 Use offline chat prompt config to set context window of loaded chat model

### Fixes
- 148923c1 Fix to raise error on hitting rate limit during Github indexing
- b8bc6bee Always remove loading animation on Desktop app if can't login to server
- 38250705 Fix `get_user_photo` to only return photo, not user name from DB

### Miscellaneous Improvements
- 689202e0 Update recommended CMAKE flag to enable using CUDA on linux in Docs
- b820daf3 Makes logs less noisy


documentation/docs/get-started/setup.mdx
```diff
@@ -134,7 +134,7 @@ python -m pip install khoj-assistant
  # CPU
  python -m pip install khoj-assistant
  # NVIDIA (CUDA) GPU
- CMAKE_ARGS="DLLAMA_CUBLAS=on" FORCE_CMAKE=1 python -m pip install khoj-assistant
```
sabaimran, line 137: What the previous argument incorrect? Or what effect would it have had?

src/khoj/database/migrations/0035_convert_png_to_webp.py
```diff
@@ -0,0 +1,77 @@
...
+                    # Encode the WebP image back to base64
+                    webp_image_bytes = webp_image_io.getvalue()
+                    chat["message"] = base64.b64encode(webp_image_bytes).decode()
+                    chat["intent"]["type"] = "text-to-image-v3"
+                    webp_image_io.close()
+
+            if chat["by"] == "khoj" and chat["intent"]["type"] == "text-to-image2":
+                print("❗️ Please MANUALLY update PNG images created by Khoj in your AWS S3 bucket to WebP format.")
+                # Convert PNG url to WebP url
+                chat["message"] = chat["message"].replace(".png", ".webp")
```
sabaimran, line 34: I think we can just leave these as pngs, as self-hosted folks probably won't go through the effort of migrating & would remain on the `text-to-image2` type. We don't even really need to output the warning, the end user shouldn't have to worry about the underlying image type. 

You can create a custom management command for anyone self-hosting who might want to migrate to run this operation: https://docs.djangoproject.com/en/5.0/howto/custom-management-commands/.

src/khoj/interface/web/chat.html
```diff
@@ -312,6 +314,8 @@
                     imageMarkdown = `![](data:image/png;base64,${message})`;
                 } else if (intentType === "text-to-image2") {
                     imageMarkdown = `![](${message})`;
+                } else if (intentType === "text-to-image-v3") {
+                    imageMarkdown = `![](data:image/webp;base64,${message})`;
```
sabaimran, line 318: It would still be using a URL, right? Is the image type specification required, or a nice to have?

src/khoj/processor/conversation/offline/utils.py
```diff
@@ -23,23 +24,33 @@ def download_model(repo_id: str, filename: str = "*Q4_K_M.gguf"):
...
-            chat_model = Llama(model_path, **kwargs)
-        else:
-            chat_model = Llama.from_pretrained(repo_id=repo_id, filename=filename, **kwargs)
+        chat_model = load_model(model_path, repo_id, filename, kwargs)
 
-    logger.debug(f"{'Loaded' if model_path else 'Downloaded'} chat model to {device.upper()}")
+    # Now load the model with context size set based on:
+    # 1. context size supported by model and
+    # 2. configured size or machine (V)RAM
+    kwargs["n_ctx"] = infer_max_tokens(chat_model.n_ctx(), max_tokens)
```
sabaimran, line 36: Psyched for this fix 🔥 should be such a quality of life improvement for people self-hosting the offline version.

src/khoj/routers/helpers.py
```diff
@@ -550,21 +558,38 @@ async def text_to_image(
...
+
             with timer("Upload image to S3", logger):
-                image_url = upload_image(image, user.uuid)
-            return image, status_code, improved_image_prompt, image_url
+                image_url = upload_image(webp_image_bytes, user.uuid)
+            if image_url:
+                intent_type = "text-to-image-v2"
+            else:
+                intent_type = "text-to-image-v3"
+                image = base64.b64encode(webp_image_bytes).decode("utf-8")
```
sabaimran, line 578: Let's use an enum for these image types and include descriptions of what they imply? Hard to follow and maintain the code otherwise.

src/khoj/interface/web/chat.html
sabaimran: Update, the url-based `text-to-image` specification is just `text-to-image-v2`.

documentation/docs/get-started/setup.mdx
debanjum: This is just an update based on the [fixed documentation of llama-cpp-python](https://github.com/abetlen/llama-cpp-python?tab=readme-ov-file#supported-backends)

src/khoj/database/migrations/0035_convert_png_to_webp.py
debanjum: Good point. Created, tested the management command to ensure it works

PR #706, opened 2024-04-15
Add a database lock for jobs that shouldn't be run by multiple workers

sabaimran:
- Some jobs (such as background indexing for Notion, GitHub) should just be run once per day. Because some setups have multiple simultaneous workers for Khoj, these are all triggered around the same time and then run for a while. Prevent this behavior by using a time-restricted DB lock that ensures only one given job is doing indexing at a given moment.


src/khoj/configure.py
```diff
@@ -303,9 +304,16 @@ def configure_middleware(app):
     app.add_middleware(SessionMiddleware, secret_key=os.environ.get("KHOJ_DJANGO_SECRET_KEY", "!secret"))
 
 
-@schedule.repeat(schedule.every(22).to(26).hours)
+@schedule.repeat(schedule.every(22).to(25).hours)
 def update_search_index():
     try:
+        if ProcessLockAdapters.is_process_locked(ProcessLock.Operation.UPDATE_EMBEDDINGS):
+            logger.info("🔒 Skipping update search index due to lock")
```
debanjum, line 311: nit: Let's consistently use the term "content index" instead of "search index"
```suggestion
            logger.info("🔒 Skipping update content index due to lock")
```

src/khoj/configure.py
```diff
@@ -303,9 +304,16 @@ def configure_middleware(app):
...
+            return
+        ProcessLockAdapters.set_process_lock(
+            ProcessLock.Operation.UPDATE_EMBEDDINGS, max_duration_in_seconds=60 * 60 * 12
```
debanjum, line 314: Should the max duration of the lock be shorter than 12 hours? I'd have expected something closer in the range of a few minutes (12-30mins)

src/khoj/configure.py
```diff
@@ -303,9 +304,16 @@ def configure_middleware(app):
     app.add_middleware(SessionMiddleware, secret_key=os.environ.get("KHOJ_DJANGO_SECRET_KEY", "!secret"))
 
 
-@schedule.repeat(schedule.every(22).to(26).hours)
+@schedule.repeat(schedule.every(22).to(25).hours)
 def update_search_index():
     try:
+        if ProcessLockAdapters.is_process_locked(ProcessLock.Operation.UPDATE_EMBEDDINGS):
```
debanjum, line 310: Should we also add a check here to skip running the job if the Operation succeeded in the last ~24 hours?

src/khoj/database/adapters/__init__.py
```diff
@@ -405,6 +406,32 @@ async def aget_user_search_model(user: KhojUser):
...
+        if process_lock.started_at + timedelta(seconds=process_lock.max_duration_in_seconds) < datetime.now(
+            tz=timezone.utc
+        ):
+            process_lock.delete()
+            return False
+        return True
+
+    @staticmethod
+    def remove_process_lock(process_name: str):
+        return ProcessLock.objects.filter(name=process_name).delete()
```
debanjum, line 432: Not sure, but wondering if instead of creating and deleting the process locks we should update the state of the process locks instead?

This will allow other background processes to gather from the process lock DB table:
1. When a specific operation last ran (e.g via a new `completed_at` field in table)
2. If the previous operation run failed or succeeded (e.g via new `operation_state` field in table with new, started, failed, succeeded choices)

debanjum (approved):
Excited about these changes! It'll eliminate duplicate, parallel indexing jobs in multi-worker setups and could be super useful for the scheduled jobs feature as well

src/khoj/configure.py
sabaimran: Hmmmmmm, I actually don't know how long the complete indexing job generally takes. But maybe we can set it to 2 or 3 hours to be safe. I'll also add a timer here so we can collect some information about the expected run time of this job.

sabaimran: And in this range, I'd rather err on the side of there being a single job that runs a long time, rather than risk having multiple workers spin off the same indexing job again.

src/khoj/database/adapters/__init__.py
sabaimran: > When a specific operation last ran (e.g via a new completed_at field in table)

Hmm. I think this would be moving away from the purpose of acting as a db lock, and more moving into a role as a job manager. I think we should take care to avoid conflating those concepts. If we have a separate place where we manage `job`s generally, we should definitely add metadata like that in there. 

> If the previous operation run failed or succeeded (e.g via new operation_state field in table with new, started, failed, succeeded choices)

If there's a DB lock with a certain name present in the table, we implicitly have this information. If it's started, the process lock is there. If it's failed (without the process exiting), the lock would be removed. If it's succeeded, the lock would be removed. 

The tricky thing is when it fails silently, leaving us in a deadlock situation until the `max_duration_in_seconds` kicks in.

PR #712, opened 2024-04-19
Make conversations optionally shareable

sabaimran:
# Incoming
- Shared conversations are viewable by anyone, without a login wall
- Can share a conversation from the three dot menu
- Add a new model for Public Conversation
- The rationale for a separate model is that public and private conversations have different assumptions. Separating them reduces some of the code specificity on our server-side code and allows us for easier interpretation and stricter security. Separating the data model makes it harder to accidentally view something that was meant to be private
- Add a new, read-only view for public conversations. It has a list of available agents on the side panel.


debanjum (approved):
This is great! It'll simplify sharing responses by Khoj with others instead of the previous clunky copy/paste each message + mention it was generated by Khoj mechanism

src/khoj/routers/api_chat.py
```diff
@@ -128,6 +133,60 @@ def chat_history(
     return {"status": "ok", "response": meta_log}
 
 
+@api_chat.get("/public/history")
```
debanjum, line 136: Maybe we can call this `/public/chat`? It'll generalize better once we allow for user created public agents and automations

src/khoj/routers/api_chat.py
```diff
@@ -150,6 +209,66 @@ async def clear_chat_history(
...
+    request: Request,
+    common: CommonQueryParams,
+    public_conversation_slug: str,
+):
+    user = request.user.object
+
+    # Load Conversation History
+    public_conversation = PublicConversationAdapters.get_public_conversation_by_slug(public_conversation_slug)
+
+    # Duplicate Conversation History to Public Conversation
```
debanjum, line 224: nit: This comment seems incorrect. It's doing the reverse here?
```suggestion
    # Duplicate Public Conversation to Conversation History
```

src/khoj/routers/api_chat.py
```diff
@@ -150,6 +209,66 @@ async def clear_chat_history(
...
+        **common.__dict__,
+        metadata=chat_metadata,
+    )
+
+    redirect_uri = str(request.app.url_path_for("chat_page"))
+
+    return Response(status_code=200, content=json.dumps({"status": "ok", "next_url": redirect_uri}))
+
+
+@api_chat.post("/public/duplicate")
```
debanjum, line 244: The /duplicate name of the api endpoints seems unclear as to what it does IMO. The directionality of the duplication isn't specified, so it could mean duplicate from private to public or public to private. Maybe just POST "/public" should be good enough?

```suggestion
@api_chat.post("/public")
```

src/khoj/routers/web_client.py
```diff
@@ -364,3 +365,56 @@ def computer_config_page(request: Request):
             "khoj_version": state.khoj_version,
         },
     )
+
+
+@web_client.get("/view/{public_conversation_slug}", response_class=HTMLResponse)
```
debanjum, line 370: Should we call it `/public/chat`? It will generalize better to public agents and automations later as well

```suggestion
@web_client.get("/public/chat/{public_conversation_slug}", response_class=HTMLResponse)
```

sabaimran: How about `share` instead? Nice and simple.

src/khoj/database/models/__init__.py
```diff
@@ -233,6 +234,32 @@ class Conversation(BaseModel):
...
+        slug = re.sub(r"\W+", "-", instance.slug.lower())[:50]
+        observed_random_numbers = set()
+        while PublicConversation.objects.filter(slug=slug).exists():
+            try:
+                random_number = choice([i for i in range(0, 1000) if i not in observed_random_numbers])
+            except IndexError:
+                raise ValidationError(
+                    "Unable to generate a unique slug for the Public Conversation. Please try again later."
+                )
+            observed_random_numbers.add(random_number)
```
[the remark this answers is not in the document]
sabaimran: Yes, that's exactly the purpose. I don't think the collision incidence would be that much of a problem, because the slug would have to exactly match in order for that to happen. But can update, it would be nicer to have the shareable links have uniform random strings.

src/khoj/routers/web_client.py
```diff
@@ -367,6 +368,59 @@ def computer_config_page(request: Request):
     )
 
 
+@web_client.get("/share/{public_conversation_slug}", response_class=HTMLResponse)
```
debanjum, line 371: Should we put this under /share/chat? It'll make adding /share/agent, /share/automaton later easier?

src/khoj/routers/api_chat.py
```diff
@@ -132,6 +137,60 @@ def chat_history(
...
+        if n > 0 and meta_log.get("chat"):
+            meta_log["chat"] = meta_log["chat"][-n:]
+        # Else return all messages except latest N
+        elif n < 0 and meta_log.get("chat"):
+            meta_log["chat"] = meta_log["chat"][:n]
+
+    update_telemetry_state(
+        request=request,
+        telemetry_type="api",
+        api="chat_history",
```
debanjum, line 187: api field doesn't match current api name
```suggestion
        api="get_chat_share",
```

src/khoj/routers/api_chat.py
```diff
@@ -132,6 +137,60 @@ def chat_history(
     return {"status": "ok", "response": meta_log}
 
 
+@api_chat.get("/share/history")
+def chat_public_history(
```
debanjum, line 141: Should we simplify the endpoint to get /share?

```suggestion
@api_chat.get("/share")
def get_shared_chat(
```

src/khoj/routers/api_chat.py
```diff
@@ -154,6 +213,66 @@ async def clear_chat_history(
...
+    conversation = ConversationAdapters.get_conversation_by_user(user, request.user.client_app, conversation_id)
+
+    public_conversation = ConversationAdapters.make_public_conversation_copy(conversation)
+
+    public_conversation_url = PublicConversationAdapters.get_public_conversation_url(public_conversation)
+
+    update_telemetry_state(
+        request=request,
+        telemetry_type="api",
+        api="duplicate_chat_history_public_conversation",
```
debanjum, line 267:
```suggestion
        api=post_chat_share",
```

src/khoj/interface/web/chat.html
```diff
@@ -1562,11 +1562,79 @@
...
+                                shareButton.addEventListener('click', function(event) {
+                                    event.preventDefault();
+                                    let confirmation = confirm('Are you sure you want to share this chat session? This will make the conversation public.');
+                                    if (!confirmation) return;
+                                    let duplicateURL = `/api/chat/share?client=web&conversation_id=${incomingConversationId}`;
+                                    fetch(duplicateURL , { method: "POST" })
+                                        .then(response => response.ok ? response.json() : Promise.reject(response))
+                                        .then(data => {
+                                            if (data.status == "ok") {
+                                                flashStatusInChatInput("✅ Conversation shared successfully");
```
debanjum, line 1579: Should we copy link to the public conversation onto the clipboard automatically here when the conversation is created? It reduces steps required to share a conversation

sabaimran: Yup, I added that in.

src/khoj/routers/api_chat.py
sabaimran: I think it makes sense to leave it as `/history` for nice parity with the private method. Can update the name.

PR #721, opened 2024-04-23
Add Feedback Buttons on Chat

MythicalCow:
### Description and Rationale for Changes
This feature includes thumbs up and thumbs down buttons on Khoj's chat responses that provide automated feedback. When a thumbs up/down button is clicked, the code sends an email to team@khoj.dev with the following:
* user query
* khoj's response
* whether the sentiment of the user was good or bad. 

This is critical in improving Khoj's nondeterministic LLM model for a better user experience.
### List of Changes
* new endpoint in `api_chat.py` (/feedback) that can be used to trigger mail sending).
* thumbs up and thumbs down buttons implemented in `chat.html`
* new function in `routers/email.py` to handle feedback email sending via resend
* `feedback.html` template for a formatted email with the feedback.
### Demo Images
Screenshot from 2024-04-22 20-01-52
Screenshot 2024-04-23 142314


src/khoj/interface/web/chat.html
```diff
@@ -336,6 +357,21 @@
 
             return renderMessage(message, by, dt, references, false, "return");
         }
+        //handler function for posting feedback data to endpoint
+        function sendFeedback(_uquery="", _kquery="", _sentiment="") {
+            console.log("Function Call");
```
sabaimran, line 362: Please remove the debug logs.

src/khoj/interface/web/chat.html
```diff
@@ -372,7 +408,47 @@
...
+                var thumbsUpButtons = document.querySelectorAll('.thumbs-up-button');
+                var thumbsDownButtons = document.querySelectorAll('.thumbs-down-button');
+
+                //clear out old feedback buttons as we can't access this data anymore
+                thumbsUpButtons.forEach(function(element) {
+                    element.parentNode.removeChild(element);
+                });
+                thumbsDownButtons.forEach(function(element) {
+                    element.parentNode.removeChild(element);
+                });
```
sabaimran, line 421: Why do you need to do this? Each message should be able to keep its feedback buttons. The data should still be accessible for the user/AI query.

src/khoj/processor/conversation/prompts.py
```diff
@@ -432,7 +432,7 @@
 - Break messages into multiple search queries when required to retrieve the relevant information.
 - Use site: google search operators when appropriate
 - You have access to the the whole internet to retrieve information.
-- Official, up-to-date information about you, Khoj, is available at site:khoj.dev, github or pypi.
+- Official, up-to-date information about you, Khoj, is available at site:khoj.dev
```
sabaimran, line 435: Please revert the changes in this file 🙂

src/khoj/routers/api_chat.py
```diff
@@ -58,6 +58,19 @@
 
 api_chat = APIRouter()
 
+from khoj.routers.email import feedback
+from pydantic import BaseModel
+
+
+class TextData(BaseModel):
```
sabaimran, line 65:
```suggestion
class FeedbackData(BaseModel):
```

It's better to be more specific in the nomenclature to make it clear what a given class is meant for.

src/khoj/routers/email.py
```diff
@@ -47,3 +46,38 @@ async def send_welcome_email(name, email):
...
+
+
+class TextData(BaseModel):
+    uquery: str
+    kquery: str
+    sentiment: str
+
+
+async def feedback(uquery, kquery, sentiment):
+    # console debug messages
```
sabaimran, line 61: nit: you can remove this

src/khoj/routers/email.py
```diff
@@ -47,3 +46,38 @@ async def send_welcome_email(name, email):
...
     )
+
+
+from pydantic import BaseModel
+
+
+class TextData(BaseModel):
+    uquery: str
+    kquery: str
+    sentiment: str
```
sabaimran, line 57: This can be removed from here.

src/khoj/routers/email.py
```diff
@@ -47,3 +46,38 @@ async def send_welcome_email(name, email):
...
+
+
+async def feedback(uquery, kquery, sentiment):
```
sabaimran, line 60: With function naming convention, it's preferred to name it so that the action being completed is super clear just by reading it.

Also adding a parameter for user email.

```suggestion
async def send_query_feedback(uquery, kquery, sentiment, user_email):
```

src/khoj/routers/api_chat.py
```diff
@@ -58,6 +58,19 @@
...
+    uquery: str
+    kquery: str
+    sentiment: str
+
+@api_chat.post("/feedback")
+async def sendfeedback(data: TextData):
+    await feedback(data.uquery,data.kquery,data.sentiment)
```
sabaimran, line 72: We should limit usage of this API only to authenticated users. And we should also include the email in case we need to follow-up.

```suggestion
@api_chat.post("/feedback")
@requires(["authenticated"])
async def sendfeedback(request: Request, data: TextData):
    user: KhojUser = request.user.object
    await feedback(data.uquery, data.kquery, data.sentiment, user.email)
```

src/khoj/routers/email.py
```diff
@@ -47,3 +46,38 @@ async def send_welcome_email(name, email):
...
+    logger.debug(f"User Query: {uquery}\n")
+    logger.debug(f"Khoj Response: {kquery}\n")
+    logger.debug(f"Sentiment: {sentiment}\n")
+    # rendering feedback email using feedback.html as template
+    template = env.get_template("feedback.html")
+    html_content = template.render(
+        uquery=uquery if not is_none_or_empty(uquery) else "N/A",
+        kquery=kquery if not is_none_or_empty(kquery) else "N/A",
+        sentiment=sentiment if not is_none_or_empty(sentiment) else "N/A",
+    )
```
sabaimran, line 72: Can you also include the `user_email` in the template?

src/khoj/routers/email.py
```diff
@@ -47,3 +46,38 @@ async def send_welcome_email(name, email):
...
+    # send feedback from two fixed accounts
+    r = resend.Emails.send(
```
sabaimran, line 74: Can you add a check here to see if resend is enabled before making the template/sending the email? If it's not enabled, then add the feedback to a log file, like `query_feedback.log`. Since some of our users are self-hosted, they might still want to aggregate the feedback, but won't be able to directly send the email.

sabaimran (commented):
This is awesome! It'll be super useful being able to collect in-app feedback. Just a few comments, but the structure is great.

src/khoj/processor/conversation/prompts.py
MythicalCow: These changes weren't directly made by me. They seem to have been made when I pulled the files from master. I'll have them removed!

src/khoj/routers/api_chat.py
MythicalCow: good point!

src/khoj/routers/email.py
MythicalCow: where can I obtain the user's email?

src/khoj/routers/api_chat.py
MythicalCow: Got it.

src/khoj/routers/email.py
MythicalCow: Sure!

src/khoj/processor/conversation/prompts.py
sabaimran: You may need to pull from master again.

src/khoj/routers/email.py
MythicalCow: Are you referring to the console debug messages?

sabaimran: Just the comment!

src/khoj/routers/email.py
sabaimran: Update, let's address the local feedback log file later on. Don't worry about it for now.

sabaimran (approved):

PR #723, opened 2024-04-24
Fix Chat UI and Indexing on Desktop App

debanjum:
- 49cae6a4 Make valid file extension checking case insensitive on Desktop app
- 48d4bf18 Skip indexing non-existent folders on Desktop app
- 502bb477 Pass auth headers to fix lazy load of chat messages on Desktop app
- 34bfef0b Set chat-message height to height of content in web, desktop

### DIFF
--- a/src/interface/desktop/chat.html
+++ b/src/interface/desktop/chat.html
@@ -675,28 +675,35 @@
             await loadChat();
         });
 
-        async function loadChat() {
-            // Load chat history
+        async function getChatHistoryUrl() {
             const hostURL = await window.hostURLAPI.getURL();
             const khojToken = await window.tokenAPI.getToken();
             const headers = { 'Authorization': `Bearer ${khojToken}` };
 
-            let firstRunSetupMessageRendered = false;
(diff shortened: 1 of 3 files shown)


src/interface/desktop/chat.html
```diff
@@ -675,28 +675,35 @@
...
             let chatHistoryUrl = `${hostURL}/api/chat/history?client=desktop`;
             if (chatBody.dataset.conversationId) {
                 chatHistoryUrl += `&conversation_id=${chatBody.dataset.conversationId}`;
             }
+            return { chatHistoryUrl, headers };
+        }
+
+        async function loadChat() {
+            // Load chat history and body
+            const hostURL = await window.hostURLAPI.getURL();
```
sabaimran, line 694: This can be nixed.

sabaimran (approved):

src/interface/desktop/chat.html
debanjum: `hostURL` is used in `loadChat`function futher down

PR #729, opened 2024-04-29
Make it easier to manage server-level chat settings

sabaimran:
- Configure a default model and a summarizer model using the `ServerChatSettings`
- Use the summarizer model when reading the results of a web page. This allows the admin to tweak responses depending on the use case of a function
- Fix a bug for how online results were being created for web page functions


githubloves1:
Make it easier to manage server-level chat settings

PR #734, opened 2024-05-04
Serve image assets from Khoj domain, not directly from S3 bucket

debanjum:
- f6ce9ff2 Serve generated images from Khoj domain instead of directly from S3
- 274b1e71 Rename assets URL from Khoj S3 bucket to assets.khoj.dev

### DIFF
--- a/src/khoj/database/management/commands/serve_generated_images_from_server.py
+++ b/src/khoj/database/management/commands/serve_generated_images_from_server.py
@@ -1,0 +1,60 @@
+from django.core.management.base import BaseCommand
+
+from khoj.database.models import Conversation
+from khoj.utils.helpers import ImageIntentType, is_none_or_empty
+
+
+class Command(BaseCommand):
+    help = "Serve Khoj generated images from a different URL."
+
+    def add_arguments(self, parser):
+        # Pass Source URL
+        parser.add_argument(
+            "--source",
+            action="store",
+            help="URL from which generated images are currently served.",
+        )
+        # Pass Destination URL
+        parser.add_argument("--destination", action="store", help="URL to serve generated image from going forward.")
+
+        # Add a new argument 'reverse' to the command
+        parser.add_argument(
+            "--reverse",
+            action="store_true",
+            help="Revert to serve generated images from source instead of destination URL.",
+        )
+
+    def handle(self, *args, **options):
+        updated_count = 0
+        if not options.get("source") or not options.get("destination"):
+            self.stdout.write(
+                self.style.ERROR(
+                    "AWS_IMAGE_UPLOAD_BUCKET environment variable or --source, --destination args needs to be set."
+                )
+            )
+            return
+
+        destination = options["source"] if options["reverse"] else options["destination"]
+        source = options["destination"] if options["reverse"] else options["source"]
+        for conversation in Conversation.objects.all():
+            conversation_updated = False
+            for chat in conversation.conversation_log.get("chat", []):
+                if (
+                    chat.get("by", "") == "khoj"
+                    and not is_none_or_empty(chat.get("message"))
+                    and chat.get("message", "").startswith(source)
+                    and chat.get("intent", {}).get("type", "") == ImageIntentType.TEXT_TO_IMAGE2.value
+                ):
+                    if chat.get("message", "").endswith(".webp"):
+                        # Convert source url to destination url
+                        chat["message"] = chat["message"].replace(source, destination)
+                        conversation_updated = True
+                        updated_count += 1
+
+            if conversation_updated:
+                print("Save the updated conversation")
+                conversation.save()
+
+        if updated_count > 0:
+            success = f"Successfully converted {updated_count} image URLs from {source} to {destination}.".strip()
+            self.stdout.write(self.style.SUCCESS(success))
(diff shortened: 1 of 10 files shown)


src/khoj/interface/web/chat.html
```diff
@@ -3,7 +3,7 @@
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0 maximum-scale=1.0">
-        <meta property="og:image" content="https://khoj-web-bucket.s3.amazonaws.com/khoj_hero.png">
+        <meta property="og:image" content="https://assets.khoj.dev/khoj_hero.png">
```
sabaimran, line 6: We have some other files that are being loaded through a CDN - `pico.min.css`, `intl-tel-input`, `intlTelInput.min.js` are some that come to mind. We should be able to put these in `assets.khoj.dev` as well. Let's just make sure that we find them all before merging this change.

src/khoj/database/management/commands/serve_generated_images_from_server.py
```diff
@@ -0,0 +1,60 @@
...
+        for conversation in Conversation.objects.all():
+            conversation_updated = False
+            for chat in conversation.conversation_log.get("chat", []):
+                if (
+                    chat.get("by", "") == "khoj"
+                    and not is_none_or_empty(chat.get("message"))
+                    and chat.get("message", "").startswith(source)
+                    and chat.get("intent", {}).get("type", "") == ImageIntentType.TEXT_TO_IMAGE2.value
+                ):
+                    if chat.get("message", "").endswith(".webp"):
```
sabaimran, line 48: nit: is it necessary to check for the file ending? And if so, it can also go in the conditional on top?

src/khoj/database/management/commands/serve_generated_images_from_server.py
```diff
@@ -0,0 +1,60 @@
...
+                        # Convert source url to destination url
+                        chat["message"] = chat["message"].replace(source, destination)
+                        conversation_updated = True
+                        updated_count += 1
+
+            if conversation_updated:
+                print("Save the updated conversation")
```
sabaimran, line 55: May be useful to output some metadata, like the URL changed, or the conversation ID?

src/khoj/database/management/commands/serve_generated_images_from_server.py
```diff
@@ -0,0 +1,60 @@
...
+                    "AWS_IMAGE_UPLOAD_BUCKET environment variable or --source, --destination args needs to be set."
+                )
+            )
+            return
+
+        destination = options["source"] if options["reverse"] else options["destination"]
+        source = options["destination"] if options["reverse"] else options["source"]
+        for conversation in Conversation.objects.all():
+            conversation_updated = False
+            for chat in conversation.conversation_log.get("chat", []):
```
sabaimran, line 41: Can you use tqdm to track the progress here? The job may take a while to run and it would be good to have a sense of the completion rate.

src/khoj/database/management/commands/serve_generated_images_from_server.py
```diff
@@ -0,0 +1,60 @@
...
+                        conversation_updated = True
+                        updated_count += 1
+
+            if conversation_updated:
+                print("Save the updated conversation")
+                conversation.save()
+
+        if updated_count > 0:
+            success = f"Successfully converted {updated_count} image URLs from {source} to {destination}.".strip()
+            self.stdout.write(self.style.SUCCESS(success))
```
sabaimran, line 60: Is `print` not sufficient for printing the output?

src/khoj/database/management/commands/serve_generated_images_from_server.py
```diff
@@ -0,0 +1,60 @@
+from django.core.management.base import BaseCommand
+
+from khoj.database.models import Conversation
+from khoj.utils.helpers import ImageIntentType, is_none_or_empty
+
+
+class Command(BaseCommand):
```
sabaimran, line 7: Perhaps the name of this file would be more apt? `change_generated_images_bucket_name`.

src/khoj/database/management/commands/serve_generated_images_from_server.py
```diff
@@ -0,0 +1,60 @@
...
+            action="store_true",
+            help="Revert to serve generated images from source instead of destination URL.",
+        )
+
+    def handle(self, *args, **options):
+        updated_count = 0
+        if not options.get("source") or not options.get("destination"):
+            self.stdout.write(
+                self.style.ERROR(
+                    "AWS_IMAGE_UPLOAD_BUCKET environment variable or --source, --destination args needs to be set."
```
sabaimran, line 32: the `AWS_IMAGE_UPLOAD_BUCKET` isn't being used in the script. Would this be more accurate?
```suggestion
                    "Set both --source, --destination args to respective URLs"
```

src/khoj/interface/web/chat.html
sabaimran: I've done this change for the katex files in https://github.com/khoj-ai/khoj/commit/9ae828cf11156896fc4e37db2ce9add558d260ed

debanjum: pico.min.css is already available locally in code, intl-tel-input files aren't strictly necessary to move to assets.khoj.dev but I'll move them at some point soon

sabaimran (approved):
Looks good!

PR #738, opened 2024-05-15
Update telemetry.md disabling instructions

joshavant:
I was just setting up a self-hosted Khoj and noticed that this detail was hidden in the Telemetry configuration instructions.


debanjum:
Thanks for clarifying that in the docs!

PR #760, opened 2024-05-26
Add support for Anthropic models

sabaimran:
# Incoming
Long-time coming, add first-party support for using Anthropic models inside the Khoj codebase. I went with a first-party approach (instead of leveraging an OpenAI-API compatible server, for instance) because the Anthropic models have slight nuances in the way messages are handled. For example, the first message to the model always has to be a user message. It is also less good at honoring the prompt output format when it's intended to be JSON. 

Closes #760


src/khoj/routers/helpers.py
```diff
@@ -542,8 +561,7 @@ def send_message_to_model_wrapper_sync(
         )
 
     elif conversation_config.model_type == "openai":
-        openai_chat_config = ConversationAdapters.get_openai_conversation_config()
-        api_key = openai_chat_config.api_key
+        api_key = conversation_config.openai_config.api_key
```
debanjum, line 564: Should we rename `openai_config` to something more generic, e.g `online_model`? 

```suggestion
        api_key = conversation_config.online_model.api_key
```

debanjum (approved):

PR #774, opened 2024-05-28
Improved Command Menu and Help Command

MythicalCow:
### Summary of Changes

* The command menu (triggered by "/") now has a clickable list of possible commands, that automatically fill into the chat when pressed.
* The `/help` command now searches `khoj.dev` pages to provide useful assistance to the user.

### Demo Images

Screenshot 2024-05-28 115417
Screenshot 2024-05-28 115440

### DIFF
--- a/src/khoj/processor/tools/online_search.py
+++ b/src/khoj/processor/tools/online_search.py
@@ -3,7 +3,7 @@ import json
 import logging
 import os
 from collections import defaultdict
-from typing import Callable, Dict, Optional, Tuple, Union
+from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import aiohttp
 import requests
@@ -47,9 +47,9 @@ async def search_online(
     conversation_history: dict,
     location: LocationData,
     send_status_func: Optional[Callable] = None,
-    custom_filters: str = None,
+    custom_filters: List[str] = [],
 ):
-    query += custom_filters
+    query += " ".join(custom_filters)
     if not online_search_enabled():
         logger.warn("SERPER_DEV_API_KEY is not set")
         return {}
--- a/src/khoj/processor/tools/online_search.py
+++ b/src/khoj/processor/tools/online_search.py
@@ -43,8 +43,13 @@ MAX_WEBPAGES_TO_READ = 1
 
 
 async def search_online(
-    query: str, conversation_history: dict, location: LocationData, send_status_func: Optional[Callable] = None
+    query: str,
+    conversation_history: dict,
+    location: LocationData,
+    send_status_func: Optional[Callable] = None,
+    custom_filters: str = None,
 ):
+    query += custom_filters
     if not online_search_enabled():
         logger.warn("SERPER_DEV_API_KEY is not set")
         return {}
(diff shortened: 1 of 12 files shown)


src/khoj/routers/api_chat.py
```diff
@@ -595,6 +586,16 @@ async def send_rate_limit_message(message: str):
...
 
+        if conversation_commands == [ConversationCommand.Help]:
+            conversation_config = await ConversationAdapters.aget_user_conversation_config(user)
+            if conversation_config == None:
+                conversation_config = await ConversationAdapters.aget_default_conversation_config()
+            model_type = conversation_config.model_type
+            formatted_help = help_message.format(model=model_type, version=state.khoj_version, device=get_device())
+            # Adding specification to search online specifically on khoj.dev pages.
+            defiltered_query += " site:khoj.dev"
+            conversation_commands.append(ConversationCommand.Online)
```
sabaimran, line 597: This is good! There's a slight variation in how I think we should handle this.
1. Rather than explicitly adding the qualifier `site:khoj.dev` to the query here, I think it would be better if we exposed a parameter in the `search_online` function for explicit filters. This will generalize better as we add additional commands and make the code more readable. 
2. I also think we should move this logic back to the top of the function in the event that the user has just run `/help` and nothing else. We want to exit early and save on compute costs if that's the case.

sabaimran (commented):
Tested it out locally and this is super neat! Great work. Just a few fixes and we should be good.

src/khoj/processor/tools/online_search.py
```diff
@@ -43,8 +43,13 @@
 
 
 async def search_online(
-    query: str, conversation_history: dict, location: LocationData, send_status_func: Optional[Callable] = None
+    query: str,
+    conversation_history: dict,
+    location: LocationData,
+    send_status_func: Optional[Callable] = None,
+    custom_filters: str = None,
```
sabaimran, line 50: nit: make this of type `List[str]` and append each like
```
query += ' '.join(explicit_filters)
```

src/khoj/routers/api_chat.py
```diff
@@ -528,6 +528,20 @@ async def send_rate_limit_message(message: str):
             await conversation_command_rate_limiter.update_and_check_if_valid(websocket, cmd)
             q = q.replace(f"/{cmd.value}", "").strip()
 
+        custom_filters = ""
```
sabaimran, line 531: Based on above comment, would want to change to something like 
`custom_filters = []`

and then later, 

`custom_filters.append("site:khoj.dev")`.

src/khoj/routers/api_chat.py
```diff
@@ -752,13 +758,18 @@ async def chat(
...
 
+    _custom_filters = ""
     if conversation_commands == [ConversationCommand.Help]:
-        conversation_config = await ConversationAdapters.aget_user_conversation_config(user)
-        if conversation_config == None:
-            conversation_config = await ConversationAdapters.aget_default_conversation_config()
-        model_type = conversation_config.model_type
-        formatted_help = help_message.format(model=model_type, version=state.khoj_version, device=get_device())
-        return StreamingResponse(iter([formatted_help]), media_type="text/event-stream", status_code=200)
+        if not q:
```
sabaimran, line 763: Would this work? Wouldn't `q` at least contain `/help` at this point? You can test it by disable the websocket in the code in the client.

sabaimran (commented):
Awesome, thanks for making the adjustments! Just a few more fixes and it should be good.

sabaimran (approved):

PR #797, opened 2024-06-04
Fix multilingual font rendering

### LINKED ISSUE #756: Suggestion for Persian Font Improvement
Hello,

Your work is outstanding. I have used the web version of Khoj and I would like to suggest changing the Persian font. The current font is not very legible, and there are many issues with the letters being improperly placed next to each other in words. Thank you for all the hard work you put into this. Wishing you success.

sabaimran:
- Fallback to an Arabic language font which contains more Asian language characters. Closes #756

Screenshot 2024-06-04 at 14 07 24


sabaimran:
Showing before and after:
| Before (Tajawal) | After (Noto Sans - Arabic)|
|-|-|
| Screenshot 2024-06-06 at 12 10 39|Screenshot 2024-06-06 at 12 12 23|

debanjum (approved):
Made some font-size and spacing fixes but otherwise looks good! We can extend Noto to support CJK languages if any user's request it later as well

PR #799, opened 2024-06-06
Improve Intermediate Steps in Formulating Chat Response

debanjum:
## Major
- f91cdf8e Fix showing headings in intermediate step in generating chat response
- 18f7e6e7 Remove "Path" prefix from org ancestor heading in compiled entry
- dd2225b1 Use Text output mode to disambiguate from Default data source lookup
  Previously if default output was selected by Khoj, we'd end up doing
  a documents search as well, even when Khoj selected internet or
  general data source to lookup.
  
  This update disambiguates the default information mode from the text
  output mode. To avoid doing documents search when not deemed necessary
  by Khoj

## Minor
- Fix openai chat actor, director unit tests

### DIFF
--- a/src/khoj/routers/api_chat.py
+++ b/src/khoj/routers/api_chat.py
@@ -590,9 +590,7 @@ async def websocket_endpoint(
         )
 
         if compiled_references:
-            headings = "\n- " + "\n- ".join(
-                set([" ".join(c.get("compiled", c).split("Path: ")[1:]).split("\n ")[0] for c in compiled_references])
-            )
+            headings = "\n- " + "\n- ".join(set([c.get("compiled", c).split("\n")[0] for c in compiled_references]))
             await send_status_update(f"**📜 Found Relevant Notes**: {headings}")
 
         online_results: Dict = dict()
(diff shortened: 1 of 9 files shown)


src/khoj/routers/api_chat.py
```diff
@@ -590,9 +590,7 @@ async def send_rate_limit_message(message: str):
         )
 
         if compiled_references:
-            headings = "\n- " + "\n- ".join(
-                set([" ".join(c.get("compiled", c).split("Path: ")[1:]).split("\n ")[0] for c in compiled_references])
-            )
+            headings = "\n- " + "\n- ".join(set([c.get("compiled", c).split("\n")[0] for c in compiled_references]))
```
sabaimran, line 593: Why don't we add the `filename` as a property in the reference and use that instead? In `api.py`, when `compiled_references` is being formed just before the return statement.

sabaimran (commented):
nice fix, that bug was pretty annoying.

src/khoj/routers/api_chat.py
debanjum: We're using the first line of the heading because it also contains the heading ancestry of the note, at least when org-mode files are referenced. This granularity would be useful in long org-mode files with lots of notes.

Issue #787, opened 2024-05-31
[IDEA] Add better support for document summarization


sabaimran:
## Describe the feature you'd like
When a document is uploaded to Khoj, we chunk the file and create `Entry`s in the database to represent parts of the file.

This creates a limitation, because we don't have the full file saved to the database. If a user requests any kind of summarization option, we aren't able to fulfill that request. 

To fix this, store the raw file along with the `Entry` objects. We should add a new data model for the `File` object that stores the raw text of the file. It should look something like this:

```
class FileObject(BaseModel):
    file_name = models.CharField...
    raw_text = models.Textfield...
```

And update the `Entry` object like so:
```
class Entry(BaseModel):
...
    file_object = models.ForeignKey(FileObject,...)
...
```

We might run into a size constraint on really large files. In that case, we would want to throw an error. Eventually, we can upload the file to S3 instead of throwing an error.

When a file with an existing filename for the user lands for processing, check the hashes of the current file and the incoming file and see if they're different. If so, process the new file and update the entries as usual, updating the `FileObject` with the new reference.

You'll also want to add a conversation command called `summarize` which should lookup the file names that are accessible to the given user and find the best match for whatever the user is requesting. It should get the whole text and send it to the LLM for response.

Relevant files:
- [markdown_to_entries.py](https://github.com/khoj-ai/khoj/blob/features/shareable-automations/src/khoj/processor/content/markdown/markdown_to_entries.py) for processing each of the uploaded markdown files. Each of the file types or data sources we support has its own such file (i.e., `pdf_to_entries`)
- [embeddings.py](https://github.com/khoj-ai/khoj/blob/features/shareable-automations/src/khoj/processor/embeddings.py) for computing embeddings of the relevant chunks
- [indexer.py](https://github.com/khoj-ai/khoj/blob/features/shareable-automations/src/khoj/routers/indexer.py) for the API endpoints that read in a new file
- [models.__init__.py](https://github.com/khoj-ai/khoj/blob/features/shareable-automations/src/khoj/database/models/__init__.py#L297-L325) for the database model that describes `Entry` objects

Issue #509, opened 2023-10-19
Add better support for meta-level/summarization questions


sabaimran:
One limitation of Khoj currently is that it's not very good at answering meta-level question. For example, `"How many notes do I have about Spanish lessons?"` or "`What was the first issue related to offline models in the Khoj repository?"` won't be very effective. More categorical, specific questions however do perform well.

This may need re-thinking of how we index data. Now that we're introducing a Postgres backend (refer to efforts in the PRs from #487 and others tagged with [Multi-User]), we could add an extra layer for the LLM to run any subqueries that would be necessary for executing this request.

For example, `"How many notes do I have about Spanish lessons?"` can be converted to something like `SELECT * FROM database_embeddings WHERE distance <0.1`, where `distance` is computed between the vectorized column and the query string.

I do think this would generally be useful for any kind of meta-analysis.

See [relevant discussion in Discord](https://discord.com/channels/1112065956647284756/1112066447947083826/1164599027506106400).

FetchFast:
Or would a memgpt style approach of memory be a better way to go? (see memgpt github code)

PR #788, opened 2024-05-31
New Feature: Adding File Filtering to Conversations

MythicalCow:
* Interactive file menu  #UI to add/remove files on each conversation as references.
* Backend changes implemented to load selected file filters from a conversation into the querying process.

Screenshot 2024-05-30 174005
Screenshot 2024-05-30 173917

PR #800, opened 2024-06-06
Add Ability to Summarize Documents

MythicalCow:
* Uses entire file text and summarizer model to generate document summary.
* Uses the contents of the user's query to create a tailored summary.
* Integrates with File Filters #788 for a better UX.

image


src/khoj/routers/api_chat.py
```diff
@@ -532,6 +591,27 @@ async def send_rate_limit_message(message: str):
...
+                try:
+                    file_object = await FileObjectAdapters.async_get_file_objects_by_name(file_filters[0])
+                    contextual_data = " ".join([file.raw_text for file in file_object])
+                    if not q:
+                        q = "Create a general summary of the file"
+                    await send_status_update(f"**🧑🏾‍💻 Constructing Summary Using:** {file_object[0].file_name}")
+                    response = await extract_relevant_summary(q, contextual_data)
+                    await send_complete_llm_response(str(response))
+                except Exception as e:
+                    await send_complete_llm_response(f"Error summarizing file.")
```
sabaimran, line 612: Add a log line here to log the exception, like
```suggestion
            logger.error(f"Error summarizing file for {user.email}: {e}")

```

src/khoj/database/models/__init__.py
```diff
@@ -325,6 +326,12 @@ class EntrySource(models.TextChoices):
     corpus_id = models.UUIDField(default=uuid.uuid4, editable=False)
 
 
+class FileObject(BaseModel):
```
sabaimran, line 329: Let's store the `user` on the `FileObject`.

```suggestion
class FileObject(BaseModel):
    # Same as Entry but raw will be a much larger string
    file_name = models.CharField(max_length=400, default=None, null=True, blank=True)
    raw_text = models.TextField()
    user = models.ForeignKey(KhojUser, on_delete=models.CASCADE)
```

src/khoj/routers/api_chat.py
```diff
@@ -532,6 +591,27 @@ async def send_rate_limit_message(message: str):
...
+            file_filters = conversation.file_filters
+            if len(file_filters) == 0:
+                await send_complete_llm_response(
+                    "No files selected for summarization. Please add files using the section on the left."
+                )
+            elif len(file_filters) > 1:
+                await send_complete_llm_response("Only one file can be selected for summarization.")
+            else:
+                try:
+                    file_object = await FileObjectAdapters.async_get_file_objects_by_name(file_filters[0])
```
sabaimran, line 604: Be sure to add support for this in the non-websocket chat API as well!

sabaimran (commented):
Sweet, excited to roll this out soon!

src/khoj/database/adapters/__init__.py
```diff
@@ -846,8 +847,59 @@ async def aget_text_to_image_model_config():
...
+    @staticmethod
+    async def async_overwrite_raw_text(file_object: FileObject, new_raw_text: str):
+        await sync_to_async(lambda: setattr(file_object, "raw_text", new_raw_text))()
+        await sync_to_async(file_object.save)()
+
+    @staticmethod
+    async def async_create_file_object(file_name: str, raw_text: str):
+        return await sync_to_async(FileObject.objects.create)(file_name=file_name, raw_text=raw_text)
+
+    @staticmethod
```
sabaimran, line 884: You can simplify this syntax a little bit. Same with most of the other async methods here, there's usually a corresponding method in the Django ORM that you can use with the `a` prefix that acts as a drop-in replacement. Like `first` -> `afirst`, `create` -> `acreate`, etc.

```suggestion
    @staticmethod
    async def async_create_file_object(file_name: str, raw_text: str):
        return await FileObject.objects.acreate(file=file_name, raw_text=raw_text)

```

src/khoj/database/adapters/__init__.py
```diff
@@ -846,8 +847,59 @@ async def aget_text_to_image_model_config():
...
+        return await sync_to_async(FileObject.objects.filter)(file_name=file_name).delete()
+
+    @staticmethod
+    async def async_delete_all_file_objects():
+        return await sync_to_async(FileObject.objects.all)().delete()
+
+
 class EntryAdapters:
-    word_filer = WordFilter()
+    word_filer = WordFilter()  # typo here should be word_filter
```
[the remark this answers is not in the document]
MythicalCow: it says word_filer instead of word_filter. I was going to change it but didn't get around to it. Let's address this separately.

src/khoj/routers/api_chat.py
```diff
@@ -856,6 +857,30 @@ async def chat(
...
+            llm_response = "Only one file can be selected for summarization."
+            return StreamingResponse(content=llm_response, media_type="text/event-stream", status_code=200)
+        else:
+            try:
+                file_object = await FileObjectAdapters.async_get_file_objects_by_name(user, file_filters[0])
+                contextual_data = " ".join([file.raw_text for file in file_object])
+                summarizeStr = "/" + ConversationCommand.Summarize
+                if q.strip() == summarizeStr:
+                    q = "Create a general summary of the file"
+                response = await extract_relevant_summary(q, contextual_data)
```
sabaimran, line 876: As is, your output won't stream iteratively. It'll wait until the whole response is generated and then send it in one go. Contrast in the code how `agenerate_chat_response` is used. It's piping the llm response as it comes to the client. 

This is less than ideal, especially because the summarizer can take a bit longer than other queries. But it would require too much refactoring at the moment to get it working, so it's out of scope for this PR. Just wanted to note this so we can improve it later.

src/khoj/database/migrations/0045_fileobject_user.py
sabaimran, line 1: Since this is all going in in one review, combine both of the migrations. The best way to do this is to locally rollback your database to 0043 (`python3 src/khoj/manage.py migrate database 0043`) and then delete the new migration files (0044, 0045). Then run `python3 src/khoj/manage.py makemigrations` again, and only one file should be generated.

src/khoj/database/adapters/__init__.py
```diff
@@ -731,7 +732,7 @@ async def aget_summarizer_conversation_config():
         if server_chat_settings is None or (
             server_chat_settings.summarizer_model is None and server_chat_settings.default_model is None
         ):
-            return await ChatModelOptions.objects.filter().afirst()
+            return await ChatModelOptions.objects.filter().prefetch_related("openai_config").afirst()
```
sabaimran, line 735: 🚀 Closes #809

src/khoj/database/adapters/__init__.py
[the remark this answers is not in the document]
sabaimran: Oh damn, I still didn't see it after reading the comment lol. Sounds good.

src/khoj/database/migrations/0044_conversation_file_filters.py
```diff
@@ -1,17 +0,0 @@
-# Generated by Django 4.2.10 on 2024-05-29 19:56
-
-from django.db import migrations, models
```
sabaimran, line 3: We shouldn't overwrite this migration, since this migration has already been merged and deployed. We should just add a new one `45` on top of it. I'll fix this.

sabaimran: To resolve any issues with your local setup afterwards, run this in a psql shell for the khoj db:
```
alter table database_conversation drop column file_filters;
drop table database_fileobject;
```

src/khoj/routers/api_chat.py
```diff
@@ -583,6 +586,41 @@ async def send_rate_limit_message(message: str):
             await conversation_command_rate_limiter.update_and_check_if_valid(websocket, cmd)
             q = q.replace(f"/{cmd.value}", "").strip()
 
+        if conversation_commands == [ConversationCommand.Summarize]:
```
sabaimran, line 589: We should also add this to the tool selection process so it can be dynamically inferred.

src/khoj/routers/api_chat.py
```diff
@@ -583,6 +586,41 @@ async def send_rate_limit_message(message: str):
...
+            response_log = ""
+            if len(file_filters) == 0:
+                response_log = "No files selected for summarization. Please add files using the section on the left."
+                await send_complete_llm_response(response_log)
+            elif len(file_filters) > 1:
+                response_log = "Only one file can be selected for summarization."
+                await send_complete_llm_response(response_log)
+            else:
+                try:
+                    file_object = await FileObjectAdapters.async_get_file_objects_by_name(user, file_filters[0])
```
sabaimran, line 600: Need to add error handling here in case the file_object isn't there. This would be the case for any file that's been previously uploaded.

src/khoj/routers/api_chat.py
```diff
@@ -583,6 +586,45 @@ async def send_rate_limit_message(message: str):
             await conversation_command_rate_limiter.update_and_check_if_valid(websocket, cmd)
             q = q.replace(f"/{cmd.value}", "").strip()
 
+        if ConversationCommand.Summarize in conversation_commands:
```
sabaimran, line 589: Can you add some unit tests for this in `test_openai_chat_actors.py` and ideally `test_offline_chat_actors.py`? If you can't get offline chat working, I can run these locally.

tests/test_openai_chat_director.py
```diff
@@ -289,6 +289,137 @@ def test_answer_not_known_using_notes_command(chat_client_no_background, default
...
+        .values_list("file_path", flat=True)
+    )
+    # pick the file that has "Xi Li.markdown" in the name
+    summarization_file = ""
+    for file in file_list:
+        if "Birthday Gift for Xiu turning 4.markdown" in file:
+            summarization_file = file
+            break
+    assert summarization_file != ""
+
```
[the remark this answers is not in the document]
MythicalCow: So this portion of the code is to get the proper path because this will be slightly different for a user testing locally. There might be a better way to do this but this was the simplest way for me to get the path containing the appropriate file.

tests/test_openai_chat_director.py
```diff
@@ -289,6 +289,137 @@ def test_answer_not_known_using_notes_command(chat_client_no_background, default
     assert response_message == prompts.no_entries_found.format()
 
 
+# ----------------------------------------------------------------------------------------------------
+@pytest.mark.django_db(transaction=True)
+@pytest.mark.chatquality
+def test_summarize_one_file(chat_client, default_user2: KhojUser):
+    message_list = []
+    conversation = create_conversation(message_list, default_user2)
```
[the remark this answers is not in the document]
MythicalCow: Yeah, that makes sense from a testing architecture standpoint. I will think about this in the future!

tests/test_openai_chat_director.py
```diff
@@ -289,6 +289,137 @@ def test_answer_not_known_using_notes_command(chat_client_no_background, default
...
+    # post "imaginary.markdown" file to the file filters
+    response = chat_client.post(
+        "api/chat/conversation/file-filters",
+        json={"filename": "imaginary.markdown", "conversation_id": str(conversation.id)},
+    )
+    query = urllib.parse.quote("/summarize")
+    response = chat_client.get(f"/api/chat?q={query}&conversation_id={conversation.id}&stream=true")
+    response_message = response.content.decode("utf-8")
+    # Assert
+    assert response_message == "No files selected for summarization. Please add files using the section on the left."
```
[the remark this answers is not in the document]
MythicalCow: Sounds good!

sabaimran:
Closes #509 
Closes #787

PR #802, opened 2024-06-07
Add language-specific syntax highlighting via highlight.js

### LINKED ISSUE #697: Improve rendering for code blocks in the web UI
Can we use a 3rd party library to show syntax highlighting for different languages?

sabaimran:
- Add highlight.js to our assets CDN for fast load and compliance with the CSP
- See other stylesheet options here: https://cdnjs.com/libraries/highlight.js

See https://github.com/highlightjs/highlight.js ✨

Closes #697


src/interface/desktop/chat.html
```diff
@@ -10,6 +10,8 @@
     </head>
     <script type="text/javascript" src="./assets/purify.min.js?v={{ khoj_version }}"></script>
     <script type="text/javascript" src="./assets/markdown-it.min.js"></script>
+    <link rel="stylesheet" href="https://assets.khoj.dev/higlightjs/solarized-dark.min.css">
```
debanjum, line 13: Why is solarized dark css required here instead of just solarized?

debanjum (approved):

src/interface/desktop/chat.html
sabaimran: I like solarized dark 😬

sabaimran: I'm going to merge this for now because public conversation references are broken, but we can come back to the stylesheet later. It's simple to update.

PR #803, opened 2024-06-07
Fix: Adding Support for Uploading Multiple Files

MythicalCow:
* Adding support for uploading multiple files at a time.


src/khoj/interface/web/chat.html
```diff
@@ -801,17 +801,20 @@
...
 
             const fileInput = document.createElement('input');
             fileInput.type = 'file';
+            fileInput.multiple = true;
             fileInput.addEventListener('change', function() {
-                const selectedFile = fileInput.files[0];
-                uploadDataForIndexing(selectedFile);
+                const selectedFiles = fileInput.files;
+                for (let i = 0; i < selectedFiles.length; i++) {
+                    uploadDataForIndexing(selectedFiles[i]);
```
sabaimran, line 816: Rather than sending each of the files individually for indexing, can you create a batch request? You should be able to do it by modifying `formData` in the `uploadDataForIndexing` method with data for each of the files. Process each file, and add it to `formData` like this, before passing `formData` to the indexing API. 

```
for (let file of files) {
    reader.onload = function (event) {
        fileContents = event.target.result;
        let fileType = file.type;
        if (fileType === "") {
            if (fileName.split('.').pop() === "org") {
                fileType = "text/org";
            } else if (fileName.split('.').pop() === "md") {
                fileType = "text/markdown";
            } else if (fileName.split('.').pop() === "txt") {
                fileType = "text/plain";
            } else if (fileName.split('.').pop() === "html") {
                fileType = "text/html";
            } else if (fileName.split('.').pop() === "pdf") {
                fileType = "application/pdf";
            }
        }

        let fileObj = new Blob([fileContents], { type: fileType });
        formData.append("files", fileObj, file.name);
    };
    reader.readAsText(file);
}
```

MythicalCow: Will fix this. Thanks!

src/khoj/interface/web/chat.html
```diff
@@ -838,28 +836,24 @@
...
-                    alert("Sorry, that file type is not yet supported");
-                }
-                var overlayText = document.getElementById("dropzone-overlay");
-                if (overlayText != null) {
-                    overlayText.remove();
+            for (let file of files) {
+                if (!file || (!allowedExtensions.includes(file.type) && !allowedFileEndings.includes(file.name.split('.').pop()))) {
+                    dropzone.classList.remove('dragover');
+                    if (file) {
+                        alert("Sorry, that file type is not yet supported");
```
sabaimran, line 846: This would cause the entire upload to be skipped if there's even one unsupported file. Maybe change this so that all the problematic files are aggregated into a list and then included in the alert message, so the user can remove them and try again? Or we can just skip the problematic files, and provide an alert saying which files are going to be skipped. Your call.

src/khoj/interface/web/chat.html
```diff
@@ -869,59 +863,78 @@
...
+                        }
 
-                let fileObj = new Blob([fileContents], { type: fileType });
-                formData.append("files", fileObj, file.name);
-                console.log(formData);
+                        let fileObj = new Blob([fileContents], { type: fileType });
+                        formData.append("files", fileObj, file.name);
+                        resolve();
+                    };
+                    reader.onerror = reject;
```
sabaimran, line 897: What would happen if one of the read operations fails?

src/khoj/interface/web/chat.html
```diff
@@ -869,59 +863,78 @@
...
+                })
+                .then((data) => {
+                    console.log(data);
+                    dropzone.classList.remove('dragover');
+                    var overlayText = document.getElementById("dropzone-overlay");
+                    if (overlayText != null) {
+                        overlayText.remove();
+                    }
+                    // Display indexing success message
+                    flashStatusInChatInput("✅ File indexed successfully");
```
sabaimran, line 918: Awesome! And this should still work the same for the single file upload scenario, right?

sabaimran (commented):
Just a comment! Looks good, exciting to see this supported.

sabaimran (approved):
🚀🚀🚀

PR #804, opened 2024-06-07
UX Improvement: Keyboard Shortcuts for Recent Messages

MythicalCow:
* added support for using up and down arrows to place recent messages into chat input.
* individual recent message buffers based on the current conversation.


src/khoj/interface/web/chat.html
```diff
@@ -2170,6 +2214,24 @@
                             </svg>
                         </button>
                         <textarea id="chat-input" class="option" oninput="onChatInput()" onkeydown=incrementalChat(event) autofocus="autofocus" placeholder="Type / to see a list of commands"></textarea>
+                        <!-- Shortcut Handler for Accessing Old Messages -->
+                        <script>
```
debanjum, line 2218: Should we put this js block in an existing `<script>` block?

src/khoj/interface/web/chat.html
```diff
@@ -2170,6 +2214,24 @@
...
+                            document.addEventListener('keydown', function(event) {
+                                const chatInput = document.getElementById('chat-input');
+                                if (document.activeElement === chatInput) {
```
debanjum, line 2221: Can this be simplified to something like below:
```suggestion
                            let chatInput = document.getElementById('chat-input');
                            chatInput.addEventListener('keydown', function(event) {
```

src/khoj/interface/web/chat.html
```diff
@@ -2039,11 +2060,33 @@
...
         }
+
+        function inputAutoFiller(key){
+            var chatInput = document.getElementById("chat-input");
+            if (key === "up") {
+                if (userMessageIndex > 0) {
+                    userMessageIndex -= 1;
+                    chatInput.value = userMessages[userMessageIndex];
+                } else {
+                    chatInput.value = "";
```
debanjum, line 2071: Should we cycle back to the most recent message here?
```suggestion
                    chatInputIndex = userMessages.length;
                    chatInput.value = "";
```

debanjum (commented):
Exciting stuff! Left some suggestions

src/khoj/interface/web/chat.html
```diff
@@ -2170,6 +2214,24 @@
                             </svg>
                         </button>
                         <textarea id="chat-input" class="option" oninput="onChatInput()" onkeydown=incrementalChat(event) autofocus="autofocus" placeholder="Type / to see a list of commands"></textarea>
+                        <!-- Shortcut Handler for Accessing Old Messages -->
+                        <script>
+                            document.addEventListener('keydown', function(event) {
+                                const chatInput = document.getElementById('chat-input');
+                                if (document.activeElement === chatInput) {
+                                    if (event.key === 'ArrowUp') {
```
debanjum, line 2222: I think we should trigger showing the chat history only when the chat input is empty. Otherwise it'll conflict with scrolling up in a multi-line chat input

src/khoj/interface/web/chat.html
```diff
@@ -2039,11 +2060,33 @@
                 console.error("Error loading file filters:", error);
             });
         }
+
+        function inputAutoFiller(key){
+            var chatInput = document.getElementById("chat-input");
```
debanjum, line 2065: Should we also show a dropdown menu with the chat history here? Similar to the dropdown menu UX of the slash commands?

debanjum (commented):
Tested the changes out. It feels super useful! Left a few ideas/suggestions

src/khoj/interface/web/chat.html
```diff
@@ -2170,6 +2214,24 @@
...
                         </button>
                         <textarea id="chat-input" class="option" oninput="onChatInput()" onkeydown=incrementalChat(event) autofocus="autofocus" placeholder="Type / to see a list of commands"></textarea>
+                        <!-- Shortcut Handler for Accessing Old Messages -->
+                        <script>
+                            document.addEventListener('keydown', function(event) {
+                                const chatInput = document.getElementById('chat-input');
+                                if (document.activeElement === chatInput) {
+                                    if (event.key === 'ArrowUp') {
+                                        inputAutoFiller('up');
+                                    } else if (event.key === 'ArrowDown') {
```
debanjum, line 2224: Should we show a dropdown menu for this with the chat history, similar to how the slash commands UX works?

src/khoj/interface/web/chat.html
MythicalCow: The only problem is that once the first chat history message is loaded, the chat input won't be empty anymore.

src/khoj/interface/web/chat.html
MythicalCow: Possibly. I don't know if it makes sense to show so many messages at a time.

src/khoj/interface/web/chat.html
MythicalCow: I didn't want to do this as it becomes confusing for the user to determine if they are at the start/end of the list. In the terminal, I prefer to hit a halt because otherwise I get lost in where I am on the list.

debanjum: Reasonable point. I was thinking it'd work fine if we use the dropdown to cycle through the recent messages, as the dropdown would provide the visual feedback on where you are in the list?

src/khoj/interface/web/chat.html
debanjum: Oh yeah good point!

Maybe we can use a different keybinding or use buttons for this or keep the up/down arrow bound to cycling through recent messages when an item from chat history was selected (vs scrolling when text was input via keyboard)?

Also will the current behaviour work off selecting recent messages work on phones (given they use touch/scroll events instead of arrow keys)?

PR #821, opened 2024-06-18
Add support for text to speech in chat responses

sabaimran:
- Integrate [Eleven Labs](https://elevenlabs.io/) for streaming speech responses for chat messages
- It currently only works once the entire audio message has been streamed to the client, which can make the wait time pretty long. Compensate for this by adding a loading symbol while waiting for the stream.


src/khoj/processor/speech/text_to_speech.py
```diff
@@ -0,0 +1,51 @@
...
+import os
+
+import requests  # Used for making HTTP requests
+from bs4 import BeautifulSoup
+from markdown_it import MarkdownIt
+
+# Define constants for the script
+CHUNK_SIZE = 1024  # Size of chunks to read/write at a time
+XI_API_KEY = os.getenv("ELEVEN_LABS_API_KEY", None)  # Your API key for authentication
+VOICE_ID = "29vD33N1CtxCmqQRPOHJ"  # ID of the voice model to use. MALE.
```
sabaimran, line 11: We could expose a list of `VOICE_ID`s that the user can configure in their settings, similar to how we configure chat models (eventually).

src/khoj/processor/speech/text_to_speech.py
```diff
@@ -0,0 +1,51 @@
+import json  # Used for working with JSON data
+import os
+
+import requests  # Used for making HTTP requests
+from bs4 import BeautifulSoup
+from markdown_it import MarkdownIt
+
+# Define constants for the script
+CHUNK_SIZE = 1024  # Size of chunks to read/write at a time
+XI_API_KEY = os.getenv("ELEVEN_LABS_API_KEY", None)  # Your API key for authentication
```
debanjum, line 10: Variable name maybe confusion. Maybe just call it ELEVEN_API_KEY (similar to ELEVEN_API_URL) or rename ELEVEN_API_URL to XI_API_URL

src/khoj/processor/speech/text_to_speech.py
debanjum, line 11: Is there a name for this voice model? maybe useful to add to comment if so

src/khoj/interface/web/chat.html
debanjum, line 1: Should we automatically trigger a spoken response when user asks question via speech? This will make interacting with Khoj when mobile a lot easier. 

Example scenario(s): You're walking/on_public_transport with headphones, you press the speech to text button to ask question, Khoj responds with speech, you don't need to look at your screen

debanjum (approved):
Looks good! This will make it easier to interact with Khoj when mobile and share Khoj's responses with folks in physical proximity

src/khoj/interface/web/chat.html
sabaimran: I really like this idea, but I think this should be a separate view altogether, and would require a separate spec/design. Can discuss offline.

PR #823, opened 2024-06-18
Support Indexing Images via OCR

MythicalCow:
* Added support for uploading `.jpeg`, `.jpg`, and `.png` files to Khoj.
* Updating indexer to generate raw text and entries using RapidOCR


src/khoj/interface/web/chat.html
```diff
@@ -48,8 +48,8 @@
 
 To get started, just start typing below. You can also type / to see a list of commands.
 `.trim()
-        const allowedExtensions = ['text/org', 'text/markdown', 'text/plain', 'text/html', 'application/pdf'];
-        const allowedFileEndings = ['org', 'md', 'txt', 'html', 'pdf'];
+        const allowedExtensions = ['text/org', 'text/markdown', 'text/plain', 'text/html', 'application/pdf', 'image/jpeg', 'image/png'];
+        const allowedFileEndings = ['org', 'md', 'txt', 'html', 'pdf', 'jpg', 'jpeg', 'png'];
```
debanjum, line 52: Let's add jpegs, pngs to the (valid) `binaryFileTypes` on the desktop app's `main.js`? This will allow folks to sync images containing text from their desktop app

src/khoj/processor/content/images/image_to_entries.py
```diff
@@ -0,0 +1,124 @@
...
+        # Extract required fields from config
+        if not full_corpus:
+            deletion_file_names = set([file for file in files if files[file] == b""])
+            files_to_process = set(files) - deletion_file_names
+            files = {file: files[file] for file in files_to_process}
+        else:
+            deletion_file_names = None
+
+        # Extract Entries from specified Pdf files
+        with timer("Extract entries from specified PDF files", logger):
```
debanjum, line 35: nit: s/PDF/image
```suggestion
        # Extract Entries from specified image files
        with timer("Extract entries from specified Image files", logger):
```

src/khoj/processor/content/images/image_to_entries.py
```diff
@@ -0,0 +1,124 @@
...
+                        continue
+                except ImportError:
+                    logger.warning(f"Unable to process file: {image_file}. This file will not be indexed.")
+                    continue
+                entry_to_location_map += zip(
+                    image_entries_per_file, [image_file] * len(image_entries_per_file)
+                )  # this is an indexed map of image_entries for the pdf.
+                entries.extend(image_entries_per_file)
+                file_to_text_map[image_file] = image_entries_per_file
+            except Exception as e:
```
debanjum, line 96: Are so many layers of try/catch required? Do we expect the code running in the outermost try/catch to fail under normal operating conditions? If not, we should continue to raise exceptions for those to make unexpected failures in this code path apparent

debanjum (changes requested):
Left some comments. Can't wait to index all my handwritten notes! 😃

src/khoj/processor/content/images/image_to_entries.py
```diff
@@ -0,0 +1,124 @@
...
+                            image_entries_per_file = [text[1] for text in result]
+                    except ImportError:
+                        logger.warning(f"Unable to process file: {image_file}. This file will not be indexed.")
+                        continue
+                except ImportError:
+                    logger.warning(f"Unable to process file: {image_file}. This file will not be indexed.")
+                    continue
+                entry_to_location_map += zip(
+                    image_entries_per_file, [image_file] * len(image_entries_per_file)
+                )  # this is an indexed map of image_entries for the pdf.
```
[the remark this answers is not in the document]
MythicalCow: typo while porting over from pdf_to_entries. Nice catch!

src/khoj/processor/content/images/image_to_entries.py
MythicalCow: good point! fixed.

tests/test_image_to_entries.py
```diff
@@ -0,0 +1,21 @@
...
+    assert len(entries[1]) == 31
+
+
+def test_jpg_to_jsonl():
+    with open("tests/data/images/nasdaq.jpg", "rb") as f:
+        image_bytes = f.read()
+    data = {"tests/data/images/nasdaq.jpg": image_bytes}
+    entries = ImageToEntries.extract_image_entries(image_files=data)
+    assert len(entries) == 2
+    assert len(entries[1]) == 143
```
debanjum, line 21: Why are there so many `entries[1]`? Is this counting the length of the OCR-ed string?

debanjum (approved):

tests/test_image_to_entries.py
MythicalCow: Those are the number of entries generated from the images. Perhaps we can condense it down a little bit

debanjum: Oh why so many entries? Does the test image have a lot of text?

MythicalCow: not quite. It could just be the way rapidocr splits the text

MythicalCow: I can probably chunk it together or something.

debanjum: Yeah joining the string may make sense if it seems like RapidOCR is splitting the text in a non-sensical way

sabaimran: Addressed this by taking it all as one blurb and joining the individual chunks with spaces. Will update corresponding unit tests.

sabaimran (approved):

PR #830, opened 2024-06-20
Enable using Stable Diffusion 3 for Image Generation via API

debanjum:
- Support Stable Diffusion 3 via API.
  Server Admin needs to setup model similar to DALLE-3 via Django Admin Panel
- Use shorter prompt generator to prompt SD3 to create better images
- Allow users to set paint model to use from web client config page


src/khoj/database/models/__init__.py
```diff
@@ -263,6 +265,11 @@ class UserSearchModelConfig(BaseModel):
     setting = models.ForeignKey(SearchModelConfig, on_delete=models.CASCADE)
 
 
+class UserPaintModelConfig(BaseModel):
```
sabaimran, line 268: Let's call this `UserImageModelConfig` for clarity? It'll make the relationship between this object and `TextToImageModelConfig` much more obvious.

```suggestion
class UserImageModelConfig(BaseModel):
```

src/khoj/database/models/__init__.py
```diff
@@ -234,9 +234,11 @@ class ModelType(models.TextChoices):
 class TextToImageModelConfig(BaseModel):
     class ModelType(models.TextChoices):
         OPENAI = "openai"
+        STABILITYAI = "stability-ai"
 
     model_name = models.CharField(max_length=200, default="dall-e-3")
     model_type = models.CharField(max_length=200, choices=ModelType.choices, default=ModelType.OPENAI)
+    api_key = models.CharField(max_length=200, default=None, null=True, blank=True)
```
sabaimran, line 241: Field is not being used for OpenAI models.

src/khoj/routers/helpers.py
```diff
@@ -747,74 +762,100 @@ async def text_to_image(
...
+            model_type=text_to_image_config.model_type,
+        )
+
+    if send_status_func:
+        await send_status_func(f"**🖼️ Painting using Enhanced Prompt**:\n{improved_image_prompt}")
+
+    if text_to_image_config.model_type == TextToImageModelConfig.ModelType.OPENAI:
+        with timer("Generate image with OpenAI", logger):
+            try:
                 response = state.openai_client.images.generate(
```
sabaimran, line 800: Credentials for the OpenAI flow are setup in `configure.py` as such:
```
state.openai_client = openai.OpenAI(api_key=openai_config.api_key)
```

This is a little bit confusing and un-intuitive as a developer, IMO. At the least, we should also try and use the `api_key` field if it is an OpenAI model OR make the `TextToImage` model take an optional association with an `OpenAIProcessorConfig` object. The latter might be more intuitive for the admin.

sabaimran:
Psyched to try out SD in Khoj! Might be neat to do a little exploration on what each of these models is better at doing and document that for users.

src/khoj/database/models/__init__.py
debanjum: Fixed that. Field is now used when set

src/khoj/routers/helpers.py
debanjum: Added optional field to associated OpenAI config. Can use api_key or openai_config field for openai model, only api_key for other text to image models. Add validation logic with helpful error messages to guide admin setup

src/khoj/database/models/__init__.py
debanjum: Renamed as requested

PR #832, opened 2024-06-22
Improve Online Search: Parallelize Search, Use Jina Reader API by default

debanjum:
### Overview
Khoj wil be able to do online search out of the box, even for self-hosted users

- Default to Jina search, reader API when no Serper.dev, Olostep API keys
- Run online searches in parallel to process multiple queries faster

### Details
- Jina provides a [reader API](https://github.com/jina-ai/reader) for online search and web page reading. 
  It requires no API key. This provides a good default to enable 
  online search for self-hosted readers requiring no additional setup. 

- Jina search API also returns webpage contents with the results, so
  just use those directly when Jina Search API used instead of
  trying to read webpages separately. The extract relvant content from
  webpage step using a chat model is still used from the
  `read_webpage_and_extract_content` func in this case.

- Parse search results from Jina search API into same format as
  Serper.dev for accurate rendering of online references by clients

- Run online searches in parallel with AsyncIO to process multiple 
  queries faster


documentation/docs/features/online_search.md
```diff
@@ -1,17 +1,21 @@
 # Online Search
 
-By default, Khoj will try to infer which information-sourcing tools are required to answer your question. Sometimes, you'll have a need for outside questions that the LLM's knowledge doesn't cover. In that case, it will use the `online` search feature.
+Khoj will research on the internet to ground it's responses, when real-time or other information outside it's existing knowledge is requested. It will always show any online references it used to respond to your requests.
```
sabaimran, line 3:
```suggestion
Khoj will research on the internet to ground its responses, when it determines that it would need fresh information outside its existing knowledge to answer the query. It will always show any online references it used to respond to your requests.
```

documentation/docs/features/online_search.md
```diff
@@ -1,17 +1,21 @@
...
 
-For example, these queries would trigger an online search:
+By default, Khoj will try infer which information sources, it needs to read to answer your question. This can include reading your documents or researching information online. You can also explicitly trigger an online search by adding the `/online` prefix to your chat query.
```
sabaimran, line 5:
```suggestion
By default, Khoj will try to infer which information sources it needs to read to answer your question. This can include reading your documents or researching information online. You can also explicitly trigger an online search by adding the `/online` prefix to your chat query.
```

src/khoj/processor/tools/online_search.py
```diff
@@ -23,6 +23,10 @@
 SERPER_DEV_API_KEY = os.getenv("SERPER_DEV_API_KEY")
 SERPER_DEV_URL = "https://google.serper.dev/search"
 
+JINA_READER_API_URL = "https://r.jina.ai/"
+JINA_SEARCH_API_URL = "https://s.jina.ai/"
+JINA_API_KEY = os.getenv("JINA_API_KEY")
```
sabaimran, line 28: Wait, is this required for the integration to work? The documentation suggests that it should work out of the box. 

If there's a scenario were someone might benefit from providing the key, maybe add that in the documentation?

tests/test_offline_chat_director.py
```diff
@@ -75,18 +74,14 @@ def test_chat_with_online_content(client_offline_chat):
     response_message = response_message.split("### compiled references")[0]
 
     # Assert
-    expected_responses = ["http://www.paulgraham.com/greatwork.html"]
+    expected_responses = ["https://www.paulgraham.com/greatwork.html"]
```
sabaimran, line 77: Did this work? It shouldn't. His website doesn't have a certificate setup. 
Screenshot 2024-06-24 at 11 26 08

sabaimran (commented):
Awesome being able to provide a better out of the box search experience to self-hosted users.

tests/test_offline_chat_director.py
debanjum: Yeah, I changed it because the online online chat director test was returning https link. Updated test to check if any of the http, https or www. variants are returned

_Side note: The offline chat director tests seem to hang on my machine, unsure why, need to investigate_

PR #836, opened 2024-06-28
Allow just one worker to be the background schedule leader

sabaimran:
In order to avoid additional load on the DB and server in production, elect only one worker to be the background schedule leader to run and execute automations. The other workers are started in a `pause`d state, allowing them to delete and add new jobs, but not running jobs.


src/khoj/utils/state.py
```diff
@@ -31,6 +31,7 @@
 chat_lock = threading.Lock()
 SearchType = utils_config.SearchType
 scheduler: BackgroundScheduler = None
+is_schedule_leader: bool = False
```
sabaimran, line 34: Rather than storing a `bool`, we could store the process lock in state.

src/khoj/configure.py
```diff
@@ -386,3 +386,13 @@ def upload_telemetry():
 def delete_old_user_requests():
     num_deleted = delete_user_requests()
     logger.debug(f"🗑️ Deleted {num_deleted[0]} day-old user requests")
+
+
+@schedule.repeat(schedule.every(17).minutes)
+def wakeup_scheduler():
```
sabaimran, line 392: Disclaimer.

I'm not actually sure if this is required, but I *think* the `leader` worker won't automatically be aware of new jobs being added to the store unless `wakeup` is explicitly called. There are some background operations that will trigger `wakeup`, but I don't think it's dependable, so this may be a good workaround.

PR #852, opened 2024-07-16
Refactor Config API and Settings pages for Reuse and Consistency

debanjum:
### Major
- Reuse get config data logic across config pages on web client
- Make config api endpoint urls consistent
- Rename API path /api/config to /api/configure
- Move Web client settings page to be under `/settings` from the previous `/config` url path
- Move Desktop settings page to be under `/settings` from the previous `/config` url path

### Minor
- Pass isMobileWidth prop to SidePanel via chat share interface
- Turn prettier off instead of throwing error for now
- Do no explicitly add line-clamp plugin as it's in Tailwind by default

PR #857, opened 2024-07-18
Split Configure API into Content, Model API paths

debanjum:
## Major: Breaking Changes
- Move API endpoints under /configure/<type>/model to /api/model/<type>
- Move API endpoints under /api/configure/content/ to /api/content/
- Accept file deletion requests by clients during sync
- Split /api/v1/index/update into /api/content PUT, PATCH API endpoints

## Minor: Create New API Endpoint
- Create API endpoints to get user content configurations

Related: #852


src/khoj/routers/api.py
```diff
@@ -276,6 +278,49 @@ async def transcribe(
...
+    request: Request,
+    name: str,
+    client: Optional[str] = None,
+):
+    user = request.user.object
+
+    split_name = name.split(" ")
+
+    if len(split_name) > 2:
+        raise HTTPException(status_code=400, detail="Name must be in the format: Firstname Lastname")
```
sabaimran, line 304: Is this rule necessary? I guess it's fine if people want to have their middle name's or something. We can just put a character limit.

src/khoj/routers/api_content.py
```diff
@@ -0,0 +1,509 @@
...
+                image=None,
+                github=None,
+                notion=None,
+                plaintext=None,
+            )
+            state.config.content_type = default_content_config
+            save_config_to_file_updated_state()
+            configure_search(state.search_models, state.config.search_type)
+
+        # Extract required fields from config
```
sabaimran, line 395: Is this log line relevant?

sabaimran (approved):
Neat getting the API to a state where it's more semantically coherent.

src/khoj/routers/api.py
debanjum: Not it's not actually required. We could just split into last word = last name and everthing else = first name and put a character limit. But maybe can deal with improving the username management separately?

src/khoj/routers/api_content.py
debanjum: Fixed

PR #858, opened 2024-07-19
Use Single HTTP API for Robust, Generalizable Chat Streaming

debanjum:
Stream Status Messages via Streaming Response from server to web client

## Overview
Use a single API controller for advanced chat streaming and non-streaming mode. Improve code maintainability and generalize to other Khoj client apps. 

Update web client to use the streamed response to show train of thought, stream response and render references.

### Motivation
- Using a single chat API to simplify code, increase dev velocity
- GET API will allow streaming Khoj train of thought, references etc. in a structured form from server to Khoj client apps
   - So streaming Khoj's train of thought and other status updates to clients from server should now be possible
   - Passing auth headers from client apps is possible
      - Websockets don't allow passing request headers
   - Don't need to pass references via `### compiled references:` custom delimiter
- Khoj chat doesn't need bi-directional syncing yet. So using HTTP stream to send updates should be good enough

### Major
- [X] Stream Khoj's train of thought, response via simple HTTP streaming
- [X] Reduce to single chat API controller for streaming and non-streaming chat response
- [X] Update Clients to use new HTTP streaming to render train of thought, extract references, stream response
  - [X] Update web client to use new streaming to render train of thought, extract references, stream response
  - [X] Start streaming Khoj train of thought to Khoj Obsidian plugin using new streaming mode
  - [X] Start streaming Khoj train of though to Khoj Desktop Clients using new streaming mode
  - No updates to Whatsapp, Emacs required as they use non-streaming response, which has the original interface
### Minor
- [X] Save message to conversation history even if client disconnects
- [X] Stream references in a structured form via HTTP streaming instead of using `### compiled references:` separator
- [X] Log total response time and time to first token for overall chat latency analysis


src/interface/obsidian/src/chat_view.ts
```diff
@@ -531,42 +550,38 @@ export class KhojChatView extends KhojPaneView {
...
+        return chatMessageEl;
     }
 
     async renderIncrementalMessage(htmlElement: HTMLDivElement, additionalMessage: string) {
-        this.result += additionalMessage;
+        this.chatMessageState.rawResponse += additionalMessage;
         htmlElement.innerHTML = "";
         // Sanitize the markdown to render
-        this.result = DOMPurify.sanitize(this.result);
+        this.chatMessageState.rawResponse = DOMPurify.sanitize(this.chatMessageState.rawResponse);
```
sabaimran, line 575: Isn't this being sanitized 2x? Since it's being passed directly to `markdownTextToSanitizedHtml` after.

src/khoj/routers/api_chat.py
```diff
@@ -526,141 +521,142 @@ async def set_conversation_title(
...
-
-    async def send_status_update(message: str):
-        nonlocal connection_alive
-        if not connection_alive:
-            return
+    async def event_generator(q: str):
+        start_time = time.perf_counter()
+        ttft = None
+        chat_metadata: dict = {}
+        connection_alive = True
```
sabaimran, line 549: Will the API continue executing when the client breaks the connection, or would it exit immediately? If it's the latter (which I think it is), then we don't need to maintain this state.

src/khoj/routers/api_chat.py
```diff
@@ -526,141 +521,142 @@ async def set_conversation_title(
...
+        user: KhojUser = request.user.object
+        event_delimiter = "␃🔚␗"
```
sabaimran, line 551: I haven't understood why we need this. Wouldn't it be sufficient for the client to try parsing the message as a JSON response, and if it is not a JSON response, then assume that it's the LLM stream?

src/khoj/routers/api_chat.py
```diff
@@ -810,38 +829,52 @@ async def send_rate_limit_message(message: str):
...
             except ValueError as e:
                 logger.warning(
-                    f"Error directly reading webpages: {e}. Attempting to respond without online results", exc_info=True
+                    f"Error directly reading webpages: {e}. Attempting to respond without online results",
+                    exc_info=True,
                 )
 
+        ## Send Gathered References
+        async for result in send_event(
+            "references",
```
sabaimran, line 842: What do you think of defining an Enum for the well-known event types we're sending to the client? It'll help organize/wrangle the information passage here a bit better.

src/khoj/routers/api_chat.py
```diff
@@ -871,310 +910,49 @@ async def send_rate_limit_message(message: str):
...
-                try:
-                    await send_message(f"{item}")
-                except ConnectionClosedOK:
-                    connection_alive = False
-                    logger.info(f"User {user} disconnected web socket. Emitting rest of responses to clear thread")
-
-        await send_message("end_llm_response")
-
-
-@api_chat.get("", response_class=Response)
```
sabaimran, line 905: Nice 🎊

sabaimran (commented):
Tested clients locally, looks good! Excited to get rid of the websocket.

src/khoj/routers/api_chat.py
debanjum: A single message chunk can have partial JSON or multiple JSONs. For both those scenarios the above suggested mechanism would fail. The message chunking is decided by the server and is variable. So relying on a special delimiter helps us avoid parsing status messages on the clients

src/khoj/routers/api_chat.py
debanjum: Yeah, the API continues to execute after the client breaks (i.e `connection_alive=False`), it just doesn't yield any events afterwards. See usage in send_event

src/khoj/routers/api_chat.py
debanjum: Good idea! Updated code to use enum to track chat stream event types

sabaimran (approved):

Issue #819, opened 2024-06-14
[IDEA] Obisidan UX feature improvements


sabaimran:
## Add support for limiting conversation to current file
We have file filters in the web UI that allow us to easily limit a specific conversation to a set of files. It would be useful if I could easily limit my chat in Obsidian to that specific file.
## Add recursive re-chat to the chat window
You should be able to cycle through past messages easily, the way we've just added to the khoj web view with #804 .

shantanuSakpal:
hey, this is an interesting idea that i would like to work on. I have a few doubts, 
you said "...could easily limit my chat in Obsidian to **that** specific file." -- what do you mean by "that file", do you mean the file that i uploaded or the file/note i am currently editing in obsidian

sabaimran:
Hey @shantanuSakpal , glad you're excited about it!

> you said "...could easily limit my chat in Obsidian to that specific file." -- what do you mean by "that file", do you mean the file that i uploaded or the file/note i am currently editing in obsidian

Your intuition is correct, I was imagining it'd be useful to limit context to the file you're currently editing.

To store the context, as you're suggesting, one potential remedy is that a call to `/summarize` would keep that metadata in the chat history, but it would have to be limited to that chat then. 

When a file is indexed by Khoj, it extracts & stores all the text. The summaries are generated based on a file's raw text.

PR #861, opened 2024-07-19
Cycle through chat history in chat input on Obsidian

shantanuSakpal:
This pull request implements the  recursive re-chat functionality as mentioned in issue #819.
I have used an array ```userMessages``` to store the user's input messages and user can toggle through the messages using up and down arrow key by making a new function ```handleArrowKeys```

i ran ```git pull``` before commiting changes.
I have ran ```pre-commit install -t pre-push -t pre-commit``` before creating this pr.
Also i ran ```pytest``` and ``mypy``` 
There were some test failing, but none of them were due to my changes.


shantanuSakpal:
hey @sabaimran , i have added the other have of the feature mentioned in issue #819, just to get an idea of the code and structure. Will soon update about the first feature

src/interface/obsidian/src/chat_view.ts
```diff
@@ -1314,4 +1322,35 @@ export class KhojChatView extends KhojPaneView {
 
         return referencesDiv;
     }
+
+	//function to loop through the user's past messages
+	handleArrowKeys(event: KeyboardEvent) {
+		const chatInput = event.target as HTMLTextAreaElement;
+
+		if (event.key === 'ArrowUp') {
```
sabaimran, line 1330: Nice! Logic looks good. We have this bug in the web view as well, but I think here the `handleArrowKeys` method should _only_ run if the chat input field is currently empty.

sabaimran (commented):
This looks good! Thanks for picking it up, @shantanuSakpal !

src/interface/obsidian/src/chat_view.ts
shantanuSakpal: ah yes, if user accidently presses up, he losses his current input.
what if, rather than checking if input field is empty, we store the current input as well, as user types it. this way if he presses up then presses down, he is back to where he left.

sabaimran: Interesting idea, but I think this would be un-intuitive. If someone has something typed into the input box, they might want to just press the up/down array keys to navigate within the input. They won't be expecting to have their message replaced by previous messages, at which point they might be shocked or frustrated haha.

shantanuSakpal: but there is problem with this, if user presses up once, the previous message with be filled , hence its no longer empty, so they cant toggle further. 

solution 1: when up is pressed, given input is empty, we can show a floating box, containing all the previous messaages (like the one to show list of commands in web),  cropped to first 10 words lets say, when users clicks on any one, the whole message is filled in input.
we can also make the shortcut to be (ctrl + up) so that even when user has something in input, he can press (ctrl +up) and choose any previous message which will be filled after the current input.

solution 2:  when user clicks up, message at index (n-1)  is filled, if he clicks up again, we will check if the user has made any changes to the message (chatInput === messagesArray[index]), if same, means he has not made any changes, so clicking up should fill (n-2)th message ans so on. 
if users makes any changes, we will set the current index to -1 and now clicking up will do nothing.

sabaimran: 1. this UX would be tricky to implement and may not translate that well to the mobile plugin mode. 
2. but again, what if they have already entered text into the input box before initiating the recursive search?

Assessing the options further, I think your original intuition is best. Let's go with this:

>  what if, rather than checking if input field is empty, we store the current input as well, as user types it. this way if he presses up then presses down, he is back to where he left.

This is similar to how the up/down experience is in the terminal, which is actually decent experience-wise once you learn what to expect. You can capture the input in `onChatInput` and store it in a state variable.

shantanuSakpal: okay, i did this terminal-type experience, however, i noticed that user will most probably be writing multi line chats, so he will be using the up arrow to go up in the input quite often, unlike terminal, where the commands are mostly one liner. 
so now, instead of only up arrow, user need to click (ctrl + up) and (ctrl+down) to toggle. 
is this fine?

sabaimran: yup! looks great, thanks for adapting. let me try it out locally.

src/interface/obsidian/src/chat_view.ts
```diff
@@ -76,6 +77,10 @@ export class KhojChatView extends KhojPaneView {
 
         // Clear text after extracting message to send
         let user_message = input_el.value.trim();
+		// Store the message in the array if it's not empty
+		if (user_message) {
+			this.userMessages.push(user_message);
```
sabaimran, line 82: We should also preload the `this.userMessages` list with the values of the user messages retrieved in `getChatHistory`. You can do it in the `chatLogs.forEach` loop. Check if the message was sent by the user, and if so, push it into the array. Otherwise looks good!

src/interface/obsidian/src/chat_view.ts
```diff
@@ -161,7 +169,10 @@ export class KhojChatView extends KhojPaneView {
 
         // Get chat history from Khoj backend and set chat input state
         let getChatHistorySucessfully = await this.getChatHistory(chatBodyEl);
-        let placeholderText = getChatHistorySucessfully ? "Message" : "Configure Khoj to enable chat";
+
+		// Check the user's os and set the modifier key accordingly
+		const modifierKey : string = Platform.isMacOS ? 'cmd' : 'ctrl';
+        let placeholderText : string = getChatHistorySucessfully ? `(${modifierKey}+↑/↓) for prev messages` : "Configure Khoj to enable chat";
```
sabaimran, line 175: Should also check whether `this.userMessages.length > 1` before rendering the help tip. If there are no user messages and successful history load, then use `"Message"`.

sabaimran (commented):
Just a few minor changes, but looks good! Tested it out locally.

src/interface/obsidian/src/chat_view.ts
shantanuSakpal: done, pls check.

sabaimran: looks good! would be good to just address the new merge conflicts and should be good to go.

shantanuSakpal: @sabaimran  resoved the merge conflicts.

tmp_pdf_file_1721415206.018488.pdf
debanjum, line 1: Can we remove this tmp PDF file from this PR? It doesn't seem required

shantanuSakpal: yeah sure. dont know when or how it was added though.

shantanuSakpal: @debanjum  done

sabaimran (approved):

debanjum (approved):
PR looks good. Just switched tabs to spaces and fixed a few minor things

PR #874, opened 2024-08-01
Use new http stream format in the new UX

sabaimran:
Use updated format for HTTP streamed responses from the Khoj server in the new chat UX
Remove references to the websocket connected field, as websocket use has been deprecated


src/interface/web/app/share/chat/page.tsx
```diff
@@ -276,11 +239,66 @@ export default function SharedChat() {
...
+                    if (jsonData.image || jsonData.detail) {
+                        let responseWithReference = handleImageResponse(chunk.data, true);
+                        if (responseWithReference.response) currentMessage.rawResponse = responseWithReference.response;
+                        if (responseWithReference.online) currentMessage.onlineContext = responseWithReference.online;
+                        if (responseWithReference.context) currentMessage.context = responseWithReference.context;
+                    } else if (jsonData.response) {
+                        currentMessage.rawResponse = jsonData.response;
+                    }
+                    else {
+                        console.log("any message", chunk);
```
debanjum, line 288: Is this log line required? And if so, can it be made more informative and maybe a console.debug statement?

src/interface/web/app/share/chat/page.tsx
```diff
@@ -276,11 +239,66 @@ export default function SharedChat() {
                     timestamp: (new Date()).toISOString(),
                     rawQuery: queryToProcess || "",
                 }
+                setProcessQuerySignal(true);
                 setMessages(prevMessages => [...prevMessages, newStreamMessage]);
             }
         })();
     }, [conversationId]);
 
+    function processMessageChunk(rawChunk: string) {
```
debanjum, line 248: `processMessageChunk` and some other related functions seem to be defined twice, here on the `share/page.tsx` and `chat/page.tsx` Can they be deduplicated into `common/chatFunctions.ts` to ease maintenance?

sabaimran: It's a little bit tricky because they're also doing a bunch of state manipulation, but I'll see what can be improved here.

PR #883, opened 2024-08-06
Fix indexing content from Emacs

aam-at:
Previously `force` was passed as a query param to the single indexing API. After the recent API updates, it is meant to select the API method to use (PATCH vs PATCH). Converting `force` argument to a bool fixes implementing this new behavior


debanjum (approved):
Ah, good catch, this should fix indexing your documents on Khoj from Emacs!

PR #885, opened 2024-08-07
Add model_config for crossencoder model

aam-at:
Add `model_config` for crossencoder model, so the server admin can use models which require `trust_remote_code` to run locally, e.g. [Alibaba-NLP/gte-multilingual-reranker-base](https://huggingface.co/Alibaba-NLP/gte-multilingual-reranker-base).


debanjum (approved):
Thanks for adding support to pass cross encoder model configs. Code changes look good!

PR #887, opened 2024-08-11
Improve search model config display for admin

aam-at:
Currently, the search model config display for admins only shows the `id` of the search model config, which is not very informative. This commit enhances the admin console by displaying the name of the search model config (`name`), as well as the bi-encoder model (`bi_encoder`) and cross-encoder model (`cross_encoder`) along the `id`.


debanjum (approved):
Good idea! This should improve the admin experience to update Khoj search models

PR #888, opened 2024-08-11
Add user conversation config to admin console

aam-at:
The user can't select the chat model under settings because user conversation model is not set. This commit add an option for the user to configure the conversation model.

Screenshot from 2024-08-11 22-26-55

### DIFF
--- a/src/khoj/database/admin.py
+++ b/src/khoj/database/admin.py
@@ -26,6 +26,7 @@ from khoj.database.models import (
     SpeechToTextModelOptions,
     Subscription,
     TextToImageModelConfig,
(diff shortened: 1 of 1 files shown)


sabaimran:
Hey @aam-at ! Thanks for raising the PR, but this issue was addressed separately in this PR: https://github.com/khoj-ai/khoj/pull/886 with this snippet of code:

```
    selected_chat_model_config = (
        ConversationAdapters.get_conversation_config(user) or ConversationAdapters.get_default_conversation_config()
    )
```

We also happened to add the same change you described in that same PR. Cheers, and thanks for all the contributions!

PR #896, opened 2024-08-25
Add support for custom search model-specific thresholds

sabaimran:
- Add a `bi_encoder_confidence_threshold` to the search models that allows us to have model-specific confidence intervals


src/khoj/routers/api.py
```diff
@@ -83,7 +83,7 @@ async def search(
         n=n,
         t=t,
         r=r,
-        max_distance=max_distance,
+        max_distance=max_distance or math.inf,
```
debanjum, line 86: should this just pass max_distance since `math.inf` is decided later in the call stack if search model doesn't have any max_distance set and no max_distance passed via API?

```suggestion
        max_distance=max_distance,
```

debanjum (approved):
Looks good!

PR #898, opened 2024-08-26
Update docker-compose.yml

lvnilesh:
Handles deprecation of version reference


sabaimran:
Thanks for raising this @lvnilesh . Looks like the field is deprecated and the version parameter was anyway un-enforced. Thanks! https://docs.docker.com/reference/compose-file/version-and-name/ .

PR #909, opened 2024-09-12
Support using image generation models like Flux via Replicate

debanjum:
- Make generate better image prompt chat actor add composition details
- Generate vivid images with DALLE-3
- Support using image generation models like Flux via Replicate


src/khoj/routers/helpers.py
```diff
@@ -961,17 +964,67 @@ async def text_to_image(
...
                 return
 
-    with timer("Convert image to webp", logger):
-        # Convert png to webp for faster loading
-        webp_image_bytes = convert_image_to_webp(decoded_image)
+    elif text_to_image_config.model_type == TextToImageModelConfig.ModelType.REPLICATE:
+        with timer("Generate image using Replicate", logger):
+            try:
+                # Create image generation task on Replicate
+                create_prediction_url = f"https://api.replicate.com/v1/models/{text2image_model}/predictions"
```
sabaimran, line 980:
```suggestion
                replicate_prediction_url = f"https://api.replicate.com/v1/models/{text2image_model}/predictions"
```

src/khoj/routers/helpers.py
```diff
@@ -927,11 +928,13 @@ async def text_to_image(
...
                     prompt=improved_image_prompt,
                     model=text2image_model,
+                    style="vivid",
                     response_format="b64_json",
                     extra_headers=auth_header,
                 )
                 image = response.data[0].b64_json
-                decoded_image = base64.b64decode(image)
+                # Decode base64 png and convert it to webp for faster loading
+                webp_image_bytes = convert_image_to_webp(base64.b64decode(image))
```
sabaimran, line 937: Let's move this `text_to_image` function into a separate file? Maybe we can create a `/processor/image` folder and make `generation.py` over there with this function. Ideally the per-model implementations should also be broken down further, e.g., a separate function for each OpenAI, Replicate, and StabilityAI. IMO it'll make the code more readable/organized.

src/khoj/routers/helpers.py
```diff
@@ -927,11 +928,13 @@ async def text_to_image(
                 response = state.openai_client.images.generate(
                     prompt=improved_image_prompt,
                     model=text2image_model,
+                    style="vivid",
```
sabaimran, line 931: Maybe use an all caps magic variable name to set this, since it contains specific business logic? e.g., `OPENAI_IMAGE_GEN_STYLE = "vivid"`. Generally prefer to have things like this called out in a way where it's obvious that this variable is non-standard and where to learn more.

src/khoj/routers/helpers.py
```diff
@@ -961,17 +964,67 @@ async def text_to_image(
...
+                }
+                create_prediction = requests.post(create_prediction_url, headers=headers, json=json).json()
+
+                # Get status of image generation task
+                get_prediction_url = create_prediction["urls"]["get"]
+                get_prediction = requests.get(get_prediction_url, headers=headers).json()
+                status = get_prediction["status"]
+                retry_count = 1
+
+                # Poll the image generation task for completion status
```
sabaimran, line 1002: Man, that's rough that we have to poll it. Wouldn't it block the whole thread? How long does an image take, approx? If it' just like 10-15s, it's manageable, but if it's like 45s we might have to think about a better way to do this. I suppose we could keep sending updates to the client as well.

debanjum: Yeah, the polling required for the replicate API is annoying. Time depends on the model but haven't seen image generation to take more than 15s. E.g flux schnell takes ~3s, flux dev takes ~6s and flux pro takes ~15s in my testing

PR #914, opened 2024-09-16
Add a unique_id field for identifiying conversations

sabaimran:
- Add a new `unique_id` field which becomes the public-facing identifier for conversations
- Update front-end and back-end pages to use it. Navigating to a specific conversation prefills it into the window, such as `/chat?v={uniqueId}`


sabaimran:
Note: might be better to migrate to using `unique_id` everywhere for interactions with the `Conversation` object, not just in the web app.

PR #918, opened 2024-09-19
Convert the default conversation id to a uuid, plus other fixes

sabaimran:
- Convert the existing conversation_id primary_key to a UUID
- Migrate relevant APIs and other metadata to use the new field format
- Fix a couple of minor issues in the conversation UI
  - Minimize the input bar
  - Make file drag/drop experience less jittery
  - Add conversation mutation actions in the title bar


debanjum (approved):
Changes look good! Glad we could incorporate conversation id updated at the DB level to minimize app level code changes

PR #919, opened 2024-09-19
Improve Khoj First Run, Docker Setup and Documentation

### LINKED ISSUE #882: Pre-loaded configurations
The setup for the locally hosted Docker version is still complicated (it used to be way simpler in the earlier version of Khoj), and the instructions are not very clear on how to load various models, embedding engines, etc.

Consider offering pre-loaded choices for LLMs and Embedding choices from a drop-down menu.

For example, a drop-down menu in the chat models where users can simply select the LLM provider, model, and input their API key.
For embeddings likewise: a list of all the various models users can use, from a simple drop-down menu rather than having to type it in.

I keep trying this app here and there but I only ever managed to get the initial version to ever work.

Thanks.

debanjum:
## Improve
- Intelligently initialize a decent default set of chat model options
- Create non-interactive mode. Auto set default server configuration on first run via Docker

## Fix
- Make RapidOCR dependency optional as flaky requirements
- Set default openai text to image model correctly during initialization

## Details
Improve initialization flow during first run to remove need to configure Khoj:

- Set Google, Anthropic Chat models too
  Previously only Offline, Openai chat models could be set during init

- Add multiple chat models for each LLM provider
  Interactively set a comma separated list of models for each provider

- Auto add default chat models for each provider in non-interactive
  model if the `{OPENAI,GEMINI,ANTHROPIC}_API_KEY` env var is set
  - Used when server run via Docker as user input cannot be processed to configure server during first run

- Do not ask for max_tokens, tokenizer for offline models during
  initialization. Use better defaults inferred in code instead

- Explicitly set default chat model to use
  If unset, it implicitly defaults to using the first chat model.
  Make it explicit to reduce this confusion

Resolves #882


documentation/docs/advanced/admin.md
```diff
@@ -0,0 +1,74 @@
...
+- `model-type`: Set to the appropiate model provider
+- `openai-config`: For image generation models available via OpenAI (compatible) API you can set the appropriate OpenAI Processor Conversation Settings instead of specifying the `api-key` field above
+
+### Speech to Text Model Options
+Add speech to text models with these settings. Khoj currently only supports whisper speech to text model via OpenAI API or Offline
+
+### Voice Model Options
+Add text to speech models with these settings. Khoj currently supports models from ElevenLabs
+
+### Reflective Questions
```
sabaimran, line 66: This should be removed -- these aren't being used anymore.

documentation/docs/advanced/admin.md
```diff
@@ -0,0 +1,74 @@
...
+- `api-key`: Set to your OpenAI, Stability or Replicate API key
+- `model`: Set the model name available over the selected model provider
+- `model-type`: Set to the appropiate model provider
+- `openai-config`: For image generation models available via OpenAI (compatible) API you can set the appropriate OpenAI Processor Conversation Settings instead of specifying the `api-key` field above
+
+### Speech to Text Model Options
+Add speech to text models with these settings. Khoj currently only supports whisper speech to text model via OpenAI API or Offline
+
+### Voice Model Options
+Add text to speech models with these settings. Khoj currently supports models from ElevenLabs
```
sabaimran, line 64:
```suggestion
Add text to speech models with these settings. Khoj currently supports models from [ElevenLabs](https://elevenlabs.io/).
```

documentation/docs/advanced/admin.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Reflective Questions
+Add any filler questions with these settings
+
+## User Data
+- Users, Entrys, Conversations, Subscriptions, Github configs, Notion configs, User search configs, User conversation configs, User voice configs
+
+## Miscellaneous Data
+- Process Locks: Persistent Locks for Automations
+- Client Applications:
```
sabaimran, line 74:
```suggestion
- Client Applications:
Client applications allow you to setup third party applications that can query your Khoj server using a client application ID + secret. The secret would go in a bearer token, like so.
```

documentation/docs/get-started/setup.mdx
```diff
@@ -14,37 +14,61 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 ```
 
-## Setup
+## Setup Khoj
 These are the general setup instructions for self-hosted Khoj.
-You can install the Khoj server using either Docker or Pip.
+You can install the Khoj server using either [Docker](?method=docker) or [Pip](?method=pip).
```
sabaimran, line 19: How does this deep-linking work? I don't see the `method` ID being used.

sabaimran:
Awesome getting some of this setup path cleaned, documentation improved, and default case made easier 🎉

sabaimran (approved):

documentation/docs/get-started/setup.mdx
debanjum: Good catch, fixed by converting to use `server` which is a valid query param from Tabs

Issue #915, opened 2024-09-16
[FIX] The "Create Agent" button doesn't trigger anything


vibl:
## Describe the bug
The "Create Agent" button doesn't trigger anything (apart from a GET request to /sessions, according to DevTools).

No error in the console. 

## To Reproduce
Click on the button.

## Platform
- Server:
    - [X] Cloud-Hosted (https://app.khoj.dev)
- Client:
    - [ ] Obsidian
    - [ ] Emacs
    - [ ] Desktop app
    - [X] Web browser: Chrome latest
    - [ ] WhatsApp
- OS:
    - [ ] Windows
    - [ ] macOS
    - [X] Linux
    - [ ] Android
    - [ ] iOS

sabaimran:
Hey @vibl , There's actually supposed to be a `coming soon` notification that comes up there, but it's really finnicky. This feature is in progress atm. Feel free to share any info you wnat about what you're looking for here or what you'd want. 

![Screenshot 2024-09-16 at 13 51 21](https://github.com/user-attachments/assets/dc9a5cd6-42e9-4aa8-a1fb-5ea6a9540b0f)

shantanuSakpal:
yeah, its not working for me either, on ubuntu 24, brave, looking into it, should be something minor

PR #923, opened 2024-09-25
Only Live Scrolling when at Page Bottom and Add Button to Scroll to Page Bottom on Web App

### LINKED ISSUE #758: [IDEA] Live Scrolling with Scrolling Buffer
## Live Scrolling with Scrolling Buffer:
Scrolling Buffer: Implement a buffer that stores a certain number of previously generated lines. This allows users to scroll up and see the recently generated text without affecting the ongoing generation process.
Live Scrolling: As the AI generates new lines, automatically update the displayed content within the buffer. This way, the user can scroll up and review the recent history without the page jumping or content disappearing.

shantanuSakpal:
### Overview
i have modified the ```chatHistory.tsx``` and added the following things:
1. `useEffect` to monitor if auto-scroll is needed
2. a button to scroll down 

### Details
1. Only auto scroll Khoj's streamed response when scroll is near bottom of page
   Allows scrolling to other messages in conversation while Khoj is formulating and streaming its response
2. Add button to scroll to bottom of the chat page
3. Scroll to most recent conversation turn on conversation first load
   It's a better default to anchor to most recent conversation turn (i.e most recent user message)
4. Smooth scroll when Khoj's chat response is streamed
   Previously the scroll would jitter during response streaming
5. Anchor scroll position when fetch and render older messages in conversation
   Allow users to keep their scroll position when older messages are fetched from server and rendered

Resolves #758

### DIFF
--- a/src/interface/web/app/components/chatHistory/chatHistory.tsx
+++ b/src/interface/web/app/components/chatHistory/chatHistory.tsx
@@ -76,6 +76,7 @@ export default function ChatHistory(props: ChatHistoryProps) {
     >(null);
     const [fetchingData, setFetchingData] = useState(false);
     const [isNearBottom, setIsNearBottom] = useState(true);
+    const [scrollPosition, setScrollPosition] = useState(0);
     const isMobileWidth = useIsMobileWidth();
     const scrollAreaSelector = "[data-radix-scroll-area-viewport]";
 
@@ -118,6 +119,12 @@ export default function ChatHistory(props: ChatHistoryProps) {
         const observer = new IntersectionObserver(
             (entries) => {
                 if (entries[0].isIntersecting && hasMoreMessages) {
+                    const scrollArea = scrollAreaRef.current?.querySelector(
+                        scrollAreaSelector,
+                    ) as HTMLElement;
+                    if (scrollArea) {
+                        setScrollPosition(scrollArea.scrollHeight - scrollArea.scrollTop);
+                    }
                     setFetchingData(true);
                     fetchMoreMessages(currentPage);
                     setCurrentPage((prev) => prev + 1);
@@ -148,6 +155,12 @@ export default function ChatHistory(props: ChatHistoryProps) {
             }
         }
     }, [props.incomingMessages]);
+
+    const adjustScrollPosition = () => {
+        const scrollArea = scrollAreaRef.current?.querySelector(scrollAreaSelector) as HTMLElement;
+        if (!scrollArea) return;
+        scrollArea.scrollTo({ top: scrollArea.scrollHeight - scrollPosition, behavior: "auto" });
+    };
 
     function fetchMoreMessages(currentPage: number) {
         if (!hasMoreMessages || fetchingData) return;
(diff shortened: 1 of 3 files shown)


debanjum (commented):
Hey @shantanuSakpal, thanks a ton for creating these changes! I've pushed some changes:
- Fixes to the auto-scroll logic
- Styling Improvements to the scroll to bottom button
- Clean up the old code to detect when at bottom of screen

Please have a look and try them out to verify these changes works for you as well

shantanuSakpal:
hey , umm, there is a slight problem , i am having issues running the offline model, and when i use the open ai gpt-4o, it give the whole result in one go, so i cannot test the live scrollling feature. 
but if it works fine on you machine , i think it should to good to push to prod.

shantanuSakpal:
also 
image
this is a bit hard to see dont you think?
maybe use white text for dark mode? 
image

debanjum:
@shantanuSakpal, oh yeah to test streaming related changes you'd need to use `yarn watch` instead of `yarn dev` in the web app director and test the web app at http://localhost:42110 instead of http://localhost:3000. This is useful for testing changes across the web app and the khoj python server. 

Note: `yarn watch` automatically runs `yarn export` on any edit to the web app files (i.e under `/src/interface/web`). This builds the web app and copies them to the servers static serving folder.

shantanuSakpal:
Alright, did ```yarn watch```, and its works great :+1: .
And also i noticed it works faster like this, as compared to ```yarn dev```. Can you tell me why is it? is it because the files are already compiled ? 

also added fix for issue #915. 
made a pr [here](https://github.com/khoj-ai/khoj/pull/925) but it will also be merged with this pr right?

debanjum:
> also added fix for issue https://github.com/khoj-ai/khoj/issues/915.
> made a pr https://github.com/khoj-ai/khoj/pull/925 but it will also be merged with this pr right?

Thanks! The 2 PRs should be merged separately. But PR #925 seems to contain changes from this PR (#923). Could you update PR #925 to just contain the changes meant for fixing that issue (e.g keeping only the tooltip delayDuration code)? Of course, it's a small change so we can merge that after verifying fix works soon

shantanuSakpal:
i am having some issue. When i run ```yarn watch```, it is not showing the latest changes on http://localhost:42110/ . However when i run server using ```yarn dev```, it shows the latest changes on  http://localhost:3000/.
could it be because it is caching the build, but it should still build the changed files again right?

sabaimran:
`yarn watch` takes some time to rebuild and package the static files. It can take up to 10-15 seconds. `yarn dev` does work more or less instantly. The build files should definitely be updated with `watch`.

I'll typically quickly prototype anything I need to with `yarn dev`, and then use `yarn watch` when I need to verify it. Of course, this doesn't work for streaming due to the way next handles streamed responses.

Are you seeing the logs indicate re-builds being triggered in your front-end server?

shantanuSakpal:
Yes, I do see that it rebuilds the app, but still I don't see the changes.

sabaimran:
Just tested this out! This is awesome, it makes the chat experiences so much smoother. Thanks @shantanuSakpal ! 🎊 

To see the changes, you'll also have to refresh the page on `http://localhost:42110`. Perhaps that's the issue you're running into? And make sure you're in the same folder where you're running the server.

PR #925, opened 2024-09-26
added tool tip popup on click for create agent

shantanuSakpal:
This pr fixes issue #915.

So the problem was the tool tip was visible on hover, but it was slow, so before the tool tip popped up, the user would click on the button and this stopped the tool tip from popping up. 

so i reduced the popup delay to 100ms. now as soon as user hovers over the button , he will see that its a feature coming soon!


sabaimran:
hey @shantanuSakpal , not sure why you closed this, but it won't be necessary to fix this since we're rolling out the agent creation flow soon. 

Thanks for finding the issue though!

PR #929, opened 2024-09-30
Make Online Search Location Aware

debanjum:
## Overview
Add user country code as context for doing online search with serper.dev API.
This should find more user relevant results from online searches by Khoj

### Major
- Default to using system clock to infer user timezone on js clients
- Infer country from timezone when only timezone received by chat API
- Localize online search results to user country when location available

### Minor
- Add `__str__` func to `LocationData` class to deduplicate location string generation

### DIFF
--- a/src/interface/desktop/chat.html
+++ b/src/interface/desktop/chat.html
@@ -60,6 +60,7 @@
         let region = null;
         let city = null;
         let countryName = null;
+        let countryCode = null;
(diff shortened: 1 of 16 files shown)


sabaimran (approved):
Unclear how this will affect overall results quality, but let's see how it goes. Should be better for people using desktop/web clients to get localized current events, which is the main value prop of online results!

PR #936, opened 2024-10-16
Make Chat and Online Search Resilient and Faster

debanjum:
### New
- Support using [Firecrawl](https://firecrawl.dev) to read web pages
- Add, switch and re-prioritize web page reader(s) to use via the admin panel

### Speed
- Improve response speed by aggregating web page read, extract queries to run only once for each web page

### Response Resilience
- Fallback through enabled web page readers until web page read
- Enable reading web pages on the internal network for self-hosted Khoj running in anonymous mode
- Try respond even if web search, web page read fails during chat
- Try respond even if document search via inference endpoint fails

### Fix
- Return data sources to use if exception in data source chat actor

### Configure web page readers to use
 - Only the web scraper set in Server Chat Settings via the Django admin panel, if set
 - Otherwise use the web scrapers added via the Django admin panel (in order of priority), if set
 - Otherwise, use all the web scrapers enabled by settings API keys via environment variables (e.g `FIRECRAWL_API_KEY`, `JINA_API_KEY` env vars set), if set
 - Otherwise, use Jina to web scrape if no scrapers explicitly defined
 
For self-hosted setups running in anonymous-mode, the ability to directly read webpages is also enabled by default. This is especially useful for reading webpages in your internal network that the other web page readers will not be able to access.

### Aggregate webpage extract queries to run once for each distinct web page

Previously, we'd run separate webpage read and extract relevant
content pipes for each distinct (query, url) pair.

Now we aggregate all queries for each url to extract information from
and run the webpage read and extract relevant content pipes once for
each distinct URL.

Even though the webpage content extraction pipes were previously being
run in parallel. They increased the response time by
1. adding more ~duplicate context for the response generation step to read
2. being more susceptible to variability in web page read latency of the parallel jobs

The aggregated retrieval of context for all queries for a given
webpage could result in some hit to context quality. But it should
improve and reduce variability in response time, quality and costs. 

This should especially help with speed and quality of online search 
for offline or low context chat models.


src/khoj/database/adapters/__init__.py
```diff
@@ -1031,6 +1037,52 @@ async def aget_advanced_conversation_config(user: KhojUser = None):
...
+        return None
+
+    @staticmethod
+    async def aget_enabled_webscrapers():
+        enabled_scrapers = []
+        server_webscraper = await ConversationAdapters.aget_server_webscraper()
+        if server_webscraper:
+            # Only use the webscraper set in the server chat settings
+            enabled_scrapers = [
+                (server_webscraper.type, server_webscraper.api_key, server_webscraper.api_url, server_webscraper.name)
```
sabaimran, line 1054: Should we create a simple pydantic object, like `WebScraperService` to make this code more maintainable? I find tuples mind-numbing 😅

src/khoj/database/adapters/__init__.py
```diff
@@ -1031,6 +1037,52 @@ async def aget_advanced_conversation_config(user: KhojUser = None):
...
+            # Jina is the default fallback scrapers to use as it does not require an API key
+            api_url = os.getenv("JINA_READER_API_URL", "https://r.jina.ai/")
+            enabled_scrapers.append((WebScraper.WebScraperType.JINA, os.getenv("JINA_API_KEY"), api_url, "Jina"))
+
+            # Only enable the direct web page scraper by default in self-hosted single user setups.
+            # Useful for reading webpages on your intranet.
+            if state.anonymous_mode or in_debug_mode():
+                enabled_scrapers.append((WebScraper.WebScraperType.DIRECT, None, None, "Direct"))
+
+        return enabled_scrapers
```
sabaimran, line 1084: Can we create a cache for this result? Unless the server setting is changed, it should always return the same `enabled_scrapers`. We can have a post `save` signal on the `WebScraper` to trigger `aget_enabled_webscrapers`, though this won't work in a distributed (multi-container) environment. We can also use time-based invalidation. What do you think? I'm okay with also merging this in as-is only because the frequency of scraping shouldn't be terribly high and it's just making two calls to the DB.

src/khoj/database/models/__init__.py
```diff
@@ -243,13 +243,59 @@ class GithubRepoConfig(BaseModel):
     github_config = models.ForeignKey(GithubConfig, on_delete=models.CASCADE, related_name="githubrepoconfig")
 
 
+class WebScraper(BaseModel):
+    class WebScraperType(models.TextChoices):
+        FIRECRAWL = "firecrawl", gettext_lazy("Firecrawl")
```
sabaimran, line 248: Why `gettext_laxy`?

src/khoj/processor/tools/online_search.py
```diff
@@ -10,27 +10,34 @@
...
 SERPER_DEV_API_KEY = os.getenv("SERPER_DEV_API_KEY")
 SERPER_DEV_URL = "https://google.serper.dev/search"
 
-JINA_READER_API_URL = "https://r.jina.ai/"
 JINA_SEARCH_API_URL = "https://s.jina.ai/"
 JINA_API_KEY = os.getenv("JINA_API_KEY")
 
-OLOSTEP_API_KEY = os.getenv("OLOSTEP_API_KEY")
-OLOSTEP_API_URL = "https://agent.olostep.com/olostep-p2p-incomingAPI"
+FIRECRAWL_TO_EXTRACT = is_env_var_true("FIRECRAWL_TO_EXTRACT")
```
sabaimran, line 39: Do we need this? I'd think the settings via server chat settings / selection logic should be sufficient in determining whether this scraper is available.

src/khoj/processor/tools/online_search.py
```diff
@@ -83,33 +90,36 @@ async def search_online(
...
+        for link, data in webpages.items()
     ]
     results = await asyncio.gather(*tasks)
 
     # Collect extracted info from the retrieved web pages
-    for subquery, webpage_extract, url in results:
+    for subqueries, url, webpage_extract in results:
         if webpage_extract is not None:
-            response_dict[subquery]["webpages"] = {"link": url, "snippet": webpage_extract}
+            response_dict[subqueries.pop()]["webpages"] = {"link": url, "snippet": webpage_extract}
```
sabaimran, line 122: My reading of this is that just the top `subquery` would get used to populated the `response_dict`, no? What happens to the other ones?

src/khoj/utils/helpers.py
```diff
@@ -435,6 +437,50 @@ def is_internet_connected():
...
+
+        # Check for common internal TLDs
+        internal_tlds = [".local", ".internal", ".private", ".corp", ".home", ".lan"]
+        if any(hostname.endswith(tld) for tld in internal_tlds):
+            return True
+
+        # Check for non-standard ports
+        # if parsed_url.port and parsed_url.port not in [80, 443]:
+        #     return True
+
```
sabaimran, line 473: Remove commented out code

src/khoj/utils/helpers.py
```diff
@@ -435,6 +437,50 @@ def is_internet_connected():
         return False
 
 
+def is_internal_url(url: str) -> bool:
```
sabaimran, line 440: Neat idea, would be interesting to see different applications of this.

src/khoj/processor/tools/online_search.py
```diff
@@ -220,6 +265,54 @@ async def read_webpage_with_jina(web_url: str) -> str:
...
+    system_prompt = f"""
+{prompts.system_prompt_extract_relevant_information}
+
+{personality_context}
+User Query: {", ".join(queries)}
+
+Collate only relevant information from the website to answer the target query and in the provided JSON schema.
+""".strip()
+
+    params = {"url": web_url, "formats": ["extract"], "extract": {"systemPrompt": system_prompt, "schema": schema}}
```
sabaimran, line 307: Can we make it optional to use the system prompt? Or is that required? Alternatively, we may want to shift control back down to Khoj to do what it wants to with the raw web page content.

sabaimran (commented):
Neat stuff! Excited to see this boost our scraping reliability. 💃🏾

src/khoj/database/adapters/__init__.py
debanjum: Good point (on caching this and using save to invalidate cache). But maybe better as a future optimization?

src/khoj/database/adapters/__init__.py
debanjum: Good point. I've updated the code to use the WebScraper Django DB model directly. This has simplified this code

PR #940, opened 2024-10-17
Generate dynamic diagrams (via Excalidraw)

sabaimran:
Add support for generating dynamic diagrams in flow with Excalidraw (https://github.com/excalidraw/excalidraw). This happens in three steps:
1. Default information collection & intent determination step.
2. Improving the overall guidance of the prompt for generating a JSON, Excalidraw-compatible declaration.
3. Generation of the diagram to output to the final UI.

| Inline diagram | Expanded diagram |
|-|-|
|Screenshot 2024-10-17 at 13 58 33|Screenshot 2024-10-17 at 13 58 41|

Add support on the web UI for rendering the diagrams in-flow.

## Pending issues
- Stylesheet not loading upfront sometimes for some reason, so renders with ugly default font
- Handle response for Obsidian, Emacs, Desktop (reroute to web UI)


rehandaphedar:
Hi, not sure if this should be in a separate issue, but I think these two programs would be very good candidates for extending a feature like this:
-  [d2](https://d2lang.com/) would be a very good candidate for extending a feature like this. It's very "clean" (single `d2` binary that outputs svg, no complex dependencies).
- [Kroki!](https://kroki.io/) is a single API for multiple text to diagram programs, including d2 and excalidraw.

src/interface/web/app/components/excalidraw/excalidrawWrapper.tsx
```diff
@@ -0,0 +1,210 @@
...
+        }
+        setExcalidrawCurrentState(elements);
+    };
+
+    return (
+        <div className="relative">
+            <div
+                className={`${expanded ? "fixed inset-0 bg-black bg-opacity-50 backdrop-blur-sm z-50 flex items-center justify-center" : ""}`}
+            >
+                <Button
```
debanjum, line 172: It'd be great if the expanded view can be dismissed by pressing Esc via keyboard. Makes it quick and easy to switch out of the expanded view

src/interface/web/app/components/chatMessage/chatMessage.tsx
```diff
@@ -554,6 +574,7 @@ const ChatMessage = forwardRef<HTMLDivElement, ChatMessageProps>((props, ref) =>
                     className={styles.chatMessage}
                     dangerouslySetInnerHTML={{ __html: markdownRendered }}
                 />
+                {excalidrawData && <ExcalidrawComponent data={excalidrawData} />}
```
debanjum, line 577: Can we make it such that the `ExcalidrawComponent` only captures scroll when clicked? (Or disable scroll in excalidraw block when previous scroll event was outside the excalidraw window and <1 second ago). 

Currently when scrolling through chat history excalidraw diagrams in it capture the scroll when scrolling past them. This is unexpected and annoying.

src/interface/web/app/layout.tsx
```diff
@@ -40,7 +40,7 @@ export default function RootLayout({
 }>) {
     return (
         <html lang="en">
-            <meta
+            {/* <meta
```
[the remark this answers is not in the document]
sabaimran: Good catch. Not intentional, just had to comment out while testing out the integration. Reverted.

src/interface/web/app/components/chatMessage/chatMessage.tsx
sabaimran: Just updating to say I haven't been able to workaround this issue. I suggest we merge it now and loop back to it after the release.

PR #942, opened 2024-10-19
Chat with Multiple Images. Support Vision with Gemini

### LINKED ISSUE #716: Incorporate Gemini & Gemini Vision support
## Provide Chat Model support for Gemini

With the recent advances with Gemini and long context window features, I would like to add Gemini as a chat model available for administrator to include that as a choice of chat model.

Potentially I would like this to be assigned to me since I'd like to use it also.

### LINKED ISSUE #921: [FIX] khoj vision mode do not support multiple images
## Describe the bug
A clear and concise description of what the bug is. Please include what you were expecting to happen vs. what actually happened.

support more than one image is needed

debanjum:
## Overview
- Add vision support for Gemini models in Khoj
- Allow sharing multiple images as part of user query from the web app
- Handle multiple images shared in query to chat API

## Screenshots

| Desktop | Mobile |
|--------|--------|
| Chat with Multiple Images - Desktop | Chat with Multiple Images - Mobile |

Closes #716 
Closes #921


src/interface/web/app/components/chatInputArea/chatInputArea.module.css
```diff
@@ -1,4 +1,5 @@
 div.actualInputArea {
     display: grid;
     grid-template-columns: auto 1fr auto auto;
+    max-width: 700px;
```
sabaimran, line 4: Why is this necessary?

src/interface/web/app/components/chatInputArea/chatInputArea.tsx
```diff
@@ -460,57 +466,59 @@ export default function ChatInputArea(props: ChatInputProps) {
...
-                    </TooltipProvider>
-                )}
-                <Button
-                    className={`${(!message || recording) && "hidden"} ${props.agentColor ? convertToBGClass(props.agentColor) : "bg-orange-300 hover:bg-orange-500"} rounded-full p-1 m-2 h-auto text-3xl transition transform md:hover:-translate-y-1`}
-                    onClick={onSendMessage}
-                    disabled={props.sendDisabled}
-                >
-                    <ArrowUp className="w-6 h-6" weight="bold" />
-                </Button>
+                <div className="flex items-end pb-4">
```
sabaimran, line 469: Is the change here just that you've added a wrapper div around the component? Hard to tell with the diff.

src/interface/web/app/components/chatMessage/chatMessage.tsx
```diff
@@ -328,8 +327,14 @@ const ChatMessage = forwardRef<HTMLDivElement, ChatMessageProps>((props, ref) =>
             .replace(/\\\[/g, "LEFTBRACKET")
             .replace(/\\\]/g, "RIGHTBRACKET");
 
-        if (props.chatMessage.uploadedImageData) {
-            message = `![uploaded image](${props.chatMessage.uploadedImageData})\n\n${message}`;
+        if (props.chatMessage.images && props.chatMessage.images.length > 0) {
+            const imagesInMd = props.chatMessage.images
+                .map(
+                    (image, index) =>
+                        `<div class="${styles.imageWrapper}"><img src="${image.startsWith("data%3Aimage") ? decodeURIComponent(image) : image}" alt="uploaded image ${index + 1}" /></div>`,
```
sabaimran, line 334: Should we sanitize this input?

src/khoj/routers/api_chat.py
```diff
@@ -564,9 +564,9 @@ async def chat(
     country = body.country or get_country_name_from_timezone(body.timezone)
     country_code = body.country_code or get_country_code_from_timezone(body.timezone)
     timezone = body.timezone
-    image = body.image
+    raw_images = body.images
```
sabaimran, line 567: Should we / can we enforce any kind of size constraint on this parameter? Doesn't need to be super aggressive, but probably makes sense to bake in some rate limiting.

sabaimran (approved):
🚀 awesome being able to chat w/ multiple images & having gemini support.

src/interface/web/app/components/chatInputArea/chatInputArea.module.css
debanjum: It was added in the middle to constrain width when lots of images are added but with the carousel implementatino is not necessary. Removed

src/khoj/routers/api_chat.py
debanjum: Added API rate limiter to constrain number of images and size of images that the API accepts from clients.

PR #944, opened 2024-10-20
Create explicit flow to enable the free trial

sabaimran:
The current design is confusing. It obfuscates the fact that the user is on a free trial. This design will make the opt-in explicit and more intuitive. Adds consistency to a 7-day trial across the application. 

| New User | Enabled Free Trial |
|-|-|
|Screenshot 2024-10-22 at 12 50 49| Screenshot 2024-10-22 at 13 14 18|


src/interface/web/app/settings/page.tsx
```diff
@@ -866,10 +903,10 @@ export default function SettingsView() {
                                                         Futurist (Trial)
                                                     </p>
                                                     <p className="text-gray-400">
-                                                        You are on a 14 day trial of the Khoj
+                                                        You are on a 7 day trial of the Khoj
```
debanjum, line 906: Should we use the `LENGTH_OF_FREE_TRIAL` const here too?

src/khoj/database/adapters/__init__.py
```diff
@@ -185,11 +188,29 @@ async def aget_or_create_user_by_email(email: str) -> tuple[KhojUser, bool]:
 
     user_subscription = await Subscription.objects.filter(user=user).afirst()
     if not user_subscription:
-        await Subscription.objects.acreate(user=user, type="trial")
+        await Subscription.objects.acreate(user=user, type="standard")
```
debanjum, line 191: The subscription for users created via Google OAuth in the `create_user_by_google_token` function should also be updated to `type="standard"`. It's currently set to type="trial"

src/khoj/routers/helpers.py
```diff
@@ -1477,10 +1478,16 @@ def get_user_config(user: KhojUser, request: Request, is_detailed: bool = False)
 
     user_subscription_state = get_user_subscription_state(user.email)
     user_subscription = adapters.get_user_subscription(user.email)
+
     subscription_renewal_date = (
         user_subscription.renewal_date.strftime("%d %b %Y")
         if user_subscription and user_subscription.renewal_date
-        else (user_subscription.created_at + timedelta(days=7)).strftime("%d %b %Y")
+        else None
```
debanjum, line 1485: Should this continue to be set to `user_subscription.created_at + timedelta(days=7)).strftime("%d %b %Y")` if `renewal_date` is unset? The downstream dependencies e.g common/auth.ts, settings/page.tsx on web app) seem to expect `subscription_renewal_date` to be a (non-nullable) string

src/khoj/routers/helpers.py
```diff
@@ -1477,10 +1478,16 @@ def get_user_config(user: KhojUser, request: Request, is_detailed: bool = False)
...
+    )
+    subscription_enabled_trial_at = (
+        user_subscription.enabled_trial_at.strftime("%d %b %Y")
+        if user_subscription and user_subscription.enabled_trial_at
+        else None
```
debanjum, line 1490: Should we set `subscription_enabled_trial_at` to `user_subscription.created_at + timedelta(days=7)).strftime("%d %b %Y")` if `enabled_trial_at` is unset for some reason? The downstream dependencies (e.g common/auth.ts, settings/page.tsx on web app) seem to expect it to be a (non-nullable) string

debanjum (approved):
Looks good

src/khoj/routers/helpers.py
sabaimran: `subscription_enabled_trial_at` is important to keep state of whether the trial has been used or not, so I don't think we should fill in an arbitrary value like that. I'll make the type nullable on the TypeScript object instead.

src/khoj/routers/helpers.py
sabaimran: I don't think so. It's a bit misleading, because it suggests the user is enrolled in something when they're not. This should correctly indicate that the user has no active enrollment at the time.

src/khoj/database/adapters/__init__.py
sabaimran: Good catch!

PR #946, opened 2024-10-21
Remove user customized search model

sabaimran:
- Use a single standard search model across the server. There's diminishing benefits for having multiple user-customizable search models. 
- We may want to add server-level customization for specific tasks
- Store the search model used to generate a given entry on the `Entry` object
- Remove user-facing APIs and view
- Add a management command for migrating the default search model on the server

In a future PR (after running the migration), we'll also remove the `UserSearchModelConfig`


src/khoj/search_type/text_search.py
```diff
@@ -110,7 +110,7 @@ async def query(
     file_type = search_type_to_embeddings_type[type.value]
 
     query = raw_query
-    search_model = await sync_to_async(get_user_search_model_or_default)(user)
+    search_model = await sync_to_async(get_default_search_model)()
```
debanjum, line 113: Wouldn't this be problematic for users currently using a non-default search model until the management command is successfully run to convert their entry embeddings to the default search model?

It maybe easier to avoid this issue if:
- The entry embeddings were converted via a migration to the server default search model from user default search model 
- Or switch to using server search model in two steps.
   1. Migrate the user model and embeddings to the default search model (instead of user search model) via management command
   2. Now switch to always use the server search model instead of user search model

sabaimran: Fair point. This would fail for those users become the vector dimensions would be mismatched. Updated to continue using the user setting until migration script is run.

src/khoj/routers/api.py
```diff
@@ -149,7 +149,7 @@ async def execute_search(
     encoded_asymmetric_query = None
     if t != SearchType.Image:
         with timer("Encoding query took", logger=logger):
-            search_model = await sync_to_async(get_user_search_model_or_default)(user)
+            search_model = await sync_to_async(get_default_search_model)()
```
debanjum, line 152: should this be updated to use `get_user_default_search_model` as well to encode user query for now?

debanjum (approved):

src/khoj/database/adapters/__init__.py
```diff
@@ -446,12 +446,27 @@ async def set_user_github_config(user: KhojUser, pat_token: str, repos: list):
...
+        user_search_model = UserSearchModelConfig.objects.filter(user=user).first()
+        if user_search_model:
+            return user_search_model.setting
+
+    default_search_model = SearchModelConfig.objects.filter(name="default").first()
+
+    if default_search_model:
+        return default_search_model
     else:
         SearchModelConfig.objects.create()
```
debanjum, line 471: nit: This can be replaced with get_default_search_model to dedupe logic
```suggestion
    return get_default_search_model()
```

src/khoj/processor/content/text_to_entries.py
```diff
@@ -148,10 +148,10 @@ def update_embeddings(
                 hashes_to_process |= hashes_for_file - existing_entry_hashes
 
         embeddings = []
+        model = get_default_search_model()
```
debanjum, line 151: Should index also use get_user_default_search_model? It'll avoid mismatched embedding model usage for generating embedding for user data and user query

```suggestion
        model = get_user_default_search_model()
```

PR #947, opened 2024-10-22
Separate notes, online context from user message sent to chat models

debanjum:
## Overview
- Put context into separate user message before sending to chat model.
  This should improve model response quality and truncation logic in code
- Pass online context from chat history to chat model for response.
  This should improve response speed when previous online context can be reused
- Improve format of notes, online context passed to chat models in prompt.
  This should improve model response quality

## Details
The document, online search context are now passed as separate user
messages to chat model, instead of being added to the final user message.

This will improve

- Models ability to differentiate data from user query.
  That should improve response quality and reduce prompt injection
  probability

- Make truncation logic simpler and more robust
  When context window hit, can simply pop messages to auto truncate
  context in order of context, user, assistant message for each
  conversation turn in history until reach current user query

  The complex, brittle logic to extract user query from context in
  last user message isn't required.

### DIFF
--- a/documentation/docs/advanced/admin.md
+++ b/documentation/docs/advanced/admin.md
@@ -1,6 +1,8 @@
 # Admin Panel
 > Describes the Khoj settings configurable via the admin panel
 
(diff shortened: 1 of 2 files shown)


sabaimran:
Amazing. This is groundbreaking work 🚀.

PR #948, opened 2024-10-23
Give Vision to Anthropic models in Khoj

debanjum:
### Major
- Give Vision to Anthropic models in Khoj

### Minor
- Reuse logic to format messages for chat with anthropic models
- Make the get image from url function more versatile and reusable
- Encourage output mode chat actor to output only json and nothing else

### DIFF
--- a/src/khoj/processor/conversation/anthropic/utils.py
+++ b/src/khoj/processor/conversation/anthropic/utils.py
@@ -141,13 +141,15 @@ def format_messages_for_anthropic(messages: list[ChatMessage], system_prompt=Non
     for message in messages:
         if isinstance(message.content, list):
             content = []
-            # Sort the content as preferred if text comes after images
+            # Sort the content. Anthropic models prefer that text comes after images.
             message.content.sort(key=lambda x: 0 if x["type"] == "image_url" else 1)
             for idx, part in enumerate(message.content):
                 if part["type"] == "text":
                     content.append({"type": "text", "text": part["text"]})
                 elif part["type"] == "image_url":
-                    b64_image, media_type = get_image_from_url(part["image_url"]["url"], type="b64")
+                    image = get_image_from_url(part["image_url"]["url"], type="b64")
+                    # Prefix each image with text block enumerating the image number
+                    # This helps the model reference the image in its response. Recommended by Anthropic
                     content.extend(
                         [
                             {
@@ -156,7 +158,7 @@ def format_messages_for_anthropic(messages: list[ChatMessage], system_prompt=Non
                             },
                             {
                                 "type": "image",
-                                "source": {"type": "base64", "media_type": media_type, "data": b64_image},
+                                "source": {"type": "base64", "media_type": image.type, "data": image.content},
                             },
                         ]
                     )
--- a/src/khoj/processor/conversation/anthropic/utils.py
+++ b/src/khoj/processor/conversation/anthropic/utils.py
@@ -3,6 +3,7 @@ from threading import Thread
 from typing import Dict, List
 
 import anthropic
+from langchain.schema import ChatMessage
 from tenacity import (
     before_sleep_log,
     retry,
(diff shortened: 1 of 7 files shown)


src/khoj/processor/conversation/anthropic/utils.py
```diff
@@ -115,3 +117,49 @@ def anthropic_llm_thread(
...
+    if len(messages) == 1:
+        messages[0].role = "user"
+    elif len(messages) > 1 and messages[0].role == "assistant":
+        messages = messages[1:]
+
+    # Convert image urls to base64 encoded images in Anthropic message format
+    for message in messages:
+        if isinstance(message.content, list):
+            content = []
+            # Sort the content as preferred if text comes after images
```
sabaimran, line 144: What does this mean? If I have any image in the chat history, would it be brought to the front?

sabaimran: Update, within a given chat message, a text message giving the image a "label" is followed by the actual image.

sabaimran: https://docs.anthropic.com/en/docs/build-with-claude/vision

src/khoj/processor/conversation/utils.py
```diff
@@ -306,3 +315,22 @@ def reciprocal_conversation_to_chatml(message_pair):
 def remove_json_codeblock(response: str):
     """Remove any markdown json codeblock formatting if present. Useful for non schema enforceable models"""
     return response.removeprefix("```json").removesuffix("```")
+
+
+def get_image_from_url(image_url: str, type="pil"):
```
sabaimran, line 320: Can we use some basic typing here?

```
class ImageWithType:
    encoded_data: Any
    type: str
```

Otherwise code like `get_image_from_url(item["image_url"]["url"])[0]` gets pretty confusing to follow. Would become `get_image_from_url(item["image_url"]["url"]).encoded_data` with proposed changes.

sabaimran: Would need to update references in `src/khoj/processor/conversation/google/utils.py` and `src/khoj/processor/conversation/anthropic/utils.py`.

src/khoj/routers/helpers.py
```diff
@@ -825,10 +825,13 @@ async def send_message_to_model_wrapper(
     conversation_config: ChatModelOptions = await ConversationAdapters.aget_default_conversation_config(user)
     vision_available = conversation_config.vision_enabled
     if not vision_available and query_images:
+        logger.warning(f"Vision is not enabled for default model: {conversation_config.chat_model}.")
```
sabaimran, line 828: IMO default vision model needs to also become an option for `ServerChatSettings`.

src/khoj/processor/conversation/anthropic/utils.py
```diff
@@ -115,3 +117,49 @@ def anthropic_llm_thread(
...
+            message.content.sort(key=lambda x: 0 if x["type"] == "image_url" else 1)
+            for idx, part in enumerate(message.content):
+                if part["type"] == "text":
+                    content.append({"type": "text", "text": part["text"]})
+                elif part["type"] == "image_url":
+                    b64_image, media_type = get_image_from_url(part["image_url"]["url"], type="b64")
+                    content.extend(
+                        [
+                            {
```
sabaimran, line 153: Just noting here that this image + label recommendation is shared here: https://docs.anthropic.com/en/docs/build-with-claude/vision#example-multiple-images

> In situations where there are multiple images, introduce each image with Image 1: and Image 2: and so on. You don’t need newlines between images or between images and the prompt.
>
> Ask Claude to describe the differences between multiple images.

src/khoj/routers/helpers.py
debanjum: Good idea given we have multiple vision model types supported now. But yeah, maybe better as a follow-up to allow setting the default vision model to use. Current behavior of using first added vision model is confusing

src/khoj/processor/conversation/anthropic/utils.py
debanjum: Thanks for sharing references to these! I've added a comment as well to explain this

src/khoj/processor/conversation/anthropic/utils.py
debanjum: Also updated that comment to be a bit less confusing

src/khoj/processor/conversation/utils.py
debanjum: Fixed

PR #950, opened 2024-10-23
Separate notes, online context from user message sent to chat models

debanjum:
## Overview
- Put context into separate user message before sending to chat model.
  This should improve model response quality and truncation logic in code
- Pass online context from chat history to chat model for response.
  This should improve response speed when previous online context can be reused
- Improve format of notes, online context passed to chat models in prompt.
  This should improve model response quality

## Details
The document, online search context are now passed as separate user
messages to chat model, instead of being added to the final user message.

This will improve
- Models ability to differentiate data from user query.
  That should improve response quality and reduce prompt injection
  probability
- Make truncation logic simpler and more robust
  When context window hit, can simply pop messages to auto truncate
  context in order of context, user, assistant message for each
  conversation turn in history until reach current user query

  The complex, brittle logic to extract user query from context in
  last user message isn't required.


src/khoj/processor/conversation/prompts.py
```diff
@@ -118,6 +118,7 @@
 Ask crisp follow-up questions to get additional context, when a helpful response cannot be provided from the provided notes or past conversations.
 
 User's Notes:
+-----
```
sabaimran, line 121: Let's see how this changes response quality?

src/khoj/processor/conversation/utils.py
```diff
@@ -186,21 +188,27 @@ def generate_chatml_messages_with_context(
...
-        if chat["by"] == "khoj" and "excalidraw" in chat["intent"].get("type"):
-            message_content = chat.get("intent").get("inferred-queries")[0] + message_notes
-        else:
-            message_content = chat["message"] + message_notes
+        message_context = ""
+        if chat["by"] == "khoj" and "excalidraw" in chat["intent"].get("type", ""):
+            message_context += chat.get("intent").get("inferred-queries")[0]
+        if not is_none_or_empty(chat.get("context")):
+            references = "\n\n".join(
+                {f"# File: {item['file']}\n## {item['compiled']}\n" for item in chat.get("context") or []}
```
sabaimran, line 196: `compiled` helps us with retrieving the correct embeddings during search. Should we use `raw` for injecting context in the query?

src/khoj/processor/conversation/utils.py
```diff
@@ -186,21 +188,27 @@ def generate_chatml_messages_with_context(
...
-        message_content = construct_structured_message(message_content, chat.get("images"), model_type, vision_enabled)
+        role = "user" if chat["by"] == "you" else "assistant"
+        message_content = construct_structured_message(chat["message"], chat.get("images"), model_type, vision_enabled)
 
         reconstructed_message = ChatMessage(content=message_content, role=role)
-
         chatml_messages.insert(0, reconstructed_message)
 
-        if len(chatml_messages) >= 2 * lookback_turns:
+        if len(chatml_messages) >= 3 * lookback_turns:
```
sabaimran, line 211: We won't necessarily have `3` elements for each `turn`, right? Only if/when there's context attached to a given turn, otherwise `2`. Should be fine as an approximation.

src/khoj/processor/conversation/utils.py
```diff
@@ -186,21 +188,27 @@ def generate_chatml_messages_with_context(
...
+            message_context += chat.get("intent").get("inferred-queries")[0]
+        if not is_none_or_empty(chat.get("context")):
+            references = "\n\n".join(
+                {f"# File: {item['file']}\n## {item['compiled']}\n" for item in chat.get("context") or []}
+            )
+            message_context += f"{prompts.notes_conversation.format(references=references)}\n\n"
+        if not is_none_or_empty(chat.get("onlineContext")):
+            message_context += f"{prompts.online_search_conversation.format(online_results=chat.get('onlineContext'))}"
+        if not is_none_or_empty(message_context):
+            reconstructed_context_message = ChatMessage(content=message_context, role="context")
```
sabaimran, line 202: Is `context` a valid role for all models? Do we need this to fit into the assistant/user binary? 🤔

sabaimran: Seems like `context` isn't being used and then being reset below. Let's just set it to `user` here straight away to avoid confusion?

sabaimran (approved):
Sweet. Just a couple of clean-up related comments, but this definitely does seem neater.

PR #957, opened 2024-11-05
Support including file attachments in the chat message

sabaimran:
Now that models have much larger context windows, we can reasonably include full texts of certain files in the messages. Do this when an explicit file filter is set in a conversation. Do so in a separate user message in order to mitigate any confusion in the operation.

Pipe the relevant attached_files context through all methods calling into models.

This breaks certain prior behaviors. We will no longer automatically be processing/generating embeddings on the backend and adding documents to the "brain". You'll have to go to settings and go through the upload documents flow there in order to add docs to the brain (i.e., have search include them during question / response).


src/khoj/processor/conversation/anthropic/anthropic_chat.py
```diff
@@ -84,7 +85,12 @@ def extract_questions_anthropic(
         vision_enabled=vision_enabled,
     )
 
-    messages = [ChatMessage(content=prompt, role="user")]
+    messages = []
+
+    if attached_files:
+        messages.append(ChatMessage(content=attached_files, role="user"))
```
debanjum, line 91: Should attached_files be passed into `construct_structured_message` similar to how we handle `query_images`?

src/interface/web/app/common/chatFunctions.ts
```diff
@@ -267,6 +267,75 @@ export async function createNewConversation(slug: string) {
...
+                    let fileExtension = fileName.split(".").pop();
+                    if (fileExtension === "org") {
+                        fileType = "text/org";
+                    } else if (fileExtension === "md") {
+                        fileType = "text/markdown";
+                    } else if (fileExtension === "txt") {
+                        fileType = "text/plain";
+                    } else if (fileExtension === "html") {
+                        fileType = "text/html";
+                    } else if (fileExtension === "pdf") {
```
debanjum, line 295: Should we allow docx here as well?

src/khoj/routers/api_chat.py
```diff
@@ -707,6 +745,8 @@ def collect_telemetry():
         ## Extract Document References
         compiled_references: List[Any] = []
         inferred_queries: List[Any] = []
+        file_filters = conversation.file_filters if conversation and conversation.file_filters else []
```
debanjum, line 748: Where are these `file_filters` used?

src/khoj/routers/api_chat.py
```diff
@@ -1081,6 +1127,8 @@ def collect_telemetry():
                 query_images=uploaded_images,
                 tracer=tracer,
                 train_of_thought=train_of_thought,
+                attached_file_context=attached_file_context,
+                raw_attached_files=raw_attached_files,
```
debanjum, line 1131: Why are both attached_file_context and raw_attached_files required? And if they are required, can they grouped together into single object (e.g a `query_files` TypedDict, Class) instead?

src/khoj/utils/rawconfig.py
```diff
@@ -138,6 +138,38 @@ class SearchResponse(ConfigBase):
...
+    title: Optional[str] = None
+    conversation_id: Optional[str] = None
+    turn_id: Optional[str] = None
+    city: Optional[str] = None
+    region: Optional[str] = None
+    country: Optional[str] = None
+    country_code: Optional[str] = None
+    timezone: Optional[str] = None
+    images: Optional[list[str]] = None
+    files: Optional[list[FileAttachment]] = []
```
debanjum, line 169: Should this default to None (like `images` property above it)?

debanjum (commented):
Left some comments. But looking forward to being able to easily chat with files without indexing it in my knowledge base!

src/khoj/processor/content/pdf/pdf_to_entries.py
```diff
@@ -109,3 +91,30 @@ def convert_pdf_entries_to_maps(parsed_entries: List[str], entry_to_file_map) ->
...
 
         return entries
+
+    @staticmethod
+    def extract_text(pdf_file):
+        """Extract text from specified PDF files"""
+        try:
+            # Write the PDF file to a temporary file, as it is stored in byte format in the pdf_file object and the PDF Loader expects a file path
+            timestamp_now = datetime.utcnow().timestamp()
+            tmp_file = f"tmp_pdf_file_{timestamp_now}.pdf"
```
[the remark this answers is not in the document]
sabaimran: Maybe a combo of both? Can still have old files from processing that aren't torn down correctly or parallel processing for a single user.

src/khoj/routers/api_chat.py
sabaimran: They're used by the summarization and notes lookup portions. They essentially limit what parts of the "brain"/knowledge base the conversation has access to.

src/khoj/routers/api_chat.py
sabaimran: `attached_file_context` is packaged neatly for consumption by the LLM, uniformly sent down the evaluation tree. `raw_attached_files` are used for storing the raw data in the conversation history.

src/khoj/utils/rawconfig.py
sabaimran: No, this causes some errors in parts of the code that try to iterate over the list. IMO makes sense to make it an empty list, makes the code simpler.

src/khoj/processor/conversation/anthropic/anthropic_chat.py
```diff
@@ -149,6 +155,7 @@ def converse_anthropic(
     query_images: Optional[list[str]] = None,
     vision_available: bool = False,
     tracer: dict = {},
+    attached_files: str = None,
```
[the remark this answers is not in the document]
sabaimran: I liked `attached_files` because it gives a connotation for the relationship between the messages and the documents in primitives analogous to how we think about messaging genrally. Query is a bit generic.

debanjum (approved):
Left 2 minor comments on PR. But otherwise changes look good to merge

src/interface/web/app/common/iconUtils.tsx
```diff
@@ -216,6 +217,13 @@ export function getIconForSlashCommand(command: string, customClassName: string
     return <ArrowRight className={className} />;
 }
 
+export function getIconFromFileType(fileType: string) {
+    if (fileType == "pdf") {
+        return <FilePdf className="w-6 h-6 text-muted-foreground inline-flex mr-1" />;
+    }
+    return <File className="w-6 h-6 text-muted-foreground inline-flex mr-1" />;
+}
+
```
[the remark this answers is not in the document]
sabaimran: Ah yeah, good point. Updating.

PR #958, opened 2024-11-05
Do not CRUD on entries, files & conversations in DB for null user

debanjum:
Increase defense-in-depth by reducing paths to create, read, update or delete entries, files and conversations in DB when user is unset.

### DIFF
--- a/src/khoj/database/adapters/__init__.py
+++ b/src/khoj/database/adapters/__init__.py
@@ -8,7 +8,17 @@ import secrets
 import sys
 from datetime import date, datetime, timedelta, timezone
 from enum import Enum
-from typing import Callable, Iterable, List, Optional, Type
+from functools import wraps
+from typing import (
+    Any,
+    Callable,
+    Coroutine,
+    Iterable,
+    List,
+    Optional,
+    ParamSpec,
+    TypeVar,
+)
 
 import cron_descriptor
 from apscheduler.job import Job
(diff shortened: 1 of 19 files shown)


Freddy-del (commented):
Uh

src/khoj/database/adapters/__init__.py
```diff
@@ -416,6 +416,11 @@ def get_all_users() -> BaseManager[KhojUser]:
     return KhojUser.objects.all()
 
 
+def check_valid_user(user: KhojUser | None):
```
sabaimran, line 419: Just fruit for thought, it might be worth making this a decorator? Seems like a neater way to handle top-level function validation that we're standardizing across the code. Non-blocking, up to you.

src/khoj/database/adapters/__init__.py
```diff
@@ -953,15 +966,15 @@ def has_valid_openai_conversation_config():
...
+        if not config or user is None:
             return None
         new_config = await UserConversationConfig.objects.aupdate_or_create(user=user, defaults={"setting": config})
         return new_config
 
     @staticmethod
     async def aset_user_voice_model(user: KhojUser, model_id: str):
         config = await VoiceModelOption.objects.filter(model_id=model_id).afirst()
-        if not config:
+        if not config or user is None:
```
sabaimran, line 977: Why not use the `check_valid_user` method here and above?

src/khoj/database/adapters/__init__.py
```diff
@@ -874,6 +883,7 @@ async def acreate_conversation_session(
     def create_conversation_session(
         user: KhojUser, client_application: ClientApplication = None, agent_slug: str = None, title: str = None
     ):
+        check_valid_user(user)
```
sabaimran, line 886: We should probably make sure these errors are handled gracefully by relevant clients as well. E.g., in APIs, return an error with detail about what went wrong?

src/khoj/processor/content/github/github_to_entries.py
```diff
@@ -48,7 +48,7 @@ def wait_for_rate_limit_reset(response, func, *args, **kwargs):
         else:
             return
 
-    def process(self, files: dict[str, str] = None, user: KhojUser = None, regenerate: bool = False) -> Tuple[int, int]:
+    def process(self, files: dict[str, str], user: KhojUser, regenerate: bool = False) -> Tuple[int, int]:
```
sabaimran, line 51: Let's verifying all these functions still work as expected with `--anonymous_mode`. There might be broken assumptions with this change, though I'm not aware of anything off the top of my head.

src/khoj/search_type/text_search.py
```diff
@@ -208,7 +208,7 @@ def setup(
     text_to_entries: Type[TextToEntries],
     files: dict[str, str],
     regenerate: bool,
-    user: KhojUser = None,
+    user: KhojUser,
```
sabaimran, line 211: Did you want to move the `user` parameter to the top, similar to the other functions?

sabaimran (commented):
Just a bit of clean-up suggestions, but this will be dope in assuring certain expected patterns.

src/khoj/search_type/text_search.py
Freddy-del: Yes

src/khoj/database/adapters/__init__.py
Freddy-del: Yes

src/khoj/database/adapters/__init__.py
debanjum: Maybe we can do that as a follow-up? Sounds like a lot of work to track all the API endpoints these DB adapter functions are being called and put those calls into a try/catch block

For now clients will automatically see the API call fail with 500 internal server error. Server logs will show the issue. Which is a good enough start for now IMO?

src/khoj/search_type/text_search.py
debanjum: I'll let this be as is for now, don't want to shuffle too much code around as part of this change

src/khoj/processor/content/github/github_to_entries.py
debanjum: Yeah, tested basic chat, search, agents and automation. Tested anon and auth modes. Don't see any issues yet

PR #960, opened 2024-11-06
Generate and analyze documents with Khoj using the Code sandbox

debanjum:
- Allow passing user files as input into code sandbox for analysis
- Update prompt to give more example of complex, multi-line code
- Simplify logic for model. Run one program at a time, 
  instead of allowing model to run multiple programs in parallel
- Show Code generated charts and docs in Reference pane of web app and make them downloaded


src/khoj/processor/tools/run_code.py
```diff
@@ -79,13 +90,12 @@ async def generate_python_code(
     context: str,
     location_data: LocationData,
     user: KhojUser,
-    query_images: List[str] = None,
+    query_images: list[str] = None,
     agent: Agent = None,
     tracer: dict = {},
-) -> List[str]:
+) -> tuple[str, list[str], list[str]]:
```
sabaimran, line 96: Tuple strikes again 😅. You already know my feedback here.

src/khoj/processor/tools/run_code.py
```diff
@@ -57,15 +57,26 @@ async def run_code(
...
 
+    # Prepare Input Data
+    input_data = []
+    user_input_files: List[FileObject] = []
+    for input_file in input_files:
+        user_input_files += await FileObjectAdapters.async_get_file_objects_by_name(user, input_file)
+    for f in user_input_files:
+        input_data.append(
+            {
+                "filename": os.path.basename(f.file_name),
```
sabaimran, line 68: Where is this input_data format specification defined? Might be helpful just to include a link to the doc here in case someone tries changing it in the future.

sabaimran (approved):
Sweet, looking forward to improving the code execution reliability & depth.

src/khoj/processor/tools/run_code.py
debanjum: Added reference to that in execute_sandbox function below

src/khoj/processor/tools/run_code.py
debanjum: Fixed!

PR #998, opened 2024-12-04
Rename OpenAIProcessorConversationConfig DB model to more apt AiModelApi

debanjum:
The db model name had drifted from what it is being used for, a general chat api provider.
Given it now supports other chat api providers like anthropic and google chat models apart from apart from openai api based models. This change renames the db model and updates the docs to remove this confusion.


documentation/docs/advanced/ollama.mdx
```diff
@@ -60,7 +60,7 @@ Restart your Khoj server after first run or update to the settings below to ensu
       ```bash
       ollama pull llama3.1
       ```
-   3. Create a new [OpenAI Processor Conversation Config](http://localhost:42110/server/admin/database/openaiprocessorconversationconfig/add) on your Khoj admin panel
```
sabaimran, line 63: Looks like we should also update `litellm.md`, `lmstudio.md` and `admin.md` similarly.

src/khoj/database/models/__init__.py
```diff
@@ -79,7 +79,7 @@ class Type(models.TextChoices):
     enabled_trial_at = models.DateTimeField(null=True, default=None, blank=True)
 
 
-class OpenAIProcessorConversationConfig(BaseModel):
+class AiModelApi(BaseModel):
```
sabaimran, line 82: It seems to me that the `ModelType` field in `ChatModelOptions` is better served here than in the other. The AI model API will only fit one of those types, and it's better defined as part of the provider, rather than the chat model. Might be a helpful bit of refactor to include with this change, but may also be out of scope.

sabaimran (approved):
Just a little bit of clean-up, but looks good! Makes sense to evolve the naming.

documentation/docs/advanced/ollama.mdx
debanjum: Ah, good catch. Fixed those docs

PR #1000, opened 2024-12-05
Fix broken doc links to query filters

codeluggage:
Fixes an outdated query filters link in the emacs docs.


debanjum:
Ah, thanks for catching that! That was referencing the old docs links

PR #1011, opened 2024-12-17
Allow research mode and other conversation commands in automations

debanjum:
Previously we couldn't use research mode or use other slash commands 
in automated tasks.

This change separates determining if a chat query is triggered via
automated task from (other) conversation commands to run the query with.

This unlocks the ability to enable research mode in automations 
apart from other variations like  /image or /diagram etc

### DIFF
--- a/src/khoj/routers/helpers.py
+++ b/src/khoj/routers/helpers.py
@@ -231,7 +231,7 @@ def get_next_url(request: Request) -> str:
     return urljoin(str(request.base_url).rstrip("/"), next_path)
 
 
-def get_conversation_command(query: str, any_references: bool = False) -> ConversationCommand:
+def get_conversation_command(query: str) -> ConversationCommand:
     if query.startswith("/notes"):
         return ConversationCommand.Notes
     elif query.startswith("/help"):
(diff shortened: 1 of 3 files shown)


src/khoj/routers/helpers.py
```diff
@@ -231,34 +231,39 @@ def get_next_url(request: Request) -> str:
...
-def get_conversation_command(query: str, any_references: bool = False) -> ConversationCommand:
+def get_conversation_commands(query: str) -> tuple[list[ConversationCommand], bool]:
+    # Intialize variables
+    commands: list[ConversationCommand] = []
+    is_automated_task = False
+
+    # Automated Task can be mixed with other conversation commands
+    if query.startswith("/automated_task"):
+        is_automated_task = True
+        query = query.replace("/automated_task", "").lstrip()
```
sabaimran, line 242: Can we move this logic of removing the `automated_task` prefix and creating the `bool` outside of this function? I'd like this function to remain a source of truth and protected from higher level business logic.

debanjum: Makes sense. Extracted the logic to handle the `automated_task` conversation command in chat API endpoint and reverted most of the `get_conversation_command` function to previous state

PR #1014, opened 2024-12-20
Improve the agent management UX

sabaimran:
- Add support for seeing all steps of the agent modification flow via tabs at the top of the modal
- Separate knowledge base & tool selection into two separate parts

### DIFF
--- a/src/interface/web/app/components/agentCard/agentCard.tsx
+++ b/src/interface/web/app/components/agentCard/agentCard.tsx
@@ -91,6 +91,8 @@ import { Popover, PopoverContent, PopoverTrigger } from "@/components/ui/popover
 import ShareLink from "@/app/components/shareLink/shareLink";
 import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from "@/components/ui/tooltip";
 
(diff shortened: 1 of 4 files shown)


src/interface/web/components/ui/tabs.tsx
debanjum, line 1: If all required fields aren't populated the save fails but i don't know which required fields I didn't fill. need to go through each tab to check that. Maybe good to add some something in the tab names to indicate that the fields in that tab need to be filled/fixed (e.g a `*` suffix so Customize => Customize *) or the tab becomes red

src/interface/web/app/components/agentCard/agentCard.tsx
debanjum, line 1: Not sure if we want to redirect not logged in users to sign-in before trying to edit a card. Right now I can fill all fields in an agent card but can't create it until logged in. This maybe annoying

debanjum (approved):
The tab changes look good as give better orientation of what all needs to be filled to create an agent instead of the previous Previous > Next with no breadcrumb like help

src/interface/web/components/ui/tabs.tsx
sabaimran: Nice feedback! Included both in #1015

PR #1016, opened 2024-12-22
Improve Automation Flexibility and Automation Email Format

debanjum:
- Format AI response to send in automation email
- Let Khoj infer chat query based on user automation query
- Decide if automation emails should be sent based on response
  - Fix the `to_notify_or_not` decider AI
  - Ask reason before decision to improve to_notify decider AI
- Show error message on web app when fail to create/update automation

### DIFF
--- a/src/interface/web/app/automations/page.tsx
+++ b/src/interface/web/app/automations/page.tsx
@@ -167,68 +167,68 @@ const timestamp = Date.now();
 const suggestedAutomationsMetadata: AutomationsData[] = [
     {
         subject: "Weekly Newsletter",
-        query_to_run:
+        scheduling_request:
             "/research Compile a message including: 1. A recap of news from last week 2. An at-home workout I can do before work 3. A quote to inspire me for the week ahead",
         schedule: "9AM every Monday",
         next: "Next run at 9AM on Monday",
         crontime: "0 9 * * 1",
         id: timestamp,
-        scheduling_request: "",
-    },
-    {
-        subject: "Daily Bedtime Story",
-        query_to_run:
-            "Compose a bedtime story that a five-year-old might enjoy. It should not exceed five paragraphs. Appeal to the imagination, but weave in learnings.",
-        schedule: "9PM every night",
-        next: "Next run at 9PM today",
-        crontime: "0 21 * * *",
-        id: timestamp + 1,
-        scheduling_request: "",
+        query_to_run: "",
     },
     {
         subject: "Front Page of Hacker News",
-        query_to_run:
+        scheduling_request:
             "/research Summarize the top 5 posts from https://news.ycombinator.com/best and share them with me, including links",
         schedule: "9PM on every Wednesday",
         next: "Next run at 9PM on Wednesday",
         crontime: "0 21 * * 3",
         id: timestamp + 2,
-        scheduling_request: "",
+        query_to_run: "",
     },
     {
         subject: "Market Summary",
-        query_to_run:
+        scheduling_request:
             "/research Get the market summary for today and share it with me. Focus on tech stocks and the S&P 500.",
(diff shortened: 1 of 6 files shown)


src/interface/web/app/automations/page.tsx
```diff
@@ -164,68 +164,68 @@ const timestamp = Date.now();
 const suggestedAutomationsMetadata: AutomationsData[] = [
     {
         subject: "Weekly Newsletter",
-        query_to_run:
+        scheduling_request:
```
sabaimran, line 167: Ah, good find. `scheduling_request` is more like the user query, and `query_to_run` is more like an inferred query. The change in language should help construct more helpful automations.

src/khoj/routers/api.py
```diff
@@ -686,16 +686,20 @@ def edit_job(
 
     # Check, get automation to edit
     try:
-        automation: Job = AutomationAdapters.get_automation(user, automation_id)
+        automation: Job = await AutomationAdapters.aget_automation(user, automation_id)
```
sabaimran, line 689: Why not write a corresponding synchronous version of the `aget_automation` function and wrap the `schedule_query` function in a async to sync method so that this API can remain synchronous?

src/khoj/processor/conversation/prompts.py
```diff
@@ -1137,6 +1137,26 @@
 )
 
 
+automation_format_prompt = PromptTemplate.from_template(
```
sabaimran, line 1140: Nice, this is a good idea. Excited to try it out.

src/khoj/routers/api.py
```diff
@@ -746,11 +752,11 @@ def edit_job(
...
     trigger = CronTrigger.from_crontab(crontime, user_timezone)
     if automation.trigger != trigger:
-        automation.reschedule(trigger=trigger)
+        await sync_to_async(automation.reschedule)(trigger=trigger)
 
     # Collate info about the updated user automation
-    automation = AutomationAdapters.get_automation(user, automation.id)
-    automation_info = AutomationAdapters.get_automation_metadata(user, automation)
+    automation = await AutomationAdapters.aget_automation(user, automation.id)
+    automation_info = await sync_to_async(AutomationAdapters.get_automation_metadata)(user, automation)
```
sabaimran, line 759: In the `get_automation_metadata` function, we had some explicit regex operations to remove the `/automated_task` prefix from `query_to_run`. Let's remove that as it's not required anymore.

src/khoj/routers/api.py
```diff
@@ -665,7 +665,7 @@ def trigger_manual_job(
 
 @api.put("/automation", response_class=Response)
 @requires(["authenticated"])
-def edit_job(
+async def edit_job(
```
sabaimran, line 668: When the automation update/create encounters an error, it gets stuck on the client-side in the `Saving...` state. We should handle the error case and inform the user it didn't work.

src/khoj/routers/helpers.py
```diff
@@ -553,7 +553,7 @@ async def generate_online_subqueries(
 
 async def schedule_query(
     q: str, conversation_history: dict, user: KhojUser, query_images: List[str] = None, tracer: dict = {}
-) -> Tuple[str, ...]:
+) -> Tuple[str, str, str]:
```
sabaimran, line 556: Let's use `clean_json` instead of `json.loads` here for more graceful interpretation of the results.

sabaimran (commented):
Nice! Excited to try out these improvements.

src/khoj/routers/api.py
debanjum: Good point. Updated the automation cards to show error toast when update/create for automation fails

sabaimran (approved):

Issue #1019, opened 2024-12-29
[FIX] v1.33.0 can not chat in stream mode 


thinker007:
## Describe the bug
A clear and concise description of what the bug is. Please include what you were expecting to happen vs. what actually happened.
/khoj/src/khoj/processor/conversation/openai/utils.py:103 in             │                
                           │ completion_with_backoff                                                   │                
                           │                                                                           │                
                           │   100 │   # Calculate cost of chat                                        │                
                           │   101 │   input_tokens = chunk.usage.prompt_tokens if hasattr(chunk, "usa │                
                           │       else 0                                                              │                
                           │   102 │   output_tokens = chunk.usage.completion_tokens if hasattr(chunk, │                
                           │       chunk.usage else 0                                                  │                
                           │ ❱ 103 │   cost = chunk.usage.model_extra.get("estimated_cost") or 0  # Es │                
                           │       returned by DeepInfra API                                           │                
                           │   104 │   tracer["usage"] = get_chat_usage_metrics(model_name, input_toke │                
                           │       tracer.get("usage"), cost)                                          │                
                           │   105 │                                                                   │                
                           │   106 │   # Save conversation trace                                       │                
                           ╰───────────────────────────────────────────────────────────────────────────╯                
                           AttributeError: 'NoneType' object has no attribute 'model_extra'        
## To Reproduce

After model setting ,just chat no return 

## Platform
- Server:
    - [ ] Cloud-Hosted (https://app.khoj.dev)
    - [ ] Self-Hosted Docker
    - [ ] Self-Hosted Python package
    - [ ] Self-Hosted source code
- Client:
    - [ ] Obsidian
    - [ ] Emacs
    - [ ] Desktop app
    - [ ] Web browser
    - [ ] WhatsApp
- OS:
    - [ ] Windows
    - [ ] macOS
    - [ ] Linux
    - [ ] Android
    - [ ] iOS

### If self-hosted
- Server Version [e.g. 1.0.1]:

thinker007:
AttributeError: 'NoneType' object has no attribute 'model_extra'  still exist

thinker007:
[15:21:31.254144] ERROR    khoj.processor.conversation.openai.utils: Error in llm_thread: 'NoneType' object has no attribute 'model_extra'         utils.py:227
                           ╭───────────────────────────────────────── Traceback (most recent call last) ─────────────────────────────────────────╮             
[...]

PR #1026, opened 2025-01-02
Handle reporting chat estimated cost when some fields unavailable

thinker007:
fix AttributeError: 'NoneType' object has no attribute 'model_extra'

https://github.com/khoj-ai/khoj/issues/1019

### DIFF
--- a/src/khoj/processor/conversation/openai/utils.py
+++ b/src/khoj/processor/conversation/openai/utils.py
@@ -101,8 +101,6 @@ def completion_with_backoff(
     input_tokens = chunk.usage.prompt_tokens if hasattr(chunk, "usage") and chunk.usage else 0
     output_tokens = chunk.usage.completion_tokens if hasattr(chunk, "usage") and chunk.usage else 0
     cost = chunk.usage.model_extra.get("estimated_cost", 0) if hasattr(chunk, "usage") and chunk.usage else 0  # Estimated costs returned by DeepInfra API
-    else:
-        cost = None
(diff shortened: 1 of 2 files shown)


src/khoj/processor/conversation/openai/utils.py
```diff
@@ -100,7 +100,10 @@ def completion_with_backoff(
     # Calculate cost of chat
     input_tokens = chunk.usage.prompt_tokens if hasattr(chunk, "usage") and chunk.usage else 0
     output_tokens = chunk.usage.completion_tokens if hasattr(chunk, "usage") and chunk.usage else 0
-    cost = chunk.usage.model_extra.get("estimated_cost") or 0  # Estimated costs returned by DeepInfra API
+    if chunk.usage:
```
sabaimran, line 103: Wouldn't we want to make sure that `chunk.usage.model_extra` is present, not just `chunk.usage`? Otherwise, you would have the same problem again as the code in the above lines assumes `chunk.usage`. For the error in the stack trace you've shared, the issue seems to specifically be due to the `model_extra` field.

```suggestion
    if chunk.usage.model_extra
```

Same for the change below.

sabaimran (changes requested):
Thanks for finding this issue and creating a fix!

src/khoj/processor/conversation/openai/utils.py
thinker007: the newest pull has fix this you can merge that .

debanjum (approved):
Looks good!

sabaimran (approved):

PR #1029, opened 2025-01-05
Improve docker-compose.yml

DPS0340:
## Improvements

- Remove host port mappings on dependencies
- Add 'restart: always'
- Add default network for lookup by docker dns
- Add extra_hosts for host.docker.internal lookup

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,12 +1,13 @@
 services:
   database:
     image: ankane/pgvector
-    ports:
-      - "5432:5432"
+    restart: always
     environment:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
       POSTGRES_DB: postgres
+    networks:
+      - default
     volumes:
       - khoj_db:/var/lib/postgresql/data/
     healthcheck:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,12 +1,13 @@
 services:
   database:
     image: ankane/pgvector
-    ports:
```
sabaimran, line 4: Let's keep the `ports` fields in all the subservices. It makes it easier to manage if/when the ports need to be updated.

sabaimran (changes requested):
This is great! Simplifies a lot of the settings.

Thank you 🎉

docker-compose.yml
DPS0340: @sabaimran 

I thought that exposing host port is easy to collide with other service of user. especially `5432`.
Is it better to change host port? like 42000 <= x <= 42999.

sabaimran: Actually you're right, this should be fine. There's no *inherent* reason why we need the services exposed on the host machine, though it's helpful for debugging.

docker-compose.yml
```diff
@@ -16,22 +17,25 @@ services:
...
-      - "42113:8080"
+    restart: always
+    networks:
+      - default
     volumes:
       - khoj_search:/etc/searxng
     environment:
-      - SEARXNG_BASE_URL=http://localhost:42113/
+      - SEARXNG_BASE_URL=http://0.0.0.0:8080/
```
sabaimran, line 31: Can we change this to `localhost`? Otherwise it can conflict with other services running at `8080` on the host machine.

```suggestion
      - SEARXNG_BASE_URL=http://localhost:8080/
```

DPS0340: @sabaimran I fixed it. :)

sabaimran (approved):

sabaimran:
Tested new `docker-compose.yml` on MacOS and Linux, verified that I'm able to start up the app and chat with models and conduct online search and execute code. Looks good!

@DPS0340 , feel free to share any context for additional testing you've done.

DPS0340:
@sabaimran 

I'm able to use khoj obsidian plugin with sync :)
Now, currently I don't have other considerations. Feel free to merge 😄 

## Screenshots

image

image

PR #1051, opened 2025-01-09
Support using Embeddings Model exposed via OpenAI (compatible) API

debanjum:
This change adds the ability to use OpenAI's embedding models or any embedding model exposed behind an OpenAI compatible API (like Ollama, LiteLLM, vLLM etc.). This allows using commercial embedding models to index your content with Khoj.

Khoj previously only supported HuggingFace embedding models running locally on device or via HuggingFaceW inference API endpoint.


src/khoj/processor/embeddings.py
```diff
@@ -71,23 +82,38 @@ def embed_with_api(self, docs):
...
+            return self.embeddings_model.encode(docs, **self.docs_encode_kwargs).tolist() if docs else []
+        elif self.inference_endpoint_type == SearchModelConfig.ApiType.HUGGINGFACE:
+            embed_with_api = self.embed_with_hf
+        elif self.inference_endpoint_type == SearchModelConfig.ApiType.OPENAI:
+            embed_with_api = self.embed_with_openai
+        else:
+            logger.warning(
+                f"Unsupported inference endpoint: {self.inference_endpoint_type}. Generating embeddings locally instead."
+            )
+            return self.embeddings_model.encode(docs, **self.docs_encode_kwargs).tolist()
```
sabaimran, line 107: Wouldn't this have to be a valid embeddings model for this to work? Presumably if something was botched during setup, then you can't assume it'll be a valid hf model local model.

src/khoj/database/models/__init__.py
```diff
@@ -481,6 +481,11 @@ class SearchModelConfig(DbBaseModel):
     class ModelType(models.TextChoices):
         TEXT = "text"
 
+    class ApiType(models.TextChoices):
```
sabaimran, line 484: Neat, nice and simple.

src/khoj/database/models/__init__.py
```diff
@@ -501,6 +506,10 @@ class ModelType(models.TextChoices):
     embeddings_inference_endpoint = models.CharField(max_length=200, default=None, null=True, blank=True)
```
sabaimran, line 506: If someone was using the Azure OpenAI endpoint, would they just set a different URL here?

sabaimran (commented):
Sweet! This should definitely help Khoj be more scale-ready for generating embeddings.

src/khoj/database/models/__init__.py
debanjum: Yeah, to use an OpenAI compatible proxy, they'd just need to set the `embeddings_inference_endpoint` to their openai proxies url. If it's an Azure OpenAI endpoint, it'll use the AzureOpenAI client to create embeddings (instead of the OpenAI client) like we do for chat now

src/khoj/processor/embeddings.py
debanjum: Yeah, I wasn't sure what to do here (fallback to reasonable defaults or raise an exception). This logic is from before this PR when we only supported HF models so if API not available it falls back to downloading and use the HF model locally. I'll let this be for now to not fail indexing if HF inference API not available. We can change logic later, if required?

jppaolim:
sounds amazing. But it doesn't seem to be actually indexing via the API. I used the MD file which is not yet merged, adding text-embedding-3-small to the name of biencoder, and inputing and API key. Shall I do something else ? How to make sure it works. ?

debanjum:
Hey could you follow [these instructions](https://docs.khoj.dev/features/search#setup-self-hosting). 

Two things to remember to do:
1. On the Admin panel, ensure only the Search Model Config for the OpenAI embeddings has the name `default`.
2. Restart your Khoj server before trying to index you md files

If it's still not working, it'd be good if you can share screenshots of your OpenAI Search Model admin config page and the top level search model config page from the admin panel that lists all the search model configs

jppaolim:
Studying the log, I had passed https://api.openai.com/v1/embeddings and not https://api.openai.com/v1 as endpoint. 
Now my issue it seems to embed correctly. Then I add to adjust the treshold, I didn't perform extensive tests, but 0,18 was too low. I put 0,5 and it works for  text-embedding-3-small. Further guidance will be welcome :)

Issue #1035, opened 2025-01-07
[FIX]Agents stop outputting "Generating a well-informed response"


MartinMayday:
## Describe the bug
A clear and concise description of what the bug is. Please include what you were expecting to happen vs. what actually happened.
I want to continue a conversation from some days ago. but agent doesn't actually output "Generating a well-informed response"
I've tried: 
- reloading browser, 
- used another browser,
- /notes please summarise last session 

## To Reproduce
Find conversation in sidebar
continue conversations (user)
assistant *thinks', never outputs "Generating a well-informed response"

## Screenshots
<img width="896" alt="Screenshot 2025-01-07 at 12 46 37" src="https://github.com/user-attachments/assets/535401ea-9feb-468a-bcac-b761105dbf9b" />

## Platform
- Server:
    - [ ] Cloud-Hosted (https://app.khoj.dev)
    - [ ] Self-Hosted Docker
    - [ ] Self-Hosted Python package
    - [ ] Self-Hosted source code
- Client:
    - [ ] Obsidian
    - [ ] Emacs
    - [ x] Desktop app
    - [ ] Web browser
    - [ ] WhatsApp
- OS:
    - [ ] Windows
    - [ x] macOS
    - [ ] Linux
    - [ ] Android
    - [ ] iOS

### If self-hosted
- Server Version [e.g. 1.0.1]:

## Additional context
is there a way to explore logs of conversations?
is there a way to see how many MB sync files are uploaded? - obsidian keeps notifying something like "sync failed due to no more space"

ProjectMoon:
I am seeing this happen with ollama due to a timeout on the Khoj side. ollama logs report connection aborted. But everything in the train of thought seems to work up until the last point. Maybe the prompt is so big and the client (Khoj) is not waiting long enough?

ProjectMoon:
Also, for me, this seems to be specific to research mode enabled.

debanjum:
Hey @ProjectMoon, thanks for sharing a detailed report. 

It maybe a short timeout window or truncation issue on the Khoj side.

Let me try reproduce this issue using Khoj with a small context model on Ollama. 

Research mode does result in longer context requirements and hence being more likely to trigger the timeout issue.

If it's a timeout issue, increasing the timeout on decorator to openai api calling funcs in openai/utils.py should fix it (can use different timeouts for openai vs local models)

ProjectMoon:
[...]

PR #1065, opened 2025-01-10
Use the configured OpenAI Base URL for Automations

arcuru:
This change makes Automations (and possibly other entrypoints) use the configured OpenAI-compatible server if that has been set. Without this change it tries to use the hardcoded OpenAI provider.

All the other calls in this file use a similar method to pass in the base URL.

**I have not been able to manually test this because the docker build is taking an extremely long time to build locally.**

### DIFF
--- a/src/khoj/routers/helpers.py
+++ b/src/khoj/routers/helpers.py
@@ -1135,6 +1135,7 @@ def send_message_to_model_wrapper_sync(
 
     elif chat_model.model_type == ChatModel.ModelType.OPENAI:
         api_key = chat_model.ai_model_api.api_key
+        api_base_url = chat_model.ai_model_api.api_base_url
         truncated_messages = generate_chatml_messages_with_context(
             user_message=message,
             system_message=system_message,
@@ -1149,6 +1150,7 @@ def send_message_to_model_wrapper_sync(
         openai_response = send_message_to_model(
             messages=truncated_messages,
             api_key=api_key,
+            api_base_url=api_base_url,
             model=chat_model_name,
             response_type=response_type,
             tracer=tracer,


arcuru:
I was able to build this locally for testing and I hit at least one uncovered issue related to sending the LLM calls to an OpenAI Compatible Server, which may need to be fixed as well for this scenario.

I think the root of the uncovered problem is from these logs:

```
server-1    | [18:46:28.889975] DEBUG    khoj.processor.conversation.utils:       utils.py:537
server-1    |                            Fallback to default chat model
server-1    |                            tokenizer: gpt-4o.
server-1    |                            Configure tokenizer for model:
server-1    |                            local/chat/phi4:14b in Khoj settings to
server-1    |                            improve context stuffing.
```

It appears that the name "gpt-4o" is hardcoded as the default tokenizer, and this call ended up failing even when I created a "gpt-4o" chat model in khoj, adding a "gpt-4o" endpoint on my LiteLLM server, and even configuring my OpenAI key for the gpt-4o model directly. Setting a tokenizer for that model in server/admin/ had no effect.

There are further errors in the logs after this, but they seem to be coming from chunking while running the query to create the automation. While it's good that this is hitting a failure immediately I really hope you're not just feeding the output of the selected GUI options into the LLM to convert it to a cron job template. I don't see why an external tokenizer in this scenario is even necessary, that prompt to setup the automation must be tiny and well within my configured model context size.

TBH, I've hit so many GUI and setup bugs while trying to selfhost Khoj that I'm not going to spend more time on it. I filed the bugs that probably impact your cloud service as well so you can fix for your paying customers.

debanjum:
Hey @arcuru , unfortunate you've hit multiple issues in setting up automations for self-hosted Khoj. I wonder if you're hitting the same timeout issues as https://github.com/khoj-ai/khoj/issues/1035#issuecomment-2575913199. Anyway let me look into self-hosted Khoj + OpenAI API proxy + Automation setups and see what can be improved for a less annoying experience.

Until then I've answered some of your concerns below:

> It appears that the name "gpt-4o" is hardcoded as the default tokenizer, and this call ended up failing even when I created a "gpt-4o" chat model in khoj, adding a "gpt-4o" endpoint on my LiteLLM server, and even configuring my OpenAI key for the gpt-4o model directly. Setting a tokenizer for that model in server/admin/ had no effect.

This isn't an error just a debug log (notice the `[18:46:28.889975] DEBUG` prefix) that calculating prompt size is going to be less accurate. We fallback to use the gpt-4o tokenizer if can't infer the tokenizer to use for the current chat model, so in scenarios like this where the OpenAI API is being used with non-OpenAI models. 

This only actually becomes a problem if the max prompt size set for the chat model is close the actual max prompt size of the model and your chat history starts hitting that limit. You could set max prompt size = 10K for phi4, given it has a 14K context window (which is small by modern standards).

Not sure why you're seeing this when using chat model to gpt-4o though.

> I really hope you're not just feeding the output of the selected GUI options into the LLM to convert it to a cron job template.

The LLM isn't used to set crontime when you've explicitly specified it via GUI. We used to have the LLM set the crontime job schedule when we allowed creating an automation directly from the chat (e.g you tell Khoj in chat "Share synthetic biology news every tuesday at 9pm").

The LLM is used to convert your original query into an automation query and email subject. So if you say "Notify me if it's going to rain today" it converts it into a chat query: "Is it going to rain today?" and email subject: "Rain Notification". The results of the chat query: "Is it going to rain today?" is compared against your original query: "Notify me ..." to decide if a notification email should be sent or not.

Nonetheless, appreciate the PR and feedback!

AlipAbdullah:
Hey please inform all of them that I can opt-out all of the developers and
any community and whatever is link with my google.com domain anytime I
wanted too.

Sincerely yours,
Alip Clinton/Alip Abdullah
Google AI
Ownership Manager

PR #1073, opened 2025-01-11
Allow browsing and discovery of knowlege base in the search page

sabaimran:
Currently, it's rather opaque and difficult to substantially browse through the uploaded knowledge base. Effectively, you can only do this through the small file modal in the settings page. 

Update to include all indexed files in the search page for viewing & deletion. Function to delete all files is still in the settings page.

Add a migration that associates file objects with `entry`s using a foreign key. Add a migration command that deletes dangling fileobjects.


src/khoj/routers/api_content.py
```diff
@@ -325,6 +330,65 @@ def get_content_types(request: Request, client: Optional[str] = None):
     return list(configured_content_types & all_content_types)
 
 
+@api_content.get("/all", response_model=Dict[str, str])
+@requires(["authenticated"])
+async def get_all_content(request: Request, client: Optional[str] = None, truncated: Optional[bool] = True):
```
sabaimran, line 335: Add pagination to avoid heavy data streaming with larger knowledge bases.

src/interface/web/app/search/page.tsx
```diff
@@ -153,41 +292,200 @@ function focusNote(note: SearchResult) {
...
+                    Add Documents
+                </Button>
+            </DialogTrigger>
+            <DialogContent>
+                <DialogHeader>
+                    <DialogTitle>Build Your Knowledge Base</DialogTitle>
+                    <DialogDescription>
+                        Add your files to supercharge Khoj`&apos;`s AI with your knowledge. Get
+                        instant, personalized answers powered by your own documents and data.
+                    </DialogDescription>
```
debanjum, line 386: 1. Why wrap the apostrophe in backticks?
2. An alternate description for the dialog

```suggestion
                    <DialogTitle>Build Your Knowledge Base</DialogTitle>
                    <DialogDescription>
                        Add documents to your Khoj knowledge base.
                        Quickly search and get personalized answers from your documents.
                    </DialogDescription>
```

src/interface/web/app/search/page.tsx
```diff
@@ -59,24 +98,131 @@ function getNoteTypeIcon(source: string) {
...
+                                    <Trash className="h-4 w-4" />
+                                    <span className="text-xs">
+                                        {isDeleting
+                                            ? "Deleting..."
+                                            : "Delete"}
+                                    </span>
+                                </Button>
+                            </DropdownMenuItem>
+                        </DropdownMenuContent>
+                    </DropdownMenu>
```
debanjum, line 208: Why have a dropdown for a single action, delete? Maybe the delete trash icon can be visible instead of the dropdown 3 dot menu on every card?

src/interface/web/app/settings/page.tsx
```diff
@@ -1162,18 +853,13 @@ export default function SettingsView() {
...
-                                    {isManageFilesModalOpen && (
-                                        <ManageFilesModal
-                                            onClose={() => setIsManageFilesModalOpen(false)}
-                                        />
-                                    )}
                                     <div className="section grid gap-8">
                                         <div className="text-2xl">Content</div>
                                         <div className="cards flex flex-wrap gap-16">
                                             <Card id="computer" className={cardClassName}>
                                                 <CardHeader className="flex flex-row text-2xl">
```
debanjum, line 860: Should the text size be xl of the card title? Other cards seem to be `text-xl` too

```suggestion
                                                <CardHeader className="flex flex-row text-xl">
```

src/khoj/database/migrations/0079_entry_file_object.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if not entries:
+            break
+
+        processed_entry_ids.update([entry.id for entry in entries])
+
+        entries_to_update = []
+        for entry in entries:
+            try:
+                file_object = file_objects_map.get((entry.user_id, entry.file_path))
```
debanjum, line 38: Should we also delete orphaned file objects as part of this migration?

debanjum (approved):
Left some comments but otherwise the changes look good!

src/khoj/routers/api_content.py
debanjum: This seems to have improved rendering while testing with the 2k+ documents from the FRAMES eval knowledge base

src/interface/web/app/search/page.tsx
sabaimran: Because the trash can icon takes up more space and adds visual clutter. Three dots is much more pleasant than a rubbish bin.

Also, I'd expect it to be used infrequently 🙂 .

src/interface/web/app/search/page.tsx
sabaimran: > Why wrap the apostrophe in backticks?

It's escaping it so that it can be rendered correctly at export time.

Suggestion sounds good, will incorporate.

src/khoj/database/migrations/0079_entry_file_object.py
sabaimran: I'm inclined to say no. It's better to avoid complex data migrations taking place alongside database migrations to reduce any risk of long jobs. I'm on the fence even about running `migrate_entry_objects` here, except that self-hosted users would get left out of the new setup if it didn't run automatically, which could break server-side expectations.

The consequence of not deleting orphaned file objects automatically is that self-hosted users might see those dangling files in their `/search` page. I think that's okay in tradeoff, to avoid auto-running large data deletion operations live.

Here are some alternatives:

1. We could add the management command instructions for a Docker-friendly environment in the release notes.
2. We could re-use the older migrations stuff we'd built in `cli.py` when everything was based off of the `config.yml`, alternatively, but it'll have to be upgraded to align with how things work now.

sabaimran: Let's continue the discussion after merge. We can add/modify as needed before release.

Issue #1020, opened 2024-12-30
Add Delete chat message action to the Obsidian Plugin


debanjum:
The Khoj API supports deleting specific chat messages. The web app already uses that. But the Khoj Obsidian plugin doesn't use it yet.

This feature should 
- Add a Delete message button to each chat message footer in the Obsidian plugin's chat view
- Clicking it should 
  1. Call the Khoj server API to delete the Khoj or user message
  2. Refresh the Khoj Obsidian UI with the conversation turn (= Khoj + associated user message) deleted

samhoooo:
Hi @debanjum , if no one is working on it,  I can pick this up. Thanks

PR #1076, opened 2025-01-11
Add Delete chat message action to the Obsidian Plugin

samhoooo:
Closes https://github.com/khoj-ai/khoj/issues/1020

### Description

The Khoj API supports deleting specific chat messages. The web app already uses that. But the Khoj Obsidian plugin doesn't use it yet.

This feature should

Add a Delete message button to each chat message footer in the Obsidian plugin's chat view
Clicking it should
Call the Khoj server API to delete the Khoj or user message
Refresh the Khoj Obsidian UI with the conversation turn (= Khoj + associated user message) deleted

### Changes

- Added ability to delete individual messages in chat conversations
- Each message now has a delete button alongside copy and paste actions
- Deleting a message removes both the user's message and Khoj's response (the full conversation turn)
- Added confirmation dialog before deletion to prevent accidental deletions
- Messages are deleted both from the UI and the Khoj server API
- System messages cannot be deleted (e.g. setup hint)

### Technical Details

- Added new `deleteMessage` method to handle message deletion logic
- Added DELETE API endpoint call to `/api/chat/conversation/message`
- Delete button uses the same icon as in web version (`trash-2`)
- Added new boolean parameter `isSystemMessage` to `renderMessage` method
- Show delete button only if it is not a system message

### Testing

- [x] Test deleting user messages
- [x] Test deleting Khoj responses
- [x]  Verify both messages in a turn are removed
- [x]  Verify deletion persists after page refresh
- [x]  Verify no delete button on setup hint
- [x] Verify no delete button on error message
- [x] Verify no delete button on conversation starter
- [x] Test error cases (network failure, etc)

### Screensort

**Happy path:**

https://github.com/user-attachments/assets/7de9626f-e207-436c-be59-bb17f0a97d73

**Setup hint cannot be deleted:**

image


src/interface/obsidian/src/chat_view.ts
```diff
@@ -1487,4 +1538,49 @@ export class KhojChatView extends KhojPaneView {
...
         }
     }
+
+    // Add this new method to handle message deletion
+    async deleteMessage(messageEl: HTMLElement) {
+        const chatBodyEl = this.contentEl.getElementsByClassName("khoj-chat-body")[0] as HTMLElement;
+        const conversationId = chatBodyEl.dataset.conversationId;
+
+        // Get the turn_id from the message's data-meta attribute
+        const turnId = messageEl.getAttribute("data-meta");
```
sabaimran, line 1548: Hey @samhoooo ! This data field doesn't actually contain the `turnId` field that you need, so it just results in no change in the backend for the chat history. The `turnId` isn't currently being stored in the rendered message element. You would need to add some additional handling to store that, both in the streamed messages and in the chat history messages.  

For streamed messages, you can update `processMessageChunk` and store the data in `chatMessageState`. The `chunk.type` in that case would be `metadata`. Then include it in the message element, similar to how `data-meta` is stored. I think `finalizeChatBodyResponse` would be a good place to add the data to the element.

You can give it the attribute name `data-turnId` and set the value to the incoming `turnId`.

For messages loaded from history, it would be similar, except you'd add the `turnId` field to `renderMessageWithReferences`. Just update the logic in the `renderMessage` function to include the `turnId` as well.

Then finally, in this `deleteMesage` function, you'll `getAttribute("data-turnId")`, instead of `data-meta`.

Does that make sense? Let me know if you need help with any of it. Thanks for opening the PR.

samhoooo: Thanks @sabaimran! Really appreciate your detailed explaination.

I have updated the PR to include the `turnId` handling as described, covered both streamed and messages loaded from history. I have also tested with my own account, delete some messages from Obsidian, and able to see chat history is updated on web.

Let me know if there is anything else I should watch out, thanks again for your guidance.

PR #1078, opened 2025-01-12
Improve Obsidian Sync for Large Vaults

debanjum:
- Batch sync files by size to try not exceed API request payload size limits
- Fix force sync of large vaults from Obsidian
- Add API endpoint to delete all indexed files by file type


src/khoj/routers/api_content.py
```diff
@@ -325,6 +325,29 @@ def get_content_types(request: Request, client: Optional[str] = None):
     return list(configured_content_types & all_content_types)
 
 
+@api_content.delete("/{content_type}", status_code=200)
```
sabaimran, line 333: This API spec conflicts with out other `delete` endpoint, as below:

```
@api_content.get("/{content_source}", response_model=List[str])
```

I'd suggest we just use the existing deletion endpoint, and send it with type `computer` rather than specifying file types. Ideally, we should differentiate with `Obsidian` here, but that might be a bigger discussion.

sabaimran (changes requested):
Mostly looks good, tested it out locally. Just need to sort out the issue with `delete` APIs.

src/khoj/routers/api_content.py
debanjum: Great catch. This has been fixed by renaming the 2 conflicting delete APIs to `/api/content/type/{content_type}` and `/api/content/source/{content_source}` now

PR #1087, opened 2025-01-16
Support online search using Google Search API

debanjum:
Add official Google Search API as an online search provider.
We currently support Serper.dev and Searxng as online search providers.

### DIFF
--- a/src/khoj/processor/tools/online_search.py
+++ b/src/khoj/processor/tools/online_search.py
@@ -30,6 +30,8 @@ from khoj.utils.rawconfig import LocationData
 
 logger = logging.getLogger(__name__)
 
+GOOGLE_SEARCH_API_KEY = os.getenv("GOOGLE_SEARCH_API_KEY")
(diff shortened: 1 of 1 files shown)


src/khoj/processor/tools/online_search.py
```diff
@@ -96,18 +98,24 @@ async def search_online(
...
+        search_func = search_with_google
+    elif SERPER_DEV_API_KEY:
+        search_engine = "Serper"
         search_func = search_with_serper
     elif JINA_API_KEY:
+        search_engine = "Jina"
         search_func = search_with_jina
     else:
-        search_func = search_with_searxng
+        search_engine = "Jina"
```
sabaimran, line 111: This doesn't work without an API key anymore, so let's skip if none of the options are configured?

sabaimran (approved):

src/khoj/processor/tools/online_search.py
debanjum: Oh yeah, fixed to fallback to searxng by default (not Jina)

PR #1089, opened 2025-01-17
Add Basic Auth Support

JustinMissmahl:
Adds username/password authentication mode for self-hosted instances that want to manage users manually without OAuth providers.

Changes:

Backend:
- basic_auth.py: Login endpoint and metadata endpoint returning {"google": null}
- state.py: Added controlled_access_mode flag
- main.py: CLI flag and env var handling
- configure.py: Router mounting and auth middleware

Frontend (loginPrompt.tsx):
- Basic auth detection from metadata response
- Username/password form with loading/error states
- Works in both mobile/desktop views
- Preserves existing OAuth/email flows

Usage:
1. Enable controlled access mode via:
   - CLI: --controlled-access
   - ENV: KHOJ_CONTROLLED_ACCESS=true

2. Initial admin credentials:
   - Username: admin@example.com
   - Password: admin

3. Managing Users:
   - Login as admin at http://localhost:42110/server/admin
   - Use Django admin interface to:
     * Create new users
     * Set passwords * Manage permissions
   - Users can then login with their credentials at http://localhost:42110

Security Notes:
- Default admin should be changed after first login
- Each user needs to be manually created by admin
- No self-registration - controlled environment

Testing:
```bash
docker compose up  # Uses KHOJ_CONTROLLED_ACCESS=true
```
1. Visit http://localhost:42110 - should show username/password form
2. Login with admin credentials
3. Create test user in admin panel
4. Test login with new user

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ khoj_assistant.egg-info
 # ---
 # npm
 node_modules
+package-lock.json
 
 # Don't include the compiled obsidian main.js file in the repo.
 # They should be uploaded to GitHub releases instead.
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,22 +26,40 @@ WORKDIR /app
 COPY pyproject.toml .
 COPY README.md .
 ARG VERSION=0.0.0
-# use the pre-built llama-cpp-python, torch cpu wheel
+
+# Install dependencies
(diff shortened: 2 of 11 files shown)


debanjum (changes requested):
Thanks for creating a PR to add basic auth support! I've left some comments to simplify the code. Let me know if you have any questions/

debanjum:
Closing this PR for now. It has gone stale and comments haven't been resolved.

PR #1141, opened 2025-03-27
Use embedded postgres db to simplify self-hosted setup

debanjum:
Use [pgserver](https://github.com/orm011/pgserver/) python package as an embedded postgres db, installed directly as a khoj python package dependency.

This significantly simplifies self-hosting setup with just a `pip install khoj`. No need to also install postgres separately.

Still use standard postgres server for multi-user, production use-cases.

### DIFF
--- a/src/khoj/app/settings.py
+++ b/src/khoj/app/settings.py
@@ -10,6 +10,8 @@ For the full list of settings and their values, see
 https://docs.djangoproject.com/en/4.2/ref/settings/
 """
 
+import atexit
+import logging
 import os
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


src/khoj/app/settings.py
```diff
@@ -112,20 +114,95 @@
...
+if USE_PGSERVER:
+    # Set up logging for pgserver
+    logger = logging.getLogger("pgserver_django")
+    logger.setLevel(logging.INFO)
+    if not logger.handlers:
+        handler = logging.StreamHandler()
+        handler.setFormatter(logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s"))
+        logger.addHandler(handler)
+
+    WSGI_APPLICATION = "app.wsgi.application"
```
sabaimran, line 141: Why do you need to do WSGI with `pg_server`? It doesn't work with the async server? This might have interesting consequences on the open connections issue, namely that it might be a mitigation.

src/khoj/app/settings.py
```diff
@@ -112,20 +114,95 @@
...
+        PGSERVER_DATA_DIR = os.path.join(BASE_DIR, "pgserver_data")
+        os.makedirs(PGSERVER_DATA_DIR, exist_ok=True)
+
+        logger.info(f"Initializing pgserver with data directory: {PGSERVER_DATA_DIR}")
+
+        # Start server
+        PGSERVER_INSTANCE = pgserver.get_server(PGSERVER_DATA_DIR)
+
+        # Create pgvector extension, if not already exists
+        PGSERVER_INSTANCE.psql("CREATE EXTENSION IF NOT EXISTS vector;")
```
sabaimran, line 156: Do you know if pg_server comes with the `vector` extension pre-baked? We have it on our machines because it came with postgres.app on macos. Just a quick thing to verify. Maybe we can also distribute the extension in the Khoj app if needed(?).

src/khoj/app/settings.py
```diff
@@ -112,20 +114,95 @@
...
+            lines = db_list_output.strip().split("\n")
+            if len(lines) > 2:  # Has at least header, data, and footer
+                for line in lines[2:-1]:  # Skip header and footer
+                    db_name = line.strip()
+                    if db_name:
+                        db_list.append(db_name)
+
+        if DB_NAME not in db_list:
+            logger.info(f"Creating database: {DB_NAME}")
+            PGSERVER_INSTANCE.psql(f"CREATE DATABASE {DB_NAME};")
```
sabaimran, line 174: This logic could be simpler I believe --

```suggestion
# Create database if needed
db_exists = PGSERVER_INSTANCE.psql(f"SELECT 1 FROM pg_database WHERE datname = '{DB_NAME}';")
if not db_exists:
    logger.info(f"Creating database: {DB_NAME}")
    PGSERVER_INSTANCE.psql(f"CREATE DATABASE {DB_NAME};")
```

sabaimran (commented):
This is going to be SO COOL for self-hosted users. No more Docker setup since Postgres install may not be an issue. Good find! 🎉

I'm trying to set it up locally but it seems to be hanging. Leaving code-only review until I can get it working locally.

EDIT: got it working. Sweet!

src/khoj/app/settings.py
debanjum: Previously thought the pg server is not working as an ASGI app. But on checking again, it is not required to be a WSGI app. So reverted that portion

src/khoj/app/settings.py
debanjum: Good suggestion. I've simplified the code

sabaimran (approved):

PR #1151, opened 2025-04-05
Allow Model switching based on User Tier

debanjum:
### Overview
Enable free tier users to chat with any AI model made available on free tier 
of production deployments like [Khoj cloud](https://app.khoj.dev).

Previously model switching was completely disabled for users on free tier.

### Details
- Track price tier of each Chat, Speech, Image, Voice AI model in DB
- Update API to allow free tier users to switch between free models
- Update web app to allow model switching on agent creation, settings
  chat page (via right side pane), even for free tier users.


src/interface/web/app/common/modelSelector.tsx
```diff
@@ -228,8 +230,9 @@ function ModelItem({ model, isSelected, onSelect, onPeek }: ModelItemProps) {
             onSelect={onSelect}
             ref={ref}
             className="data-[selected=true]:bg-muted data-[selected=true]:text-secondary-foreground"
+            disabled={!isActive && model.tier !== "free"}
         >
-            {model.name}
+            {model.name} {model.tier === "standard" && <span className="text-green-500 ml-2">(Pro)</span>}
```
sabaimran, line 235: Let's be consistent in naming?

```suggestion
            {model.name} {model.tier === "standard" && <span className="text-green-500 ml-2">(Futurist)</span>}
```

src/interface/web/app/components/agentCard/agentCard.tsx
```diff
@@ -788,9 +784,18 @@ export function AgentModificationForm(props: AgentModificationFormProps) {
...
+                                                    !props.isSubscribed &&
+                                                    modelOption.tier !== "free"
+                                                }
                                             >
                                                 <div className="flex items-center space-x-2">
-                                                    {modelOption.name}
+                                                    {modelOption.name}{" "}
+                                                    {modelOption.tier === "standard" && (
+                                                        <span className="text-green-500 ml-2">
+                                                            (Pro)
```
sabaimran, line 796: Same comment as above

src/interface/web/app/settings/page.tsx
```diff
@@ -111,8 +112,9 @@ const DropdownComponent: React.FC<DropdownComponentProps> = ({ items, selected,
                                 <DropdownMenuRadioItem
                                     key={item.id.toString()}
                                     value={item.id.toString()}
+                                    disabled={!isActive && item.tier !== "free"}
                                 >
-                                    {item.name}
+                                    {item.name} {item.tier === "standard" && <span className="text-green-500 ml-2">(Pro)</span>}
```
sabaimran, line 117: Same comment as above

src/khoj/database/adapters/__init__.py
```diff
@@ -1115,30 +1115,36 @@ async def aset_user_voice_model(user: KhojUser, model_id: str):
...
     def get_chat_model(user: KhojUser):
-        subscribed = is_user_subscribed(user)
-        if not subscribed:
-            return ConversationAdapters.get_default_chat_model(user)
         config = UserConversationConfig.objects.filter(user=user).first()
         if config:
             return config.setting
+        subscribed = is_user_subscribed(user)
+        if not subscribed:
+            return ConversationAdapters.get_default_chat_model(user)
```
sabaimran, line 1123: This check should come before the config check, in case the user is not subscribed & had an older setting that's using an upgraded chat model. We can also use this function to opportunistically correct that mismatch.

src/khoj/database/models/__init__.py
```diff
@@ -202,11 +202,16 @@ class ModelType(models.TextChoices):
         ANTHROPIC = "anthropic"
         GOOGLE = "google"
 
+    class PriceTier(models.TextChoices):
```
sabaimran, line 205: We can define the `PriceTier` class outside of the data model class so that it can be reused across data models.

src/khoj/database/models/__init__.py
```diff
@@ -452,6 +462,17 @@ class ServerChatSettings(DbBaseModel):
         WebScraper, on_delete=models.CASCADE, default=None, null=True, blank=True, related_name="web_scraper"
     )
 
+    def clean(self):
+        error = {}
```
sabaimran, line 466: Rather than throwing an error, let's correct it automatically? If the admin ends up here, it's most likely a mistake on their part, so we can gracefully correct it imo.

sabaimran (commented):
nice! will be awesome to extend model switching to unsubbed users.

PR #1168, opened 2025-04-22
Add support for storing, maintaining user memories in the backend

sabaimran:
# Memory
A major component of useful AI systems is adaptation to the user context. This is a major reason why we've prioritized uploading of knowledge bases. The next steps in this direction is dynamically updating state of the user as more conversations take place.

We introduce a new paradigm for maintaining medium and long term memories in Khoj. The scope of a conversation can be thought of as short term memory. 

Medium term memory extends to the past week.

Long term memory extends to anytime in the past, where a search query results in a match.

Add support for viewing and managing the memories with user action in the settings page as well.

TODO: Hook up the memories to the query context given to the LLM.


debanjum (approved):
Changes seem ready to merge. We've added per agent memory scoping, admin+user ability to disable the memory system

PR #1202, opened 2025-06-22
allow setting the directory for the embedded postgres db with PGSERVER_DATA_DIR

petergaultney:
I was playing around with this, and ended up blowing away some data by destroying my `.venv`. 

It seems to me that it would be useful to be able to be explicit about where the embedded database should live - as well as where it _does_ live (via the info log), when not specifying.


debanjum:
Very reasonable suggestions. Changes look good!

PR #1216, opened 2025-08-19
Fix to create chat with custom agents from obsidian

Fh26697:
The function createNewConversation is never called with the agentSlug specified so its always opening a new Conversation with the Base Agent

### DIFF
--- a/src/interface/obsidian/src/chat_view.ts
+++ b/src/interface/obsidian/src/chat_view.ts
@@ -981,7 +981,7 @@ export class KhojChatView extends KhojPaneView {
                 this.currentAgent = agentSlug || null;
 
                 // Update agent selector to reflect current agent
-                const agentSelect = this.contentEl.querySelector('.khoj-agent-select') as HTMLSelectElement;
+                const agentSelect = this.contentEl.querySelector('.khoj-header-agent-select') as HTMLSelectElement;
                 if (agentSelect) {
                     agentSelect.value = this.currentAgent || '';
                 }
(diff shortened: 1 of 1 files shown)


src/interface/obsidian/src/chat_view.ts
```diff
@@ -962,7 +962,9 @@ export class KhojChatView extends KhojPaneView {
             let endpoint = `${this.setting.khojUrl}/api/chat/sessions`;
             if (agentSlug) {
                 endpoint += `?agent_slug=${encodeURIComponent(agentSlug)}`;
-            }
+            } else if (this.currentAgent)  {
+				endpoint += `?agent_slug=${encodeURIComponent(this.currentAgent)}`;
+			}
```
debanjum, line 967: Let's fallback to set agentSlug to this.currentAgent here? This will automatically handle setting the agent in the agent dropdown when the chat is first created.

Currently it is unset due to `this.currentAgent = agentSlug || null` and `agentSelect.value = this.currentAgent || ""` logic later


```suggestion
            let endpoint = `${this.setting.khojUrl}/api/chat/sessions`;
            agentSlug = agentSlug || this.currentAgent;
            if (agentSlug) {
                endpoint += `?agent_slug=${encodeURIComponent(agentSlug)}`;
            }
```

debanjum (commented):
Thanks for submitting a fix for this! Agent switching from obsidian chat works now.

There was an existing bug around populating the current agent in the agent dropdown when a new conversation is created with a non-default agent.  I'll clean that and some related functionality up and merge this pull request by end of day today.

src/interface/obsidian/src/chat_view.ts
debanjum: Fixed

PR #1221, opened 2025-09-08
Better Obsidian sync

hjamet:
### **feat(obsidian): Enhance Sync Experience with Progress Bars and Bug Fixes**

This pull request significantly improves the content synchronization experience for Obsidian users by fixing a critical bug and introducing new UI elements for better feedback and monitoring.

The previous implementation could fail with `403 Forbidden` errors when syncing a large number of files due to server-side rate limiting. This update addresses that issue and provides users with clear, real-time feedback on storage usage and sync progress.

---
### Key Changes

* **🐛 Bug Fix: Sync Robustness (Resolves 403 Errors)**
    The core sync logic in `updateContentIndex` has been refactored. Instead of sending all file updates at once, the process is now chunked. Files are grouped by type (markdown, pdf, image) and uploaded in sequential batches. This respects server rate limits and ensures that large vaults can be indexed reliably without triggering `403` errors.

* **✨ New Feature: Cloud Storage Usage Bar**
    A progress bar has been added to the settings page to display cloud storage usage.
    * **Total Limit**: The storage limit (**10 MB** for free, **500 MB** for premium) is now reliably determined by the `is_active` flag returned from the `/api/v1/user` endpoint, eliminating fragile client-side heuristics.
    * **Used Space**: The used space is calculated via a **client-side estimation** of all files configured for synchronization. This provides a clear and immediate indicator of the vault's storage footprint.

* **✨ New Feature: Real-time Sync Progress Bar**
    When a manual sync is triggered via the "Force Sync" button, a second progress bar now appears, providing real-time feedback on the operation.
    * It displays the number of files processed against the total number of files to be indexed or deleted.
    * This is implemented using a **callback mechanism** (`onProgress`) to cleanly communicate progress from the sync logic (`utils.ts`) to the UI (`settings.ts`) without coupling them.

* **🔄 UX Improvement: Auto-refresh After Sync**
    The Cloud Storage Usage bar is now automatically refreshed upon the completion of a "Force Sync". This ensures the user immediately sees the updated storage estimation without needing to reopen the settings panel.

* **🛠️ DX Improvement: esbuild Development Workflow**
    The `esbuild.config.mjs` has been enhanced to automatically copy the built `main.js` and `manifest.json` files to a specified Obsidian plugin directory during development, streamlining the testing and iteration cycle.

---
### Visuals

image

### DIFF
--- a/src/interface/obsidian/src/api.ts
+++ b/src/interface/obsidian/src/api.ts
@@ -1,0 +1,30 @@
+export async function deleteContentByType(khojUrl: string, khojApiKey: string, contentType: string): Promise<void> {
+    // Deletes all content of a given type on Khoj server for Obsidian client
+    const response = await fetch(`${khojUrl}/api/content/type/${contentType}?client=obsidian`, {
+        method: 'DELETE',
+        headers: khojApiKey ? { 'Authorization': `Bearer ${khojApiKey}` } : {},
+    });
+    if (!response.ok) {
+        const text = await response.text().catch(() => '');
+        throw new Error(`Failed to delete content type ${contentType}: ${response.status} ${text}`);
+    }
+}
+
+export async function uploadContentBatch(khojUrl: string, khojApiKey: string, method: 'PUT' | 'PATCH', files: { blob: Blob, path: string }[]): Promise<string> {
+    // Uploads a batch of files to Khoj content endpoint
+    const formData = new FormData();
+    files.forEach(fileItem => { formData.append('files', fileItem.blob, fileItem.path); });
+
+    const response = await fetch(`${khojUrl}/api/content?client=obsidian`, {
+        method: method,
+        headers: khojApiKey ? { 'Authorization': `Bearer ${khojApiKey}` } : {},
+        body: formData,
+    });
+
+    if (!response.ok) {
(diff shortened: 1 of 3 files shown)


chillu:
This is great, but appears to introduce an issue: When Force Sync is used with multiple batches (e.g., >50 files), all previous entries are deleted.

**The Problem:**
- All batches use `PUT` requests (since `regenerate=true`), which means each batch triggers `regenerate=True` on the server
- Each batch with `regenerate=True` deletes **all** entries for `file_type + file_source` at the start
- This causes Batch 2+ to delete entries that Batch 1 just added

**Example:**
- Batch 1 (markdown files 1-50): PUT → deletes all markdown entries → adds 50 files
- Batch 2 (markdown files 51-100): PUT → deletes all markdown entries **again** (including Batch 1's entries!) → adds 50 files

**Additional Issue:**
The deletion logic queries existing entries without filtering by `file_source`, which can cause cross-client interference (e.g., Obsidian sync deleting Desktop client entries with the same file paths). See https://github.com/khoj-ai/khoj/issues/1239

**Suggested Fix:**
Skip regeneration deletion when entries from previous batches exist (since the client already pre-deletes everything before sending batches)

PR #1223, opened 2025-09-17
[DRAFT]: Tauri addition

mdaiter:
Not even remotely ready for merging

@debanjum , core stats: desktop app that loads immediately, with cross-platform bindings, and no Chrome bundle with the desktop app. Super quick.

Probably...90% parity.

Feel free to use any part of this that's useful!


debanjum:
Hi @mdaiter, thanks for the PR, it was a good read. But as discussed earlier, we wouldn't be able to merge this as we don't currently have the bandwidth to migrate the Khoj desktop app to Tauri. So closing the PR for now

PR #1235, opened 2025-12-23
Enable excluding folders to sync from obsidian plugin settings

borismus:
- Add excludeFolders field to KhojSetting interface
- Rename 'Sync Folders' to 'Include Folders' for clarity
- Add 'Exclude Folders' UI section with folder picker
- Filter out excluded folders during content sync
- Show file counts when syncing (X of Y files)
- Prevent excluding root folder

This allows users to exclude specific directories (e.g., Inbox, Highlights) from being indexed, while the existing Include Folders acts as a whitelist.


debanjum:
Thanks for the clean code for a useful feature!

PR #1241, opened 2026-01-05
fix(obsidian): limit number of files to upload

kamil-rudnicki:
I had problems with uploading over 2500 notes in single batch. API returned 400 right away.


kamil-rudnicki:
in the end the problem is with particular note/notes; I have to dig dipper why it hangs

debanjum:
PR #1221 to improve batched file syncing got merged a few days ago. It'll be released in a few days. Hopefully that improves the syncing of large knowledge bases

PR #1243, opened 2026-01-08
Add 'obsidian' EntrySource to distinguish plugin-imported content

### LINKED ISSUE #1239: Add granular 'obsidian' EntrySource to distinguish plugin-imported content from other sources
Currently, all content imported via the Obsidian plugin is marked with `EntrySource = "computer"`, which is also used for other types of imports (such as desktop client or manual ingestion). This lack of granularity in the source attribution means:
- It's impossible to specifically target only Obsidian-imported content for actions (e.g. bulk deletion) from the UI or API; 
- All "computer" entries are lumped together, and deleting Obsidian content inadvertently affects unrelated entries.

For example, in [`src/khoj/routers/api_content.py`](https://github.com/khoj-ai/khoj/blob/master/src/khoj/routers/api_content.py#L381), the `EntrySource` type only allows `"computer"`, `"github"`, or `"notion"`, which are saved as `DBEntry.file_source` in the database. There is no way to distinguish between Obsidian-originated content and other sources with the current model.

**Proposed Solution:**
- Add a new `EntrySource` value (e.g. `"obsidian"`) in both frontend and backend to mark entries originating from the Obsidian plugin.
- Update validation, ingestion, and query logic to recognize and persist `"obsidian"` as a valid source.
- Adapt deletion, filtering, and API/listing features to allow distinguishing content by its specific EntrySource.

**Motivation/Use-case:**
- Safely perform operations (delete, filter, query, list) that affect only Obsidian-imported content.
- Prevent accidental data loss when deleting or modifying content ingested by other pathways.
- Enable better tracking and debugging of content origins.

majiayu000:
## Summary
- Add `OBSIDIAN` value to `EntrySource` enum for content imported via Obsidian plugin
- Pass `file_source` parameter through the indexing pipeline:
  - `indexer` -> `configure_content` -> `text_search.setup` -> content processors
- Detect Obsidian client in the indexer API and set appropriate `file_source`
- Update all content processors to accept and propagate `file_source` parameter

## Motivation
This enables users to specifically target Obsidian-imported content for operations like deletion, filtering, or querying without affecting content from other sources (desktop client, manual ingestion, etc.).

## Test plan
- [x] Ruff lint passes
- [x] Ruff format passes
- [ ] CI tests pass

Fixes #1239

### DIFF
--- a/src/khoj/database/models/__init__.py
+++ b/src/khoj/database/models/__init__.py
@@ -781,6 +781,7 @@ class Entry(DbBaseModel):
         COMPUTER = "computer"
         NOTION = "notion"
         GITHUB = "github"
+        OBSIDIAN = "obsidian"
 
     user = models.ForeignKey(KhojUser, on_delete=models.CASCADE, default=None, null=True, blank=True)
     agent = models.ForeignKey(Agent, on_delete=models.CASCADE, default=None, null=True, blank=True)
--- a/src/khoj/processor/content/docx/docx_to_entries.py
+++ b/src/khoj/processor/content/docx/docx_to_entries.py
@@ -18,7 +18,9 @@ class DocxToEntries(TextToEntries):
         super().__init__()
 
     # Define Functions
-    def process(self, files: dict[str, str], user: KhojUser, regenerate: bool = False) -> Tuple[int, int]:
+    def process(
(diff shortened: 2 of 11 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

Issue #1252, opened 2026-02-06
Allow disabling auto-scroll on new token in the web interface


JeffKwasha:
### Describe the feature

Please add a user setting to disable auto-scrolling in the web chat interface.

### Use Case

As long responses steam in, users need to scroll up to read earlier content but every new token forcibly scrolls to the bottom.

This is a standard feature in many AI tools (e.g., ChatGPT has a ‘stop scrolling’ button).

### Proposed Solution

Why This Happens
The auto-scroll behavior is implemented in the frontend (React/Next.js) using logic like this:

chatContainer.scrollTop = chatContainer.scrollHeight;
This ensures the latest message is always visible — but it ignores whether the user has manually scrolled up.

A smarter implementation (like in [Ojansen/chat-auto-scroll](https://github.com/Ojansen/chat-auto-scroll)) only auto-scrolls if the user is already at the bottom 🥇 

PS: The bulk of this feature request was written by Khoj after I asked it how to fix autoscrolling. 👍 

_No response_

yang1002378395-cmyk:
## 功能已存在

好消息！Khoj 已经实现了你描述的智能滚动行为。

**代码实现**（）：

```typescript
const detectIsNearBottom = () => {
    const { scrollTop, scrollHeight, clientHeight } = scrollAreaEl;
    const bottomThreshold = 50; // pixels from bottom
    const distanceFromBottom = scrollHeight - (scrollTop + clientHeight);
    const isNearBottom = distanceFromBottom <= bottomThreshold;
    setIsNearBottom(isNearBottom);
};

// Auto scroll while incoming message is streamed
useEffect(() => {
    if (props.incomingMessages && props.incomingMessages.length > 0 && isNearBottom) {
        scrollToBottom(true);
    }
}, [props.incomingMessages, isNearBottom]);
```

**行为说明**：
- 只有当用户滚动到底部 **50px 内** 时才会自动滚动
- 如果用户向上滚动查看历史内容，新 token 不会强制滚动到底部
- 滚回底部后自动恢复跟随

**可能的问题**：
- 50px 阈值可能太小，用户觉得"在底部"但仍触发滚动
- 是否需要增加用户设置来调整阈值？

你能确认一下当前版本的行为是否符合预期吗？

PR #1267, opened 2026-03-06
Allow disabling auto-scroll on new token in web chat

kowshik24:
Related to https://github.com/khoj-ai/khoj/issues/1252

Description:
- This PR adds user control over streaming auto-scroll in the web chat interface.
- Added a new **Chat Experience** setting: **Auto-scroll while responses stream**.
- Persisted preference via [localStorage] using `chat_auto_scroll_on_stream`.
- Updated chat scrolling behavior to only follow new tokens when the setting is enabled.
- Added an in-chat **Pause/Resume auto-scroll** quick control near the down-arrow button.
- The quick control is intentionally shown only during active streaming to reduce idle UI clutter.

### DIFF
--- a/src/interface/web/app/common/utils.ts
+++ b/src/interface/web/app/common/utils.ts
@@ -19,6 +19,8 @@ const locationFetcher = () =>
 export const toTitleCase = (str: string) =>
     str.replace(/\w\S*/g, (txt) => txt.charAt(0).toUpperCase() + txt.slice(1).toLowerCase());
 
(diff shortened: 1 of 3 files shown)


debanjum:
Hey @kowshik24, thanks for opening a PR for this. But enabling/disabling auto-scroll via settings seems overkill, we don't want to overload folks with too many settings. 

Users can stop auto scroll by just scrolling away from bottom of screen for now. If in the future auto-scroll sounds like a bad idea in general, we should just disable that. But for now I'll avoid working on that.

Closing this PR for now.

PR #1286, opened 2026-03-22
feat: add Novita AI as a chat model provider

Alex-wuhu:
## Summary

[Novita AI](https://novita.ai) offers an OpenAI-compatible LLM API with competitive open-weight models. This PR adds Novita as a first-class chat provider alongside OpenAI, Google, and Anthropic.

### Changes

- **`ChatModel.ModelType`** — adds `novita` enum value + migration `0100`
- **`constants.py`** — adds `default_novita_chat_models` list and per-model pricing
- **`conversation/novita/`** — new provider module; delegates to the shared OpenAI utilities since Novita exposes a fully OpenAI-compatible endpoint (`https://api.novita.ai/openai`)
- **`conversation/utils.py`** — adds `model_to_prompt_size` entries for all three default models
- **`routers/helpers.py`** — wires Novita into all four dispatch points (`send_message_to_model`, `send_message_to_model_wrapper_sync`, `agenerate_chat_response`, `is_ready_to_chat`)
- **`initialization.py`** — adds non-interactive setup path (reads `NOVITA_API_KEY`, sets base URL automatically); marks `moonshotai/kimi-k2.5` as vision-capable
- **`tests/helpers.py` / `tests/conftest.py`** — adds Novita to provider auto-detection and test fixture setup

### Default models

| Model | Context | Notes |
|---|---|---|
| `moonshotai/kimi-k2.5` | 262K | Default; vision + function calling |
| `zai-org/glm-5` | 202K | Function calling + structured output |
| `minimax/minimax-m2.5` | 204K | Function calling + structured output |

### Setup

Set `NOVITA_API_KEY` before starting the server (or configure via `/server/admin`). The base URL `https://api.novita.ai/openai` is set automatically.

To test with Novita models:
```
KHOJ_TEST_CHAT_PROVIDER=novita NOVITA_API_KEY=<key> pytest
```

### DIFF
--- a/src/khoj/database/migrations/0100_chatmodel_novita_provider.py
+++ b/src/khoj/database/migrations/0100_chatmodel_novita_provider.py
@@ -1,0 +1,24 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("database", "0099_usermemory"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="chatmodel",
+            name="model_type",
+            field=models.CharField(
+                choices=[
+                    ("openai", "Openai"),
+                    ("anthropic", "Anthropic"),
+                    ("google", "Google"),
+                    ("novita", "Novita"),
+                ],
+                default="google",
+                max_length=200,
+            ),
+        ),
+    ]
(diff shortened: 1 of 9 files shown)


debanjum:
Thanks for the PR but Khoj supports all model providers exposing an Openai compatible API, including Novita. So this change is not necessary

PR #1292, opened 2026-03-23
Fix UnboundLocalError in PdfToEntries.extract_text when PDF processing fails

### LINKED ISSUE #1289: UnboundLocalError in pdf_to_entries.py when uploading PDF files
**Description**
Uploading files fails with a 500 error when running Khoj self-hosted locally. No file format works — PDF, .md, .txt, .docx, and .html all fail. The error message shown is: "Something went wrong while uploading your data. Error converting files. Error: HTTP error! status: 500."

**Error log**

Oh no!
Something went wrong while uploading your data

Error converting files. Error: HTTP error! status: 500. Please try again, or contact team@khoj.dev if the issue persists.

BillionClaw:
When PyMuPDFLoader fails to process an invalid PDF file, the exception is caught but
pdf_entry_by_pages is referenced before assignment, causing an UnboundLocalError.

Initialized pdf_entry_by_pages to an empty list before the try block so the return
statement always has a valid value, even when an exception occurs.

Verified with both invalid input (returns []) and valid PDFs (returns extracted text).

Fixes #1289

### DIFF
--- a/src/khoj/processor/content/pdf/pdf_to_entries.py
+++ b/src/khoj/processor/content/pdf/pdf_to_entries.py
@@ -94,6 +94,7 @@ class PdfToEntries(TextToEntries):
     @staticmethod
     def extract_text(pdf_file):
         """Extract text from specified PDF files"""
+        pdf_entry_by_pages = []
         try:
             # Create temp file with .pdf extension that gets auto-deleted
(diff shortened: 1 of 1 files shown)


BillionClaw:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

PR #1309, opened 2026-04-16
Add uv.lock comparison tool for dependency tracking

Applied via garcia ladle from recipe: uv-lock-check

postphotos:
## feat(scripts): add uv.lock comparison tool for dependency tracking

**Source:** garcia recipe `uv-lock-check`

### Changes

- `scripts/compare_uv_lock.py` — compare local vs upstream uv.lock for breaking changes

### Why this matters

When maintaining forks with `uv.lock` files, it's critical to know when upstream makes
breaking dependency changes (new packages, downgrades, or major version bumps). This script
automates that detection and can be integrated into CI to prevent dependency drift.

### Exit codes

| Code | Meaning |
|------|---------|
| 0 | No breaking changes (minor/patch upgrades only) |
| 2 | Breaking changes detected |
| 1 | Error (missing files, parse error) |

### Usage

```bash
# Compare local vs upstream
python scripts/compare_uv_lock.py

# Compare specific files
python scripts/compare_uv_lock.py local.lock upstream.lock
```

### Verification

- [x] Script is self-contained (only needs packaging.version, with fallback)
- [x] Correctly identifies MAJOR version bumps, downgrades, and new packages
- [x] No breaking changes

### Branch

```
git checkout pr/uv-lock-check
```

---

*Generated by garcia 🍦*

### DIFF
--- a/scripts/compare_uv_lock.py
+++ b/scripts/compare_uv_lock.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+"""Compare local uv.lock with upstream uv.lock for breaking changes."""
+from __future__ import annotations
+import sys, argparse, tomllib
+from pathlib import Path
+
+try:
+    from packaging.version import Version
+except Exception:
+    Version = None
+
+def load_lock(path: Path) -> dict:
+    if not path.exists():
+        raise FileNotFoundError(path)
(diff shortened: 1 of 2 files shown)


postphotos:
Also closing this one! Sorry!

PR #1318, opened 2026-04-23
Fix issue 1282 self-hosted model selection when billing is disabled

### LINKED ISSUE #1282: Self-hosted: chat_advanced overrides all model selection, sidebar picker and chat_default are ignored
## Problem
On self-hosted Khoj (v1.42.10, billing disabled), `chat_advanced` is used for ALL LLM calls — intent classification, code generation, title generation, and chat responses. The sidebar model picker and `chat_default` server setting have no effect.

## Root Cause
In `database/adapters/__init__.py`, `ais_user_subscribed()` returns `True` when `state.billing_enabled` is False (line 422). This means all self-hosted users are treated as "subscribed."

In `aget_default_chat_model()`, for subscribed users with `fast=True`:
1. `think_paid_fast` — checked first, but these columns don't exist yet in v1.42.10 schema (migration not shipped), so always None
2. `think_paid_deep` — same, always None
3. `chat_advanced` — **this is what gets returned for every call**

`chat_default` is only reached in the `else` (non-subscribed) branch at line 1345, which never executes on self-hosted.

## Impact
1. **Sidebar model picker is ignored** — `UserConversationConfig.setting_id` is never checked because `chat_advanced` takes priority
2. **`chat_default` is dead config** — never used on self-hosted
3. **Intent classification uses the same model as chat** — no way to use a fast/cheap model for tool selection and a different model for responses
4. **Agent `chat_model` assignment is ignored for the default agent** — `get_agent_chat_model` returns dynamically resolved model which hits the same `chat_advanced` path

## Verified With
- LangFuse traces showing all 4 steps (intent, code gen, chat, title) use `chat_advanced` model regardless of sidebar selection
- Tested changing `chat_default`, sidebar model picker, agent model — all ignored
- Only changing `chat_advanced` in the database affects which model is used

## Environment
- Khoj v1.42.10 (ghcr.io/khoj-ai/khoj:latest)
- Self-hosted, billing disabled
- PostgreSQL backend
- LiteLLM as OpenAI-compatible proxy

## Suggested Fix
For self-hosted (billing disabled), either:
1. Don't treat all users as "subscribed" — or add a separate flag for self-hosted
2. Make `chat_default` work as the fast/intent model regardless of subscription status
3. Respect the sidebar model picker for the chat response step even when server settings are configured

MujtabaHadi:
## Summary
Fix self-hosted model selection so `chat_advanced` does not override the default chat model path when billing is disabled.

## Problem
On self-hosted Khoj, users could be treated as subscribed in the default-model selection path. As a result, `chat_advanced` could override normal model selection behavior, including the default chat model and user-selected model behavior.

## Fix
Update:
- `ConversationAdapters.get_default_chat_model`
- `ConversationAdapters.aget_default_chat_model`

so the subscribed/advanced branch only applies when `state.billing_enabled` is true.

## Validation
I reproduced the issue locally in a self-hosted Docker setup using Ollama and OpenAI-compatible routing.

Before the fix:
- changing `chat_advanced` controlled the actual chat response path

After the fix:
- helper/background steps used the advanced model
- the actual chat response used the default model

## Related issue
Fixes #1282


debanjum:
chat_advanced overriding chat_default is the expected behavior. We should either not set it or set it to a chat model that you prefer when self-hosting. Plus this PR has gone stale, so closing this

PR #1323, opened 2026-05-11
Fix shell command injection in operator text_editor actions (CWE-78)

andesyteoss:
## Summary

This PR fixes a shell-command-injection class of bug (CWE-78) in the operator agent's `ComputerEnvironment` text-editor actions. The pre-fix code built shell strings via ad-hoc single-quote escaping and ran them through `bash -c`. Any payload that breaks the fragile quoting in `file_path`, `file_text`, `old_str`, or `new_str` reaches the shell.

Affected file: `src/khoj/processor/operator/operator_environment_computer.py`
Affected actions: `text_editor_view`, `text_editor_create`, `text_editor_str_replace`, `text_editor_insert`.

## Vulnerable pattern (before)

```python
escaped_path = file_path.replace("'", "'\"'\"'")
escaped_old  = old_str.replace(...).replace("'", "'\"'\"'")
escaped_new  = new_str.replace(...).replace("'", "'\"'\"'")
cmd = f"sed -i.bak 's/{escaped_old}/{escaped_new}/g' '{escaped_path}'"
result = await self._execute_shell_command(cmd)   # bash -c cmd
```

The `replace("'", "'\"'\"'")` trick only handles single quotes. For `sed`, the regex side still has its own metacharacters (`\\`, `/`, `&`, newlines) and the escaping was incomplete and order-dependent. Likewise, `echo '...' > '...'` for `text_editor_create` doesn't survive arbitrary file content. A crafted `old_str`/`new_str`/`file_text`/`file_path` can terminate the quoting and append arbitrary shell, e.g. closing the `sed` script and chaining `; <cmd>` in the sandbox.

## Fix

Replace the shell-built file operations with native Python I/O:

- `text_editor_view` → new `_file_view()` using `os.path.isdir`, `os.walk` (depth-limited), and `open().readlines()` with optional line range.
- `text_editor_create` → new `_file_create()` using `open(path, 'w')`.
- `text_editor_str_replace` → new `_file_str_replace()` that reads, calls `str.replace()`, and writes back (plus a `.bak` to preserve previous behavior).
- `text_editor_insert` → new `_file_insert()` that splits, inserts at the requested line index, rejoins, and writes back.

No string is concatenated into a shell command for any of these actions, so the injection class is eliminated entirely. The `terminal` action (which is by design a shell) is untouched in behavior, but its dispatcher was switched from `subprocess.run(cmd, shell=True)` to `subprocess.run(["bash", "-c", cmd])` for slightly clearer intent — semantically equivalent.

Public API and return shapes (`{"success", "output"/"error", optional "is_dir"}`) are preserved so the surrounding match-case logic and operator agent prompts continue to work unchanged.

## Tests

Added `tests/test_operator_shell_injection.py` (347 lines) covering:

- Each new helper individually (view file/dir/range, create, str_replace single & multi, insert at start/middle/end).
- Injection-style inputs in `path`, `file_text`, `old_str`, `new_str` — strings containing `'; rm -rf /tmp/marker; #`, backticks, `$()`, newlines, backslashes, `&`, `/`. The tests assert the literal characters land on disk and that no `marker` side-effect file is ever created.
- Round-tripping through the `Action` dispatch to confirm the `match` arms call the new helpers.

All tests pass locally. Lint cleanup commit removes unused imports and reorders to satisfy the repo's style config.

## Security analysis

Preconditions to reach the sink:

1. Operator-agent feature is enabled for the user (premium-scoped in current code).
2. An attacker can influence the LLM's chosen `text_editor_*` action arguments. The realistic paths are (a) direct prompt to a willing model, and (b) indirect prompt injection through content the agent reads — a synced document, a web page the agent browses, a tool output.
3. The action runs inside the operator's Docker container (default), so impact is container-scoped RCE, not host RCE.

Why we think this is worth fixing even though `terminal` exists in the same environment: the `terminal` action is the obvious capability and is presumably the action that prompt/policy/system-message guardrails are most likely to constrain. The `text_editor_*` actions look like file edits to anyone reading a transcript, but with the old escaping they were a covert path to `bash -c`. Closing that path means a model that has been restricted to "just edit files" can no longer be tricked, via injected content, into executing arbitrary commands.

### Before submitting

We tried to disprove this. We considered whether the premium gate, the Docker sandbox, or the LLM's own refusal behavior make the bug not worth fixing. They don't: the sandbox limits blast radius but doesn't remove the bug class, and indirect prompt injection via attacker-controlled file content is a realistic path where the model itself isn't trying to escape — the malicious bytes in `old_str`/`new_str` do the escaping. Switching these paths off the shell entirely is a small, isolated change with no behavior regression.

### DIFF
--- a/src/khoj/processor/operator/operator_environment_computer.py
+++ b/src/khoj/processor/operator/operator_environment_computer.py
@@ -3,6 +3,7 @@ import asyncio
 import base64
 import io
 import logging
+import os
 import platform
 import subprocess
 from typing import Literal, Optional, Union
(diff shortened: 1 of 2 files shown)


andesyteoss:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

PR #1324, opened 2026-05-12
Fix Host header injection in share URL generation (CWE-74)

andesyteoss:
## Summary

The `duplicate_chat_history_public_conversation` endpoint in `src/khoj/routers/api_chat.py` constructs share URLs using the user-controlled `Host` HTTP header. An authenticated user can inject a crafted Host header to generate share links pointing to an attacker-controlled domain.

**CWE-74 (Injection) / Open Redirect — Medium severity**

## Vulnerability Details

The previous code extracted the domain from `request.headers.get("host")` and validated it by splitting on `:` and checking against `ALLOWED_HOSTS`. This check is bypassable using URL authority syntax:

```
Host: khoj.dev:x@evil.com
```

When split on `:`, this yields `khoj.dev` which passes the `ALLOWED_HOSTS` check. However, when the full value is used to construct a URL like `https://khoj.dev:x@evil.com/share/...`, browsers interpret `khoj.dev:x` as credentials and `evil.com` as the actual host — redirecting the victim to the attacker's domain.

The generated share URL is returned to the client and can be distributed to other users, making this a phishing/open-redirect vector.

## Fix

Replace the user-controlled `Host` header with the server-configured `KHOJ_DOMAIN` setting (already used elsewhere in the codebase for CSRF trusted origins, allowed hosts, etc.). The scheme is similarly derived from the `DISABLE_HTTPS` / `KHOJ_NO_HTTPS` environment variable instead of the request.

This is a minimal, 1-file change that removes the attack surface entirely — the share URL domain is no longer influenced by any client-supplied input.

## Testing

- Verified the fix compiles and the patched function correctly uses `KHOJ_DOMAIN` and `DISABLE_HTTPS`, both of which are already imported and used throughout the settings module.
- Confirmed `KHOJ_DOMAIN` defaults to `"khoj.dev"` and `DISABLE_HTTPS` defaults to `False`, preserving existing behavior for production and development environments.
- Confirmed no other callers or tests depend on the removed `is_local_api` import (it remains available in its source module).

## Security Analysis

Before submitting, we verified that no existing mitigation prevents this. The `ALLOWED_HOSTS` Django/Starlette middleware validates the Host header for *routing* purposes but does not prevent its misuse when the application reads it directly and embeds it in output. The `:` split check was the only defense, and it is trivially bypassed with the `user:pass@host` URL authority syntax. The fix eliminates the entire class of Host header injection for this endpoint by removing reliance on client input.

### DIFF
--- a/src/khoj/routers/api_chat.py
+++ b/src/khoj/routers/api_chat.py
@@ -23,7 +23,7 @@ from fastapi.websockets import WebSocketState
 from starlette.authentication import has_required_scope, requires
 from starlette.requests import URL, Headers
 
-from khoj.app.settings import ALLOWED_HOSTS
+from khoj.app.settings import ALLOWED_HOSTS, DISABLE_HTTPS, KHOJ_DOMAIN
 from khoj.database.adapters import (
     AgentAdapters,
     ConversationAdapters,
@@ -34,7 +34,6 @@ from khoj.database.adapters import (
 )
 from khoj.database.models import Agent, KhojUser
 from khoj.processor.conversation import prompts
-from khoj.processor.conversation.openai.utils import is_local_api
 from khoj.processor.conversation.prompts import no_entries_found
 from khoj.processor.conversation.utils import (
     OperatorRun,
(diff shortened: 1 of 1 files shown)


lewiswigmore:
Closing this to reduce the open-PR pile-up — we have multiple outstanding security contributions to this repo and that volume is not fair on your review queue. Keeping #1323 (CWE-78: Fix shell command injection in operator text_editor actions ) as the primary one to focus attention on.

Happy to revisit this finding separately later if it is still relevant. Apologies for the noise.

PR #1333, opened 2026-06-03
Add Keenable web search provider

kroki221:


debanjum:
We're not adding new web search providers to Khoj currently

ilya-bogin-keenable:
Hi, I'm Ilya from Keenable. Sorry about #1333: it went up from our fork with no description and no context. That's on us.

The reason I think it's worth a second look is that Keenable needs no API key. Every provider you turned down that week needs an account first, so a self-hosted Khoj with no keys and no SearXNG returns nothing for `/online` today. Keenable would make that case work by default, and a key only raises the rate limit.

Does that change your decision? If not, that's fine and I'll drop it.

PR #1335, opened 2026-06-06
fix: check conversation existence before accessing .title in generate_chat_title

Hariom9893:
## What
Fixes an AttributeError crash in `generate_chat_title`.

## Why
If `aget_conversation_by_user` returns `None`, the current code accesses
`conversation.title` before the `None` check, raising an AttributeError
before the 404 guard is ever reached.

## Fix
Swapped the order of the two checks so `None` is handled first.


Hariom9893:
Any update regarding the pull request , if it needs any changes please inform me so i can fix it

PR #1348, opened 2026-06-24
Support MiniMax M3 chat model

octo-patch:
## Summary

Add support for [MiniMax M3](https://platform.minimax.io/docs/api-reference/text-openai-api) as a chat model. M3 is served over an OpenAI-compatible API, so it already works through Khoj's existing OpenAI provider path (an `AiModelApi` with `model-type: OpenAI` and the MiniMax API base URL). This PR adds the two small bits of model-specific handling Khoj uses for other third-party OpenAI-compatible models.

MiniMax M2 is already recognized as an in-stream thinking model in `is_instream_thinking_model`; this extends the same handling to M3.

## Changes

- **In-stream thinking detection**: MiniMax M3, like M2 and Kimi K2 Thinking, emits its reasoning inline as `<think>...</think>` blocks in the content stream. Add `MiniMax-m3` to `is_instream_thinking_model` so its reasoning is parsed by the in-stream thought processor instead of leaking into the visible chat response.
- **Cost tracking**: register `MiniMax-M3` pay-as-you-go pricing in `model_to_cost` (`$0.6` input / `$2.4` output / `$0.12` cache read per M tokens).
- **Tests**: add unit tests for in-stream thinking detection and the M3 cost entry.

## How to use

Create an AI Model API with the MiniMax base URL (`https://api.minimax.io/v1`) and your `MINIMAX_API_KEY`, then add a chat model named `MiniMax-M3` with `model-type` set to `OpenAI`.

## Testing

- `ruff check` and `ruff format --check` pass on the changed files
- Added unit tests in `tests/test_openai_utils.py` covering in-stream thinking detection (M2/M3/Kimi true; gpt-4o/claude/gemini false) and the M3 cost lookup
- Verified `MiniMax-M3` is a valid model id via the MiniMax OpenAI-compatible `/v1/models` endpoint


tests/test_openai_utils.py
debanjum, line 1: These tests don't test anything substantial and should be removed.

debanjum (changes requested):
Thanks for the PR, changes look specific and good apart from the one comment I left

PR #1357, opened 2026-06-30
fix: use tiktoken for accurate CJK token counting in document chunking

### LINKED ISSUE #1354: Markdown chunking fails for CJK languages due to text.split() tokenizer undercounting tokens
## Summary

Markdown files containing CJK (Chinese, Japanese, Korean) text are indexed as **single entries** instead of being chunked by headings. This severely degrades search quality because the bi-encoder averages the semantics of the entire document, and search hits land on the YAML frontmatter/first heading instead of relevant sections.

## Root cause

`TextToEntries.tokenizer` in `src/khoj/processor/content/text_to_entries.py` is implemented as `text.split()`:

```python
def tokenizer(text: str) -> List[str]:
    "Tokenize text into words."
    return text.split()
```

For languages without word separators (Chinese, Japanese, Korean), `text.split()` produces far fewer tokens than the actual subword/token count. For example, a 744-tiktoken Chinese markdown file is counted as only ~152 tokens by `text.split()`.

`MarkdownToEntries.process_single_markdown_file` decides whether to split by headings using:

```python
if len(TextToEntries.tokenizer(markdown_content_with_ancestry)) <= max_tokens(256) \
   or not re.search(rf"^#{{{len(ancestry)+1},}}\s", markdown_content, flags=re.MULTILINE):
    # save as single entry
```

Because the first condition is almost always true for CJK documents, the function never reaches the heading-splitting logic.

## Version

`1.42.10` (ghcr.io/khoj-ai/khoj:latest as of 2026-06-28)

## Reproduction steps

1. Start Khoj server.
2. Upload a markdown file with multiple H2 headings and several hundred CJK characters. Example structure:
   ```markdown
   ---
   title: Note
   ---
   # Topic
   ## Section A
   ... ~300 Chinese characters ...
   ## Section B
   ... ~300 Chinese characters ...
   ## Section C
   ... ~300 Chinese characters ...
   ```
3. Check the database:
   ```sql
   SELECT file_name, COUNT(*) FROM database_entry GROUP BY file_name;
   ```
   **Expected:** multiple rows (one per section)
   **Actual:** one row per file
4. Search for a concept from `Section B`. The top result is likely the document title/frontmatter, not `Section B`.

## Expected behavior

CJK markdown files should be chunked by headings just like English markdown files, producing one `database_entry` per section when each section exceeds the token threshold.

## Actual behavior

Each CJK markdown file becomes a single entry, regardless of length or heading structure.

## Suggested fix

Replace `text.split()` with a real tokenizer, e.g. tiktoken `cl100k_base`:

```python
import tiktoken

def tokenizer(text: str) -> List[int]:
    if not hasattr(TextToEntries, "_tiktoken_enc"):
        TextToEntries._tiktoken_enc = tiktoken.get_encoding("cl100k_base")
    return TextToEntries._tiktoken_enc.encode(text)
```

All existing call sites only use `len(TextToEntries.tokenizer(chunk))`, so returning token IDs instead of strings is safe and more accurate across languages.

## Workaround

Override `src/khoj/processor/content/text_to_entries.py` in a custom image or bind-mount to use tiktoken instead of `text.split()`.

## Impact

All CJK users indexing markdown notes are affected. English notes are not affected because `text.split()` approximates token count reasonably well for whitespace-separated languages.

## Labels

bug, internationalization, search, markdown

Sanjay-Kirti:
## Summary
Fixes #1354

Replace `text.split()` tokenizer in `TextToEntries` with tiktoken `cl100k_base` encoding to fix markdown/org chunking for CJK (Chinese, Japanese, Korean) languages.

## Problem
`TextToEntries.tokenizer()` uses `text.split()` which splits on whitespace. CJK languages don't use spaces between words, so a 744-token Chinese document was counted as ~152 tokens. This caused `MarkdownToEntries` (and `OrgToEntries`) to save entire CJK documents as single entries instead of chunking by headings severely degrading search quality for CJK users.

## Changes
- **`src/khoj/processor/content/text_to_entries.py`**: Replace `text.split()` with `tiktoken.get_encoding("cl100k_base").encode()` in `TextToEntries.tokenizer()`. Encoder is lazy initialized and cached as a class attribute.
- **`tests/test_markdown_to_entries.py`**: Add 3 CJK specific test cases:
  - Chinese markdown is chunked by headings (core bug fix)
  - Small CJK content stays as single entry (regression test)
  - Mixed CJK/English markdown is properly chunked

## Token Count Improvements
Concrete examples of how tiktoken correctly counts tokens compared to the old whitespace-split method:

| Text | `text.split()` (old) | `tiktoken` (new) | Result |
|------|------|------|------|
| Chinese sentence (65 chars) | 1 token ❌ | 69 tokens ✅ | Accurate counting |
| Full CJK markdown example | 11 tokens ❌ | 212 tokens ✅ | Properly chunks now |
| Japanese sentence | 1 token ❌ | 55 tokens ✅ | Accurate counting |
| English sentence (18 words) | 18 tokens ✅ | 18 tokens ✅ | Identical behavior |

## Why this is safe
All 3 call sites only use `len(TextToEntries.tokenizer(...))` they never inspect individual tokens. The return type change from `List[str]` → `List[int]` has no impact. `tiktoken` is already a project dependency (used in `helpers.py` for chat token counting).

## Test Results
Ran the full test suite locally with Postgres/pgvector enabled:
- All 12/12 tests in `test_markdown_to_entries.py` pass (including the 3 new CJK ones)
- No regressions for English markdown processing

### DIFF
--- a/src/khoj/processor/content/text_to_entries.py
+++ b/src/khoj/processor/content/text_to_entries.py
@@ -6,6 +6,7 @@ from abc import ABC, abstractmethod
 from itertools import repeat
 from typing import Any, Callable, List, Set, Tuple
 
+import tiktoken
 from langchain_text_splitters import RecursiveCharacterTextSplitter
 from tqdm import tqdm
 
@@ -53,9 +54,11 @@ class TextToEntries(ABC):
         return "".join(filtered_text)
 
     @staticmethod
-    def tokenizer(text: str) -> List[str]:
-        "Tokenize text into words."
-        return text.split()
+    def tokenizer(text: str) -> List[int]:
+        "Tokenize text into tokens using tiktoken for accurate token counting across all languages."
+        if not hasattr(TextToEntries, "_tiktoken_enc"):
+            TextToEntries._tiktoken_enc = tiktoken.get_encoding("o200k_base")
+        return TextToEntries._tiktoken_enc.encode(text)
 
     @staticmethod
     def split_entries_by_max_tokens(
--- a/tests/test_markdown_to_entries.py
+++ b/tests/test_markdown_to_entries.py
@@ -169,7 +169,7 @@ body line 1.1
 
     # Act
     # Extract Entries from specified Markdown files
-    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=12)
+    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=28)
 
     # Assert
     assert len(entries) == 2
@@ -196,7 +196,7 @@ longer body line 2.1
 
     # Act
     # Extract Entries from specified Markdown files
-    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=12)
+    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=25)
 
     # Assert
     assert len(entries) == 2
@@ -264,6 +264,29 @@ def test_line_number_tracking_in_recursive_split():
         )
 
 
+def test_cjk_markdown_is_chunked_by_headings(tmp_path):
+    "CJK and mixed CJK/English markdown with multiple headings should be split into separate entries."
+    entry = """# 主题
+## 第一节
+这是第一节的内容。人工智能是计算机科学的一个分支，它企图了解智能的实质，并生产出一种新的能以人类智能相似的方式做出反应的智能机器。
+## 第二节
+这是第二节的内容。机器学习是人工智能的一个重要分支，它使用统计技术让计算机系统能够从数据中学习，而不需要被明确编程。
+## English Section
+This is a section written in English with enough words to make the token count meaningful for testing purposes.
+"""
+    data = {
+        f"{tmp_path}": entry,
+    }
+
+    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=20)
+
+    assert len(entries) == 2
+    assert len(entries[1]) > 1, (
+        "CJK markdown should be chunked by headings, not saved as single entry. "
+        "If this fails, the tokenizer is likely undercounting CJK tokens."
+    )
+
+
 # Helper Functions
 def create_file(tmp_path: Path, entry=None, filename="test.md"):
     markdown_file = tmp_path / filename
@@ -276,3 +299,4 @@ def create_file(tmp_path: Path, entry=None, filename="test.md"):
 def clean(text):
     "Normalize spaces in text for easier comparison."
     return re.sub(r"\s+", " ", text)
+


tests/test_markdown_to_entries.py
```diff
@@ -264,6 +264,67 @@ def test_line_number_tracking_in_recursive_split():
...
+    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=20)
+
+    assert len(entries) == 2
+    assert len(entries[1]) > 1, (
+        "CJK markdown should be chunked by headings, not saved as single entry. "
+        "If this fails, the tokenizer is likely undercounting CJK tokens."
+    )
+
+
+def test_cjk_markdown_small_content_stays_single_entry(tmp_path):
```
debanjum: This test seems to be unnecessary. If cjk token counting is working correctly (which the previous test also proves) this test will pass too, right?

tests/test_markdown_to_entries.py
```diff
@@ -264,6 +264,67 @@ def test_line_number_tracking_in_recursive_split():
...
+        f"{tmp_path}": entry,
+    }
+
+    entries = MarkdownToEntries.extract_markdown_entries(markdown_files=data, max_tokens=256)
+
+    assert len(entries) == 2
+    assert len(entries[1]) == 1, "Small CJK content should stay as a single entry"
+
+
+def test_mixed_cjk_english_markdown_is_chunked(tmp_path):
```
debanjum: We can merge this into the first test. Main thing we're verifying is cjk token counting is working better now at a high level. 1 test for that containing the 2 consecutive Chinese notes followed by 1 English note should suffice imo.

src/khoj/processor/content/text_to_entries.py
```diff
@@ -53,9 +54,11 @@ def remove_long_words(text: str, max_word_length: int = 500) -> str:
         return "".join(filtered_text)
 
     @staticmethod
-    def tokenizer(text: str) -> List[str]:
-        "Tokenize text into words."
-        return text.split()
+    def tokenizer(text: str) -> List[int]:
+        "Tokenize text into tokens using tiktoken for accurate token counting across all languages."
+        if not hasattr(TextToEntries, "_tiktoken_enc"):
+            TextToEntries._tiktoken_enc = tiktoken.get_encoding("cl100k_base")
```
debanjum: Let's use the more modern o200k_base tokenizer (used by gpt4o, gpt5+). It has better support for tokenizing indic language as well

debanjum (commented):
Thanks for submitting a PR to fix this! Left some comments, can merge the changes once they're resolved and CI passes

Sanjay-Kirti:
@debanjum Thanks for the review, I've addressed all the feedback and just pushed the updates:
1. Switched from cl100k_base to the more modern o200k_base in ext_to_entries.py for better Indic language support.
2.  changed the CJK tests in test_markdown_to_entries.py per your suggestion:
   - Removed the unnecessary small content test.
   - Merged the mixed CJK/English test into test_cjk_markdown_is_chunked_by_headings. The single test now validates chunking with 2 Chinese notes followed by 1 English note.
All tests are passing locally. Let me know if anything else is needed or if you want me to work on another issue.
Would love to contribute.

PR #1373, opened 2026-07-13
docs: add DaoXE OpenAI-compatible gateway guide

seven7763:
## Summary
Adds a self-hosted setup page for using [DaoXE](https://daoxe.com) as an OpenAI-compatible API base with Khoj (`https://daoxe.com/v1`), and links it from the existing OpenAI proxy guide.

## Why
Khoj already documents LiteLLM / Ollama / LM Studio / generic OpenAI proxies. DaoXE is another OpenAI-compatible multi-model gateway some self-hosters use when they want hosted multi-model access without per-vendor SDKs.

## Notes
- Community contribution from a DaoXE maintainer (disclosed in the page).
- No hardcoded model/price tables; model IDs come from the user's DaoXE account catalog.
- DaoXE is not available in mainland China (noted on the page).
- Docs-only; sidebar is autogenerated from the filesystem.

## Test plan
- [ ] Preview docs build / render `documentation/docs/advanced/daoxe.md`
- [ ] Confirm link from `use-openai-proxy.md` resolves to `/advanced/daoxe`

### DIFF
--- a/documentation/docs/advanced/daoxe.md
+++ b/documentation/docs/advanced/daoxe.md
@@ -1,0 +1,39 @@
+---
+sidebar_position: 2
+---
+
+# DaoXE
+
+:::info
+This is only helpful for self-hosted users. If you're using [Khoj Cloud](https://app.khoj.dev), you're limited to our first-party models.
+:::
+
+:::info
+Khoj natively supports local LLMs [available on HuggingFace in GGUF format](https://huggingface.co/models?library=gguf). Using an OpenAI API compatible gateway with Khoj is useful when you want hosted multi-model access without wiring each vendor SDK separately.
+:::
+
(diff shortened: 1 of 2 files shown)


seven7763:
Verification notes for reviewers (docs-only):

**OpenAI-compatible base**
```bash
export DAOXE_API_KEY=...
curl -sS -H "Authorization: Bearer $DAOXE_API_KEY" https://daoxe.com/v1/models | head
```

Khoj self-host path documented here uses the same OpenAI-compatible base URL pattern as other proxies in the advanced docs (`https://daoxe.com/v1`), with live account model IDs (no static public model/price table).

Happy to adjust wording if maintainers prefer a shorter link-only mention under the existing OpenAI proxy page.

debanjum:
OpenAI compatible providers work out of the box. We're not adding new first party AI provider types at this point.

PR #1376, opened 2026-07-17
Drop client IP (client_host) from usage telemetry payload

### LINKED ISSUE #1374: Telemetry payload includes client IP (client_host) while privacy docs say "we do not log your IP address"
### Summary

The self-hosted telemetry payload includes the requesting client's IP address (`client_host`),
which seems to sit uneasily with the privacy doc's promise that *"We do not log your IP address."*
Flagging it in case the payload — or the wording — needs a small adjustment. I may well be missing
context on how the receiving endpoint handles it, so please read this as a question, not an accusation.

### What the docs say

[`documentation/docs/get-started/privacy-security.md`](https://github.com/khoj-ai/khoj/blob/1e30154d1070c7b132f389638c008b490be1481b/documentation/docs/get-started/privacy-security.md#L14-L15):

> 1. We collect completely anonymized usage telemetry and send it to PostHog. …
>     - **We do not log your IP address**, nor upload any of your personal data to PostHog.

### What the code sends (observation)

[`src/khoj/routers/helpers.py`](https://github.com/khoj-ai/khoj/blob/1e30154d1070c7b132f389638c008b490be1481b/src/khoj/routers/helpers.py#L210-L211) builds the telemetry properties with the client IP:

```python
user_state = {
    "client_host": request.client.host if request.client else None,
    ...
}
```

That dict flows through `log_telemetry` (`src/khoj/utils/helpers.py`) into `request_body` unchanged, is
appended to `state.telemetry`, and is then uploaded verbatim in
[`src/khoj/configure.py`](https://github.com/khoj-ai/khoj/blob/1e30154d1070c7b132f389638c008b490be1481b/src/khoj/configure.py#L451):

```python
response = requests.post(constants.telemetry_server, json=state.telemetry)
```

to `https://khoj.beta.haletic.com/v1/telemetry` (`src/khoj/utils/constants.py`).

### Why I'm raising it (inference)

`request.client.host` is the client's IP address, and it's present in the payload the self-hosted
server uploads. The doc's line "We do not log your IP address" is about PostHog specifically, so it's
possible the relay strips `client_host` before forwarding — that step isn't in this repo, so I can't
see it. But as written, the open-source server does upload the IP, which a privacy-conscious
self-hoster reading that promise likely wouldn't expect.

### Possible directions (your call — I didn't want to presume the fix)

- Drop `client_host` from the payload if it isn't needed (note: if it's used server-side for the
  "unique … requests" counts, this would need a hashed/derived value instead), **or**
- Hash/truncate it client-side before sending, **or**
- Adjust the doc wording to describe what is actually uploaded and how the endpoint treats it.

Happy to open a small PR for whichever direction you prefer. Thank you for building Khoj with privacy
front-and-centre — that's exactly why I thought this was worth a careful note.

<sub>AI-assisted; I reproduced the code paths and the doc claim myself against commit `1e30154d1070`.</sub>

gaurav0107:
## Summary

The self-hosted Khoj server builds its usage-telemetry payload with the
requesting client's IP address under the `client_host` key in
`update_telemetry_state` (`src/khoj/routers/helpers.py`), and uploads it
verbatim to the telemetry endpoint. This contradicts the privacy
documentation (`documentation/docs/get-started/privacy-security.md`), which
states:

> **We do not log your IP address**, nor upload any of your personal data to PostHog.

As reported in #1374, `request.client.host` is the client IP, and it was
present in the payload the open-source server uploads. A privacy-conscious
self-hoster reading that guarantee would not expect the IP to leave their
machine.

Per the maintainer's confirmation on the issue ("we do drop client I.P at
posthog layer ... We should stop passing client host as telemetry so it
agrees with the docs"), this removes the field at the source rather than
adding a surrogate.

## Changes

- Remove the `client_host` entry from the `user_state` dict in
  `update_telemetry_state`. This was the sole producer of the field; it
  flowed unchanged through `log_telemetry` into the uploaded `request_body`.

No other behaviour changes. The field had no in-repo consumer (it was
write-only into the uploaded payload), and `update_telemetry_state`'s
signature is unchanged, so none of its call sites are affected.

## Test plan

- `ruff check src/khoj/routers/helpers.py` — passes.
- `ruff format --check src/khoj/routers/helpers.py` — already formatted.
- `python -m py_compile` on the changed file — valid syntax.
- Confirmed `client_host` now has zero references under `src/`.

There is no existing test covering telemetry payload composition, and the
full pytest suite requires a Postgres-backed Django setup that I could not
provision locally, so I did not add a new test. Happy to add a focused
regression test asserting the payload no longer includes the client IP if
you'd prefer one.

Fixes #1374

### DIFF
--- a/src/khoj/routers/helpers.py
+++ b/src/khoj/routers/helpers.py
@@ -208,7 +208,6 @@ def update_telemetry_state(
     client_app: ClientApplication = request.user.client_app if request.user.is_authenticated else None
     subscription: Subscription = user.subscription if user and hasattr(user, "subscription") else None
     user_state = {
-        "client_host": request.client.host if request.client else None,
         "user_agent": user_agent or "unknown",
         "referer": referer or "unknown",
         "host": host or "unknown",


gaurav0107:
Thanks @debanjum for confirming the direction on #1374. This draft implements the minimal change you described — dropping `client_host` from the telemetry payload in `update_telemetry_state` so the uploaded data matches the "we do not log your IP address" guarantee. It's a single-line removal; the field had no in-repo consumer.

A couple of notes:
- @kobihikri kindly offered to open a PR too — happy to close this if you'd prefer their submission; I put this up in case it's useful to have a ready diff.
- I couldn't run the full pytest suite locally (it needs a Postgres-backed Django setup plus the torch/transformers deps), so I didn't add a regression test. Glad to add one asserting the payload no longer includes the client IP if you'd like it before merge.

I've left it in draft — let me know if you'd like me to mark it ready for review.

debanjum:
Hi @gaurav0107, thanks for the draft PR. @kobihikri created a PR for this. I've merged it, so closing this too

PR #1408, opened 2026-08-23
Use ast.literal_eval for UI-TARS box parsing

simpleqt:
## Problem

`src/khoj/processor/operator/grounding_agent_uitars.py` parsed the `start_box` / `end_box` strings that come from the grounding model's output with `eval()` (8 call sites, e.g. `x1, y1, x2, y2 = eval(start_box)`). Because these strings are model-generated, a crafted model response could execute arbitrary Python code in the operator process.

## Fix

The boxes are coordinate list literals (e.g. `[100, 200, 300, 400]`), so `ast.literal_eval` parses them identically with no code-execution risk. Behavior is unchanged for well-formed output; malformed strings still raise, just safely. `ast` was already imported.


simpleqt:
Gentle ping — I checked and this branch is already up to date with current `master`, so there's nothing to rebase. Happy to make any changes if there's interest, just let me know. Thanks!

PR #1409, opened 2026-08-23
security: warn when running with the default Django SECRET_KEY

simpleqt:
## What changed
`SECRET_KEY` falls back to the hardcoded `'!secret'` when `KHOJ_DJANGO_SECRET_KEY` is unset. A predictable SECRET_KEY lets anyone forge session cookies, CSRF tokens and password-reset flows, so a deployment that forgets the env var is silently running with a public secret.

This PR keeps the dev-friendly default but logs a prominent warning at startup so the misconfiguration is visible in operation logs.

## Verification
- `python -m py_compile src/khoj/app/settings.py`
- No behaviour change apart from the startup log line.


simpleqt:
Gentle ping — I checked and this branch is already up to date with current `master`, so there's nothing to rebase. Happy to make any changes if there's interest, just let me know. Thanks!

